PR #1, opened 2026-02-26
add postgres as an integration

daltoniam:


postgres/queries.go
```diff
@@ -0,0 +1,195 @@
...
+
+	format := argStr(args, "format")
+	if format == "" {
+		format = "text"
+	}
+	analyze := argBool(args, "analyze")
+
+	var explain string
+	if analyze {
+		explain = fmt.Sprintf("EXPLAIN (ANALYZE, FORMAT %s) %s", format, sqlStr)
```
acmacalister: **Must Fix — SQL injection via unsanitized `format` parameter**

The `format` argument is user-provided and interpolated directly into SQL via `fmt.Sprintf`. A malicious value like `text) DROP TABLE users; --` would produce valid destructive SQL. Additionally, when `analyze: true`, the `sqlStr` is actually *executed* (EXPLAIN ANALYZE runs the statement), so destructive queries would succeed.

Have you considered validating `format` against an allow-list and wrapping this in a read-only transaction?

```go
var validFormats = map[string]bool{"text": true, "json": true, "xml": true, "yaml": true}

func explainTool(ctx context.Context, p *postgres, args map[string]any) (*mcp.ToolResult, error) {
    sqlStr := argStr(args, "sql")
    if sqlStr == "" {
        return errResult(fmt.Errorf("sql is required"))
    }

    format := argStr(args, "format")
    if format == "" {
        format = "text"
    }
    if !validFormats[strings.ToLower(format)] {
        return errResult(fmt.Errorf("invalid format: %s (must be text, json, xml, or yaml)", format))
    }

    var explain string
    if argBool(args, "analyze") {
        explain = fmt.Sprintf("EXPLAIN (ANALYZE, FORMAT %s) %s", format, sqlStr)
    } else {
        explain = fmt.Sprintf("EXPLAIN (FORMAT %s) %s", format, sqlStr)
    }

    tx, err := p.db.BeginTx(ctx, &readOnlyTx)
    if err != nil {
        return errResult(fmt.Errorf("begin transaction: %w", err))
    }
    defer func() { _ = tx.Rollback() }()

    rows, err := tx.QueryContext(ctx, explain)
    // ... scan and return
}
```

postgres/queries.go
```diff
@@ -0,0 +1,195 @@
...
+	if err != nil {
+		return errResult(err)
+	}
+
+	columns := argStr(args, "columns")
+	if columns == "" {
+		columns = "*"
+	}
+
+	q := fmt.Sprintf("SELECT %s FROM %s.%s", columns, safeSchema, safeTable) // #nosec G201 -- identifiers are sanitized via sanitizeIdentifier
```
acmacalister: **Must Fix — SQL injection via `columns`, `where`, and `order_by` parameters**

While `table` and `schema` are sanitized via `sanitizeIdentifier` (nice!), `columns`, `where`, and `order_by` are concatenated directly into the query string. For example `columns: "1; DROP TABLE users; --"` would inject arbitrary SQL.

Since `selectTool` runs in a read-only transaction the blast radius is bounded to data exfiltration, but it's inconsistent with the effort on identifier sanitization. Suggestion: either (a) document in the tool description that these accept raw SQL fragments (acceptable for MCP tools consumed by AI agents), or (b) reject semicolons/comments the same way `sanitizeIdentifier` does.

go.mod
```diff
@@ -53,6 +53,7 @@ require (
 	github.com/google/jsonschema-go v0.4.2 // indirect
 	github.com/google/uuid v1.5.0 // indirect
 	github.com/gorilla/websocket v1.5.3 // indirect
+	github.com/lib/pq v1.11.2 // indirect
```
acmacalister: **Must Fix — `lib/pq` marked as `// indirect` but is a direct dependency**

The `_ "github.com/lib/pq"` import in `postgres/postgres.go` makes this a direct dependency. Running `go mod tidy` reclassifies it, meaning this go.mod is technically "untidy" and CI checks like `go mod tidy && git diff --exit-code` would fail. Running `go mod tidy` and committing the result should fix it.

postgres/queries.go
```diff
@@ -0,0 +1,195 @@
...
+	return rawResult(data)
+}
+
+func executeTool(ctx context.Context, p *postgres, args map[string]any) (*mcp.ToolResult, error) {
+	sqlStr := argStr(args, "sql")
+	if sqlStr == "" {
+		return errResult(fmt.Errorf("sql is required"))
+	}
+
+	data, err := p.exec(ctx, sqlStr)
```
acmacalister: **Should Fix — `executeTool` runs arbitrary write SQL with no safeguards**

Unlike `queryTool` and `selectTool` (both wrapped in read-only transactions), `executeTool` passes user-provided SQL straight to `ExecContext` with no transaction, no statement-type validation, and no confirmation. A `DROP DATABASE` is one tool call away — and this is the only integration in the project with unrestricted write access to a production resource.

Have you considered one or more of:
- A `read_only` config credential flag (default `true`) that disables `postgres_execute` entirely unless opted in
- A deny-list for destructive DDL (`DROP DATABASE`, `TRUNCATE`)
- A prominent warning in the tool description ("**CAUTION: executes arbitrary SQL including DDL/DML**")
- Logging/auditing all write operations

postgres/postgres.go
```diff
@@ -0,0 +1,249 @@
...
+
+	p.connStr = connStr
+
+	db, err := sql.Open("postgres", p.connStr)
+	if err != nil {
+		return fmt.Errorf("postgres: failed to open connection: %w", err)
+	}
+	db.SetMaxOpenConns(5)
+	db.SetMaxIdleConns(2)
+	p.db = db
```
acmacalister: **Should Fix — No `Close()` for the `*sql.DB` connection pool**

Postgres is the only adapter holding a stateful resource that requires explicit cleanup. When the server shuts down, the pool is abandoned which can leave connections in `idle in transaction` on the Postgres side.

Suggestion — implement `io.Closer` on the struct:
```go
func (p *postgres) Close() error {
    if p.db != nil {
        return p.db.Close()
    }
    return nil
}
```
Then in the shutdown path: `if c, ok := integration.(io.Closer); ok { _ = c.Close() }`

(The `Integration` interface doesn't have `Close()` today, so this is additive and non-breaking.)

postgres/postgres.go
```diff
@@ -0,0 +1,249 @@
...
+	fn, ok := dispatch[toolName]
+	if !ok {
+		return &mcp.ToolResult{Data: fmt.Sprintf("unknown tool: %s", toolName), IsError: true}, nil
+	}
+	return fn(ctx, p, args)
+}
+
+// --- Query helpers ---
+
+func (p *postgres) query(ctx context.Context, query string, args ...any) (json.RawMessage, error) {
```
acmacalister: **Should Fix — `scanRows` logic duplicated**

The row-scanning logic in this `query()` method (lines 90-137) is identical to the standalone `scanRows()` in `queries.go:153-195`. This works but will drift over time. Consider refactoring `query()` to delegate:

```go
func (p *postgres) query(ctx context.Context, query string, args ...any) (json.RawMessage, error) {
    rows, err := p.db.QueryContext(ctx, query, args...)
    if err != nil {
        return nil, fmt.Errorf("query error: %w", err)
    }
    defer func() { _ = rows.Close() }()
    return scanRows(rows)
}
```

PR #3, opened 2026-02-26
added posthog integration

daltoniam:


posthog/projects.go
```diff
@@ -0,0 +1,58 @@
...
+	data, err := p.patch(ctx, path, body)
+	if err != nil {
+		return errResult(err)
+	}
+	return rawResult(data)
+}
+
+func deleteProject(ctx context.Context, p *posthog, args map[string]any) (*mcp.ToolResult, error) {
+	data, err := p.del(ctx, "/api/projects/%s/", argStr(args, "project_id"))
+	if err != nil {
```
acmacalister: ### 3.  deleteProject  intentionally skips  p.proj()  — add a comment

 posthog/projects.go:54  Uses  argStr(args, "project_id")  directly instead of  p.proj(args) . This is actually
correct for a destructive op (don't silently delete the default project), but it's the only handler that breaks the
pattern. A one-line comment would prevent a future contributor from "fixing" it.

posthog/posthog.go
```diff
@@ -0,0 +1,266 @@
+package posthog
```
acmacalister: ## Consider (Nice to Have)

### 1. Response body size limit

 doRequest  uses bare  io.ReadAll(resp.Body)  — a malformed response could OOM. This matches the existing Metabase
pattern (pre-existing debt), but worth noting for a future hardening pass:

  data, err := io.ReadAll(io.LimitReader(resp.Body, 10*1024*1024))

### 2. HTTP client timeout

 &http.Client{}  has no timeout. A hung PostHog API call blocks forever. Consider  Timeout: 30 * time.Second . Same
pattern gap as Metabase — good follow-up.

PR #12, opened 2026-02-27
fix linear issues when I was testing it

daltoniam:


linear/issues.go
```diff
@@ -24,26 +24,29 @@ const issueFields = `
 func listIssues(ctx context.Context, l *linear, args map[string]any) (*mcp.ToolResult, error) {
 	filter := map[string]any{}
 	if team := argStr(args, "team"); team != "" {
-		filter["team"] = map[string]any{"name": map[string]any{"eqCaseInsensitive": team}}
+		filter["team"] = map[string]any{"or": []map[string]any{
+			{"name": map[string]any{"eqIgnoreCase": team}},
+			{"key": map[string]any{"eq": team}},
```
acmacalister: **Consider — Team key filter could use case-insensitive matching**

Nice improvement matching by both name and key — this is a great UX win. One thought: the `key` filter uses `{"eq": team}` (exact match) while `name` uses `eqIgnoreCase`. Team keys are typically uppercase (`ENG`, `INFRA`) but users might pass `eng` or `Eng`. Have you considered using `eqIgnoreCase` for the key as well?

**Suggestion:**
```go
filter["team"] = map[string]any{"or": []map[string]any{
	{"name": map[string]any{"eqIgnoreCase": team}},
	{"key": map[string]any{"eqIgnoreCase": team}},
}}
```

linear/issues.go
```diff
@@ -69,13 +72,13 @@ func listIssues(ctx context.Context, l *linear, args map[string]any) (*mcp.ToolR
 }
 
 func searchIssues(ctx context.Context, l *linear, args map[string]any) (*mcp.ToolResult, error) {
-	data, err := l.gql(ctx, fmt.Sprintf(`query($query: String!, $first: Int, $after: String) {
-		searchIssues(term: $query, first: $first, after: $after) {
+	data, err := l.gql(ctx, fmt.Sprintf(`query($term: String!, $first: Int, $after: String) {
```
acmacalister: **Should Fix — `searchDocuments` and `searchProjects` have the same `$query`/`term` mismatch**

This rename from `$query` to `$term` is a good clarity improvement. However `searchDocuments` (`linear/extras.go:312`) and `searchProjects` (`linear/projects.go:48`) still use `$query` as the GraphQL variable name passed to the `term:` argument — the same pattern that was fixed here.

If this was fixing an actual API issue (Linear rejecting the variable name), those two functions have the same bug. If it was just a clarity rename, it would be good to apply it consistently. Either way, worth updating in this PR to keep things uniform.

**Suggestion** (for `linear/extras.go` searchDocuments):
```go
data, err := l.gql(ctx, `query($term: String!, $first: Int) {
	searchDocuments(term: $term, first: $first) {
```
with the Go map key updated from `"query"` to `"term"`. Same for `searchProjects` in `linear/projects.go`.

acmacalister (approved):

PR #13, opened 2026-02-27
make it a daemon and run in the background

daltoniam:


cmd/server/main.go
```diff
@@ -42,6 +48,110 @@ func main() {
...
+`)
+		fs.PrintDefaults()
+	}
+
+	if len(args) == 0 {
+		fs.Usage()
+		os.Exit(1)
+	}
+
+	cmd := args[0]
```
acmacalister: **Should Fix — Flag parsing doesn’t handle `--port` before command**

`switchboard daemon --port 9999 start` will fail because `args[0]` is `--port`, which won’t match any command. Only `switchboard daemon start --port 9999` works.

This is a common CLI UX footgun. Consider either documenting the limitation explicitly in the usage text, or doing a two-pass parse: extract flags first, then match the command from `fs.Args()`.

scripts/postinstall.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+set -e
+
+# Install systemd user service for the installing user.
+# nfpms runs post-install as root, so we try to detect the real user.
+
+if command -v systemctl >/dev/null 2>&1; then
+    REAL_USER="${SUDO_USER:-$(logname 2>/dev/null || echo root)}"
+    REAL_HOME=$(eval echo "~${REAL_USER}")
```
acmacalister: **Consider — `eval echo` with variable is a shell injection vector**

`eval echo "~${REAL_USER}"` uses `eval` with a variable. In practice `REAL_USER` comes from `SUDO_USER` or `logname` so the risk is low, but `getent passwd "${REAL_USER}" | cut -d: -f6` would be safer and more portable.

PR #16, opened 2026-02-28
first run of gcp implementation

### LINKED ISSUE #4: Add GCP integration

daltoniam:
closes #4

### DIFF
--- a/gcp/gcp.go
+++ b/gcp/gcp.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"time"
 
 	compute "cloud.google.com/go/compute/apiv1"
 	"cloud.google.com/go/firestore"
@@ -14,6 +15,7 @@ import (
 	monitoring "cloud.google.com/go/monitoring/apiv3/v2"
 	"cloud.google.com/go/pubsub"
 	resourcemanager "cloud.google.com/go/resourcemanager/apiv3"
+	resourcemanagerpb "cloud.google.com/go/resourcemanager/apiv3/resourcemanagerpb"
 	run "cloud.google.com/go/run/apiv2"
 	"cloud.google.com/go/storage"
 	"google.golang.org/api/option"
@@ -62,6 +64,9 @@ func (g *integration) Configure(ctx context.Context, creds mcp.Credentials) erro
 	if g.projectID == "" {
 		return fmt.Errorf("gcp: project_id is required")
 	}
+
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
 
 	var opts []option.ClientOption
 	if v := creds["credentials_json"]; v != "" {
@@ -164,12 +169,13 @@ func (g *integration) Configure(ctx context.Context, creds mcp.Credentials) erro
 }
 
 func (g *integration) Healthy(ctx context.Context) bool {
-	if g.storageClient == nil {
+	if g.projectsClient == nil {
 		return false
 	}
-	it := g.storageClient.Buckets(ctx, g.projectID)
-	_, _ = it.Next()
-	return true
+	_, err := g.projectsClient.GetProject(ctx, &resourcemanagerpb.GetProjectRequest{
+		Name: g.projectName(),
+	})
+	return err == nil
 }
 
 func (g *integration) Tools() []mcp.ToolDefinition {
--- a/gcp/storage.go
+++ b/gcp/storage.go
@@ -15,10 +15,15 @@ import (
 
 const maxGCSGetObjectSize = 10 * 1024 * 1024
 
-func storageListBuckets(ctx context.Context, g *integration, _ map[string]any) (*mcp.ToolResult, error) {
+func storageListBuckets(ctx context.Context, g *integration, args map[string]any) (*mcp.ToolResult, error) {
+	limit := argInt(args, "limit")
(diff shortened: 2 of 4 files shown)


gcp/storage.go
```diff
@@ -0,0 +1,178 @@
...
+		query.Prefix = v
+	}
+	if v := argStr(args, "delimiter"); v != "" {
+		query.Delimiter = v
+	}
+
+	bucket := g.storageClient.Bucket(argStr(args, "bucket"))
+	var objects []map[string]any
+	it := bucket.Objects(ctx, query)
+	for i := 0; i < 1000; i++ {
```
acmacalister: Hard-coded 1000 object limit. Same concern as with compute instances — this could silently truncate results. Worth documenting or making configurable.

gcp/compute.go
```diff
@@ -0,0 +1,222 @@
...
+	if v := argStr(args, "filter"); v != "" {
+		req.Filter = proto.String(v)
+	}
+	if v := argInt32(args, "max_results"); v > 0 {
+		req.MaxResults = proto.Uint32(uint32(v))
+	}
+
+	limit := argInt(args, "limit")
+	if limit <= 0 {
+		limit = defaultComputeLimit
```
acmacalister: The `defaultComputeLimit` constant is defined in `storage.go` instead of here where it's used. Consider moving it to `gcp.go` alongside other shared constants, or defining compute-specific constants in this file.

acmacalister (commented):
Strong first implementation of the GCP adapter — follows all the established patterns, includes thorough tests, and passes CI. A couple of minor things inline about context handling and type coercion, but nothing blocking.

acmacalister (commented):
Nice work on the GCP integration — solid first implementation that follows all the established patterns. Found three test failures in `config/` that need fixing (integration count hardcoded to 13), plus the unbound context in `Configure()` could cause hangs during init. Rest is non-blocking polish. Details inline.

gcp/gcp.go
```diff
@@ -0,0 +1,358 @@
...
+
+func (g *integration) Name() string { return "gcp" }
+
+func (g *integration) Configure(creds mcp.Credentials) error {
+	g.projectID = creds["project_id"]
+	if g.projectID == "" {
+		return fmt.Errorf("gcp: project_id is required")
+	}
+
+	ctx := context.Background()
```
acmacalister: The 17 client initializations use `context.Background()`, so they can't be canceled during server shutdown. If GCP APIs are slow or unreachable, `Configure` will block until timeout (could be 30+ seconds per client). Probably fine since `Configure` runs at startup, but if we ever move to lazy init or reconfigure at runtime, it'll be a problem. Consider using `context.WithTimeout(context.Background(), 30*time.Second)` to bound the worst-case init time:

```go
ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
defer cancel()
```

gcp/firestore.go
```diff
@@ -0,0 +1,138 @@
...
+func firestoreQuery(ctx context.Context, g *integration, args map[string]any) (*mcp.ToolResult, error) {
+	collection := argStr(args, "collection")
+	q := g.firestoreClient.Collection(collection).Query
+
+	if field := argStr(args, "where_field"); field != "" {
+		op := argStr(args, "where_op")
+		valueStr := argStr(args, "where_value")
+		var value any
+		if err := json.Unmarshal([]byte(valueStr), &value); err != nil {
+			value = valueStr
```
acmacalister: If `json.Unmarshal` fails here, the raw string is used as the filter value. This means `where_value: "123"` becomes the string `"123"` instead of the number `123`, which won't match numeric fields. Consider adding examples to the tool description in tools.go to show the expected JSON format for different value types.

gcp/storage.go
```diff
@@ -0,0 +1,183 @@
...
+
+	mcp "github.com/daltoniam/switchboard"
+)
+
+const maxGCSGetObjectSize = 10 * 1024 * 1024
+
+func storageListBuckets(ctx context.Context, g *integration, _ map[string]any) (*mcp.ToolResult, error) {
+	var buckets []map[string]any
+	it := g.storageClient.Buckets(ctx, g.projectID)
+	for {
```
acmacalister: Unbounded `for` loop here vs. the bounded loop with `limit` parameter in `storageListObjects`. Both work, but the bounded loop (lines 65-80) is safer since it prevents unbounded memory growth for projects with thousands of buckets. Consider using the same pattern here with a reasonable default limit.

acmacalister (approved):
Nice work on the GCP adapter — follows project patterns well, all tests pass, and the existing review feedback has been addressed. Just a couple of minor suggestions inline.

gcp/gcp_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+func TestNew(t *testing.T) {
+	i := New()
+	require.NotNil(t, i)
+	assert.Equal(t, "gcp", i.Name())
+}
+
+func TestConfigure_MissingProjectID(t *testing.T) {
+	i := New()
+	err := i.Configure(context.Background(), mcp.Credentials{})
```
acmacalister: This test passes `context.Background()` to Configure, which means it won't exercise the new timeout logic you added. Worth using `context.WithTimeout(context.Background(), 5*time.Second)` so the test verifies Configure respects cancellation.

PR #19, opened 2026-02-28
feat(search): add pagination

arjansingh:
Instead of requesting all tools all at once (leading to a max token warning or rejection on some LLM consumers), switchboard now allows limit/offset discovery to tools via search.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -63,6 +63,8 @@ make generate
 
 1. **Run `make ci`** — must pass (build, vet, test-race, lint, security)
 2. **New code must include tests**
+3. **TDD**: write failing test before implementation, verify it fails for the right reason, then write minimal code to pass
+4. **Table-driven tests**: use `t.Run` subtests when 3+ cases share the same assert structure. Keep standalone tests for cases with unique setup or assertions
 
 ## Git Workflow
 


acmacalister:
Nice approach to the token budget problem — pagination with deterministic sort and a count-only mode is well thought out, and the test coverage is thorough. A few things to look at:

---

**`server/server.go:5`** — Import order isn't `gofmt`-compliant — `"cmp"` sorts before `"context"`. Swap these two so CI doesn't flag it.

```go
import (
	"cmp"
	"context"
```

---

**`server/server.go:238`** — Worth adding a `has_more` bool here so LLM consumers don't have to compute `offset + limit < total` themselves. It's a small thing but it matches the cursor/has_more pattern the Slack and Linear integrations already use, and makes the "should I fetch the next page?" decision trivial.

```go
type response struct {
	Summary string     `json:"summary"`
	Total   int        `json:"total"`
	Offset  int        `json:"offset"`
	Limit   int        `json:"limit"`
	HasMore bool       `json:"has_more"`
	Tools   []toolInfo `json:"tools"`
}
```

---

**`server/server.go:189`** — The capacity pre-calculation loop calls `integration.Tools()` for every enabled integration, then the main loop at line 191 calls it again. Most adapters allocate a new slice each call, so we're doubling that work. With ~500 tools this is negligible, but a plain `var all []toolInfo` would be simpler and let Go's append growth handle it — the sort pass dominates cost anyway.

---

**`server/server.go:255`** — The old response was `summary + "\n\n" + MarshalIndent([]toolInfo)` — a human-readable text blob with a pretty-printed JSON array. This is now a compact `json.Marshal` of a wrapper object with `total`, `offset`, `limit`, `tools`. Any existing consumer parsing the old format will break. Since we're pre-1.0 this is probably fine, but worth calling out in the PR description so people know what changed.

acmacalister (commented):
Nice approach to the token budget problem — pagination with deterministic sort and a count-only mode is well thought out, and the test coverage is thorough. A few things inline to look at, the only hard one is the import order that'll fail gofmt.

server/server.go
```diff
@@ -2,17 +2,21 @@ package server
 
 import (
 	"context"
+	"cmp"
```
acmacalister: Import order isn't `gofmt`-compliant — `"cmp"` sorts before `"context"`. Swap these two so CI doesn't flag it.

```go
import (
	"cmp"
	"context"
```

server/server.go
```diff
@@ -175,16 +207,52 @@ func (s *Server) handleSearch(ctx context.Context, req *mcpsdk.CallToolRequest)
...
+	}
+	page := all[offset:end]
+
+	type response struct {
+		Summary string     `json:"summary"`
+		Total   int        `json:"total"`
+		Offset  int        `json:"offset"`
+		Limit   int        `json:"limit"`
+		Tools   []toolInfo `json:"tools"`
+	}
```
acmacalister: Worth adding a `has_more` bool here so LLM consumers don't have to compute `offset + limit < total` themselves. It's a small thing but it matches the cursor/has_more pattern the Slack and Linear integrations already use, and makes the "should I fetch the next page?" decision trivial.

```go
type response struct {
	Summary string     `json:"summary"`
	Total   int        `json:"total"`
	Offset  int        `json:"offset"`
	Limit   int        `json:"limit"`
	HasMore bool       `json:"has_more"`
	Tools   []toolInfo `json:"tools"`
}
```

server/server.go
```diff
@@ -154,17 +179,24 @@ func (s *Server) handleSearch(ctx context.Context, req *mcpsdk.CallToolRequest)
...
 
-	var results []toolInfo
+	enabled := s.services.Config.EnabledIntegrations()
+	var capacity int
+	for _, name := range enabled {
+		if integration, ok := s.services.Registry.Get(name); ok {
+			capacity += len(integration.Tools())
+		}
+	}
+	all := make([]toolInfo, 0, capacity)
```
acmacalister: This capacity pre-calculation loop calls `integration.Tools()` for every enabled integration, then the main loop at line 191 calls it again. Most adapters allocate a new slice each call, so we're doubling that work. With ~500 tools this is negligible, but a plain `var all []toolInfo` would be simpler and let Go's append growth handle it — the sort pass dominates cost anyway.

server/server.go
```diff
@@ -175,16 +207,52 @@ func (s *Server) handleSearch(ctx context.Context, req *mcpsdk.CallToolRequest)
...
+		Total:   total,
+		Offset:  offset,
+		Limit:   limit,
+		Tools:   page,
+	})
+
 	return &mcpsdk.CallToolResult{
 		Content: []mcpsdk.Content{
-			&mcpsdk.TextContent{Text: summary + "\n\n" + string(data)},
+			&mcpsdk.TextContent{Text: string(data)},
```
acmacalister: The old response was `summary + "

" + MarshalIndent([]toolInfo)` — a human-readable text blob with a pretty-printed JSON array. This is now a compact `json.Marshal` of a wrapper object with `total`, `offset`, `limit`, `tools`. Any existing consumer parsing the old format will break.

Since we're pre-1.0 this is probably fine, but worth calling out in the PR description so people know what changed.

PR #24, opened 2026-03-01
using scripting to compose calls together to reduce tokens

daltoniam:


acmacalister:
Really nice feature — server-side JS scripting to chain tool calls and keep intermediate results out of the conversation is a big token savings win. Clean architecture with the Executor interface and the refactored `executeTool` method. A few notes inline below.

---

### `script/engine.go:67` — No script source size limit

There's no size check on `source` before handing it to goja. A multi-megabyte script string would consume memory during parsing before the timeout even kicks in. Worth adding an early bail-out:

```go
const maxScriptSize = 64 * 1024 // 64 KB
if len(source) > maxScriptSize {
    return &Result{
        Data:    fmt.Sprintf("script too large: %d bytes (max %d)", len(source), maxScriptSize),
        IsError: true,
    }, nil
}
```

### `script/engine.go:75` — Unbounded `console.log` accumulation

The `logs` slice grows without bound — `console.log` has no call-count limit (unlike `api.call`). A script doing `for(var i=0;i<1e7;i++) console.log(i)` could accumulate significant memory. Capping at ~100 entries or ~1 MB total would be a cheap safety net.

### `script/engine.go:57` — `script.Result` duplicates `mcp.ToolResult`

`script.Result` is structurally identical to `mcp.ToolResult` (both are `Data string` + `IsError bool`). Having `Engine.Run()` return `*mcp.ToolResult` directly would eliminate this type and the conversion in `handleScriptExecute`.

### `script/engine.go:137` — `console.log` string building

This can be simplified — `parts` is already built above:

```go
line := strings.Join(parts, " ")
```

(Would need to add `"strings"` to the import block.)

### `server/server.go:102` — Schema hints for mutual exclusivity

Neither field is schema-required now, which is correct since the two modes are mutually exclusive. The handler validates this at runtime. One thing that might help LLM clients pick the right mode: a note in each field description like `"(mutually exclusive with script)"` / `"(mutually exclusive with tool_name)"`.

### `server/server.go:220` — Missing test for empty-args error path

This error path (both `script` and `tool_name` empty) doesn't have test coverage. Worth adding a small test that calls `handleExecute` with empty args and asserts the `"either tool_name or script is required"` message.

### Formatting

`gofmt -l` flags `script/engine_test.go` and `server/server_test.go` — minor alignment diffs. `gofmt -w` on both files will fix it.

acmacalister (commented):
Really nice feature — server-side JS scripting to chain tool calls and keep intermediate results out of the conversation is a big token savings win. Clean architecture with the Executor interface and the refactored executeTool method. A few inline comments, mostly around resource limits on the script engine.

PR #27, opened 2026-03-02
feat(slack): add workspace selector for Chrome extraction 

daltoniam:
also fix JS snippet

Chrome auto-extract now lists all available workspaces and lets the user pick which one to extract tokens from, instead of grabbing the first one found. The browser extraction JavaScript snippet is rewritten to handle localConfig v2–v5, use localStorage.getItem() for reliability, avoid arrow functions for compatibility, and prompt the user to choose when multiple workspaces are present.


web/templates/pages/slack_setup.templ
```diff
@@ -114,9 +114,73 @@ templ SlackSetup(page layouts.PageData, data SlackSetupData) {
...
+									form.appendChild(input);
+									form.submit();
+									return;
+								}
+								var html = '';
+								for (var i = 0; i < ws.length; i++) {
+									var label = ws[i].name || ws[i].team_id;
+									if (ws[i].url) { label += ' (' + ws[i].url.replace('https://', '') + ')'; }
+									html += '<label style="display: flex; align-items: center; gap: 0.5rem; padding: 0.5rem 0; cursor: pointer;">';
+									html += '<input type="radio" name="team_id" value="' + ws[i].team_id + '"' + (i === 0 ? ' checked' : '') + '>';
```
acmacalister: Workspace names, URLs, and team IDs from Chrome's LevelDB are injected into the DOM via `innerHTML` and string-concatenated HTML attributes here. If a workspace name contains HTML (e.g. ``), it'll execute. The `team_id` value is also concatenated directly into the `value` attribute without escaping, so a crafted ID could break out of the attribute.

Practical risk is low since the data comes from the local user's own Chrome, but it's still untrusted third-party data. Using DOM APIs instead of `innerHTML` closes this off:

```javascript
var container = document.getElementById('workspace-list');
container.innerHTML = '';
for (var i = 0; i < ws.length; i++) {
    var labelEl = document.createElement('label');
    labelEl.style.cssText = 'display: flex; align-items: center; gap: 0.5rem; padding: 0.5rem 0; cursor: pointer;';
    var input = document.createElement('input');
    input.type = 'radio';
    input.name = 'team_id';
    input.value = ws[i].team_id;
    if (i === 0) input.checked = true;
    var span = document.createElement('span');
    var text = ws[i].name || ws[i].team_id;
    if (ws[i].url) { text += ' (' + ws[i].url.replace('https://', '') + ')'; }
    span.textContent = text;
    labelEl.appendChild(input);
    labelEl.appendChild(span);
    container.appendChild(labelEl);
}
```

`textContent` auto-escapes HTML and `input.value` sets the DOM property rather than being parsed as markup.

slack/tokens.go
```diff
@@ -266,29 +275,91 @@ func extractTokenFromLevelDB(profilePath string) (string, error) {
 		Strict:   opt.NoStrict,
 	})
 	if err != nil {
-		return "", fmt.Errorf("opening LevelDB copy: %w", err)
+		return nil, fmt.Errorf("opening LevelDB copy: %w", err)
 	}
 	defer func() { _ = db.Close() }()
 
 	key := "_https://app.slack.com\x00\x01localConfig_v2"
```
acmacalister: This hardcodes `localConfig_v2` as the only LevelDB key, but the JS snippet in `extract.go` already iterates v2 through v5. If Slack migrates a user's localStorage to a newer version, the Go-side Chrome extraction and workspace listing will silently fail while the browser console snippet still works — which would be a confusing experience.

Worth iterating here to match the JS snippet:

```go
versions := []string{"localConfig_v2", "localConfig_v3", "localConfig_v4", "localConfig_v5"}
var val []byte
for _, v := range versions {
    key := "_https://app.slack.com" + v
    if found, err := db.Get([]byte(key), nil); err == nil {
        val = found
        break
    }
}
if val == nil {
    return nil, fmt.Errorf("key not found in profile %s", filepath.Base(profilePath))
}
```

slack/extract_test.go
```diff
@@ -0,0 +1,61 @@
+package slack
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestExtractionSnippet_NotEmpty(t *testing.T) {
```
acmacalister: These tests cover the snippet content and struct fields nicely, but the core new logic (`listWorkspacesFromChrome`, `readSlackLocalConfig`, the `teamID` filtering in `extractTokenFromLevelDB`) doesn't have behavioral tests yet. The most impactful addition would be testing `ExtractFromChromeForWeb` and `ListWorkspacesFromChrome` on non-darwin to verify the platform guard, and ideally `extractTokenFromLevelDB` with a synthetic LevelDB fixture to cover the team-ID-specific extraction path.

PR #39, opened 2026-03-08
Add Home Assistant integration

aleksclark:
## Summary

- Add Home Assistant REST API integration with 18 tools covering entity states, service calls, events, history, logbook, config, template rendering, error log, calendars, and intent handling
- Implements `Integration`, `FieldCompactionIntegration`, and `PlainTextCredentials` interfaces with bearer token auth against a user-provided base URL
- 38 tests at 82% coverage including dispatch parity, HTTP helpers, argument coercion, and handler integration tests

## Test plan

- [x] `go test ./integrations/homeassistant/` — 38 tests pass
- [x] `go test ./config/` — updated config tests for 13th integration
- [x] `go test ./server/` — smoke test passes with new adapter registered
- [x] `go build ./...` — compiles cleanly
- [x] `go vet ./...` — no issues

🐾 Generated with Crush

### DIFF
--- a/integrations/homeassistant/events.go
+++ b/integrations/homeassistant/events.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/url"
 
 	mcp "github.com/daltoniam/switchboard"
 )
@@ -31,7 +32,7 @@ func fireEvent(ctx context.Context, h *homeassistant, args map[string]any) (*mcp
 		body = data
 	}
 
-	result, err := h.post(ctx, fmt.Sprintf("/api/events/%s", eventType), body)
+	result, err := h.post(ctx, fmt.Sprintf("/api/events/%s", url.PathEscape(eventType)), body)
 	if err != nil {
 		return errResult(err)
 	}


acmacalister (commented):
Excellent work on the Home Assistant integration — clean structure, thorough tests, and idiomatic Go throughout. Just one pattern to address: user-provided path segments need URL escaping to prevent path traversal. Six spots inline where `entity_id` or `start_time` should use `url.PathEscape()` before being interpolated into paths.

integrations/homeassistant/events.go
```diff
@@ -0,0 +1,39 @@
...
+	var body any
+	if v := argStr(args, "event_data"); v != "" {
+		var data map[string]any
+		if err := json.Unmarshal([]byte(v), &data); err != nil {
+			return errResult(fmt.Errorf("invalid JSON for event_data: %w", err))
+		}
+		body = data
+	}
+
+	result, err := h.post(ctx, fmt.Sprintf("/api/events/%s", eventType), body)
```
acmacalister: `event_type` is user-controlled and goes straight into the URL path — same issue as the others. Should escape it.

```go
result, err := h.post(ctx, fmt.Sprintf("/api/events/%s", url.PathEscape(eventType)), body)
```

PR #41, opened 2026-03-09
feat(mcp): add context.Context to Configure interface

arjansingh:
Just threading the context into the `Configure` interface.

### DIFF
--- a/integrations/aws/aws.go
+++ b/integrations/aws/aws.go
@@ -47,7 +47,7 @@ func New() mcp.Integration {
 
 func (a *integration) Name() string { return "aws" }
 
-func (a *integration) Configure(creds mcp.Credentials) error {
+func (a *integration) Configure(ctx context.Context, creds mcp.Credentials) error {
 	region := creds["region"]
 	if region == "" {
 		region = "us-east-1"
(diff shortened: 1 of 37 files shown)


integrations/aws/aws.go
```diff
@@ -66,7 +66,7 @@ func (a *integration) Configure(creds mcp.Credentials) error {
 		))
 	}
 
-	cfg, err := awsconfig.LoadDefaultConfig(context.Background(), opts...)
+	cfg, err := awsconfig.LoadDefaultConfig(ctx, opts...)
```
acmacalister: Nice—AWS SDK's `LoadDefaultConfig` respects cancellation and deadlines, so this context flows through credential provider chains and IMDS calls. Real value on the first use.

integrations/notion/notion.go
```diff
@@ -56,7 +56,7 @@ func (n *notion) Configure(creds mcp.Credentials) error {
 		n.baseURL = strings.TrimRight(v, "/")
 	}
 
-	spaceID, userID, err := n.resolveSpaceAndUser()
+	spaceID, userID, err := n.resolveSpaceAndUser(ctx)
```
acmacalister: Good propagation. The `resolveSpaceAndUser` call now respects cancellation during the initial `getSpaces` API call.

server/server.go
```diff
@@ -170,7 +170,7 @@ func (s *Server) configureIntegrations() {
 			continue
 		}
 
-		if err := integration.Configure(ic.Credentials); err != nil {
+		if err := integration.Configure(context.Background(), ic.Credentials); err != nil {
```
acmacalister: Minor: still using `context.Background()` here. Not a blocker since server startup is typically unbounded, but if you ever add a startup timeout or graceful boot shutdown, this would be the spot to thread it through.

PR #68, opened 2026-03-19
feat: add Twitter/X integration

daltoniam:
## Summary

- Add full Twitter/X API v2 integration with 55 tools covering tweets, users, lists, DMs, spaces, and usage
- OAuth 2.0 PKCE browser-based auth flow with manual bearer token fallback
- Field compaction specs for 33 read tools to minimize token usage
- Web UI setup page with OAuth credential management, "Sign in with X" button, and available tools display
- 42 unit tests covering handlers, HTTP helpers, dispatch parity, and compaction specs

## New files

- `integrations/twitter/` — 11 source files (adapter, tools, handlers, OAuth, compaction specs, tests)
- `web/templates/pages/twitter_setup.templ` — Setup page template

## Modified files

- `cmd/server/main.go` — Register twitter adapter
- `config/config.go` — Add default credentials and env var mapping
- `config/config_test.go` — Update integration count assertions
- `web/web.go` — Add setup routes and 5 handler functions

## Test plan

- [x] `go build ./...` passes
- [x] `go test -race ./...` passes (42 new tests)
- [x] `go vet ./...` clean
- [x] `go tool golangci-lint run` — 0 issues
- [x] Dispatch parity tests enforce all 55 tools have handlers
- [x] Compaction spec tests enforce no orphan/missing specs

🐮 Generated with Crush


integrations/x/x.go
```diff
@@ -0,0 +1,328 @@
...
+func (t *xClient) del(ctx context.Context, pathFmt string, args ...any) (json.RawMessage, error) {
+	return t.doRequest(ctx, "DELETE", fmt.Sprintf(pathFmt, args...), nil)
+}
+
+// --- Result helpers ---
+
+type handlerFunc func(ctx context.Context, t *xClient, args map[string]any) (*mcp.ToolResult, error)
+
+func rawResult(data json.RawMessage) (*mcp.ToolResult, error) {
+	return &mcp.ToolResult{Data: string(data)}, nil
```
acmacalister: This local `rawResult` wrapper duplicates `mcp.RawResult` — and `json.RawMessage` is already `[]byte`, so you can call the shared helper directly. AGENTS.md marks this MUST: use `mcp.RawResult(data)` and `mcp.ErrResult(err)` instead of defining per-adapter copies.

Same applies to `errResult` below — `mcp.ErrResult` already handles the retryable-error propagation identically to what's here.

integrations/x/x.go
```diff
@@ -0,0 +1,328 @@
...
+func errResult(err error) (*mcp.ToolResult, error) {
+	if mcp.IsRetryable(err) {
+		return nil, err
+	}
+	return &mcp.ToolResult{Data: err.Error(), IsError: true}, nil
+}
+
+// --- Argument helpers ---
+
+func argStr(args map[string]any, key string) string {
```
acmacalister: AGENTS.md MUST: never define local `argStr`/`argInt`/`argBool` in adapters — use `mcp.NewArgs(args)` or the standalone `mcp.ArgStr`/`mcp.ArgInt`/`mcp.ArgBool` from `args.go` instead. The local helpers also bypass the `TestNewArgs_ErrCheckParity` parity test that all other adapters get automatically.

For example, handlers that need a string arg would become:
```go
r := mcp.NewArgs(args)
id := r.Str("id")
    return mcp.ErrResult(err)
}
```
Or using the standalone helper: `id, err := mcp.ArgStr(args, "id")`

integrations/x/oauth.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+func HandleXCallback(code, state string) error {
+	oauthState := getActiveOAuth()
+
+	if oauthState == nil {
+		return fmt.Errorf("no OAuth flow in progress")
+	}
+
+	oauthState.mu.Lock()
```
acmacalister: `HandleXCallback` holds `oauthState.mu.Lock()` while making the token exchange HTTP request to `xTokenURL` (lines 151-155). If the X API is slow, any concurrent call to `PollXOAuth` will block for the duration of the request.

Since this is a UI-only setup flow the contention is unlikely to matter in practice, but it's worth noting. The lock could be released before the HTTP call and re-acquired to update the result fields, matching the pattern Gmail's OAuth uses.

integrations/x/oauth.go
```diff
@@ -0,0 +1,279 @@
...
+	_, _ = rand.Read(b)
+	return base64.RawURLEncoding.EncodeToString(b)[:n]
+}
+
+func oauthPKCEChallenge(verifier string) string {
+	h := sha256.Sum256([]byte(verifier))
+	return base64.RawURLEncoding.EncodeToString(h[:])
+}
+
+func StartXOAuth(clientID, clientSecret, redirectURI string) (*OAuthStartResult, error) {
```
acmacalister: The OAuth helper functions (`StartXOAuth`, `HandleXCallback`, `PollXOAuth`, `RefreshAccessToken`) have no unit tests. The gmail adapter covers equivalent paths with `httptest.NewServer` — e.g. `TestHandleGmailCallback_TokenExchange` overrides the token URL so the test doesn't hit the real endpoint.

The main thing missing here is that the hardcoded `xTokenURL` constant makes it impossible to inject a test server without exporting it or adding an option. Adding a `tokenURL` field to `OAuthState` (defaulting to the constant) would unlock this without touching the public API.

acmacalister (commented):
Really solid integration — 55 tools, full compaction specs, dispatch parity tests, and an OAuth PKCE flow all in one PR. Two AGENTS.md convention violations to sort out (shared result + arg helpers), and a couple smaller notes inline.

integrations/x/tools.go
```diff
@@ -0,0 +1,498 @@
...
+			"next_token":   "Pagination token from previous response",
+			"tweet_fields": "Comma-separated tweet fields: author_id,created_at,public_metrics,source,text,conversation_id,entities",
+			"start_time":   "Oldest UTC datetime (ISO 8601: 2024-01-01T00:00:00Z)",
+			"end_time":     "Newest UTC datetime (ISO 8601)",
+			"sort_order":   "Order of results: recency or relevancy",
+		},
+		Required: []string{"query"},
+	},
+	{
+		Name: "x_search_all", Description: "Full-archive search across all tweets (requires Pro tier or higher). Use twitter_search_recent for 7-day window instead",
```
acmacalister: The description references `twitter_search_recent` but the actual tool is `x_search_recent`. Same stale name appears on line 59 in the `query` param description. An LLM following the cross-reference in `x_search_all` will look for a tool that doesn't exist.

Lines 102, 104, and 140 have the same issue — `twitter_get_user`, `twitter_get_user_by_username` should be `x_get_user` / `x_get_user_by_username`.

integrations/x/x.go
```diff
@@ -0,0 +1,286 @@
...
+	me, err := t.fetchMe()
+	if err != nil {
+		return fmt.Errorf("x: failed to resolve authenticated user: %v", err)
+	}
+	t.userID = me
+	return nil
+}
+
+func (t *xClient) fetchMe() (string, error) {
+	req, err := http.NewRequest("GET", t.baseURL+"/users/me", nil)
```
acmacalister: `fetchMe` constructs the request with `http.NewRequest` instead of `http.NewRequestWithContext`, so the context passed to `Configure` is silently dropped here. If `Configure` is called with a context that carries a deadline or cancellation, this network call won't respect it.

```go
func (t *xClient) fetchMe(ctx context.Context) (string, error) {
    req, err := http.NewRequestWithContext(ctx, "GET", t.baseURL+"/users/me", nil)
```

Then thread `ctx` from `Configure` through the call.

integrations/x/oauth.go
acmacalister, line 124: `HandleXCallback` acquires `oauthState.mu` and then makes the token exchange HTTP request while holding the lock (lines 139–156). If the X token endpoint is slow or unreachable, `PollXOAuth` will block for the full duration of that request.

Since this is a low-concurrency UI setup flow it won't cause problems in practice, but releasing the lock before the HTTP call and re-acquiring it to write `accessToken`/`refreshToken`/`done` is the safer pattern — it's how the Gmail OAuth adapter handles it too.

integrations/x/oauth.go
acmacalister, line 68: `StartXOAuth`, `HandleXCallback`, `PollXOAuth`, and `RefreshAccessToken` have no unit tests. The Gmail adapter covers the equivalent code paths using `httptest.NewServer` — e.g. `TestHandleGmailCallback_TokenExchange` overrides the token URL so tests don't hit the real endpoint.

The blocker here is that `xTokenURL` is a package-level constant, so there's no hook to inject a test server. Adding a `tokenURL` field to `OAuthState` (defaulting to `xTokenURL`) would unblock testing without changing the public API.

acmacalister (commented):
Good work overall — the adapter is well-structured, uses the shared helpers correctly, all 42 tests pass with the race detector, dispatch parity is enforced, and the compaction specs look solid. A few things worth addressing inline.

integrations/x/x.go
```diff
@@ -0,0 +1,286 @@
...
+	if t.bearerToken == "" {
+		return fmt.Errorf("x: bearer_token is required")
+	}
+	if v := creds["base_url"]; v != "" {
+		t.baseURL = strings.TrimRight(v, "/")
+	}
+
+	me, err := t.fetchMe(ctx)
+	if err != nil {
+		return fmt.Errorf("x: failed to resolve authenticated user: %v", err)
```
acmacalister, line 52: This wraps the error with `%v` instead of `%w`, so the original error type is lost. Callers using `errors.Is()` or `errors.As()` to inspect the underlying error (e.g. to distinguish a network timeout from an auth failure) won't be able to see through this wrapper.

```go
return fmt.Errorf("x: failed to resolve authenticated user: %w", err)
```

integrations/x/users.go
```diff
@@ -0,0 +1,242 @@
...
+func getUserByUsername(ctx context.Context, t *xClient, args map[string]any) (*mcp.ToolResult, error) {
+	r := mcp.NewArgs(args)
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	username := r.Str("username")
+	q := queryEncode(map[string]string{
+		"user.fields": r.Str("user_fields"),
+	})
+	data, err := t.get(ctx, "/users/by/username/%s%s", username, q)
```
acmacalister, line 35: The `username` value is interpolated directly into the URL path via `fmt.Sprintf`. X usernames are constrained to `[A-Za-z0-9_]` by the platform so this is safe in practice, but wrapping with `url.PathEscape` would be consistent with how the `salesforce` adapter handles user-supplied path segments and removes any ambiguity:

```go
data, err := t.get(ctx, "/users/by/username/%s%s", url.PathEscape(username), q)
```

Same applies to `id` in `getUser` and the other handlers that put a raw `r.Str(...)` value directly into a `%s` path segment.

acmacalister (approved):
Nice work on this one — 55 tools, full compaction specs, PKCE OAuth, and the test coverage is solid. Previous review feedback has all been addressed. Two small things inline.

PR #69, opened 2026-03-20
Add SWITCHBOARD_PORT environment variable for configurable port

pzipoy:
## Summary

Allow the server port to be configured via `SWITCHBOARD_PORT` environment variable, enabling multiple instances to run on different ports (e.g., for different user profiles on the same machine).

## Changes

- Add `defaultPort()` helper that checks `SWITCHBOARD_PORT` env var before falling back to 3847
- Log warning if `SWITCHBOARD_PORT` is set but contains an invalid value
- Update both main server and daemon commands to use `defaultPort()`
- Make dashboard footer display the actual configured port dynamically

## Usage

```bash
# Use default port 3847
switchboard

# Use custom port
SWITCHBOARD_PORT=3848 switchboard

# Flag still takes precedence
SWITCHBOARD_PORT=3848 switchboard --port 4000  # uses 4000
```

## Testing

- `make test` passes
- `make lint` passes  
- Manual testing confirms env var works and invalid values log a warning


acmacalister (approved):
Nice work — the feature is clean, builds and passes all tests with no lint issues. A couple of small things inline worth considering before merge.

acmacalister (commented):
Solid implementation — the portutil extraction and 7-case test suite address all the prior feedback cleanly. A few things worth looking at inline.

PR #77, opened 2026-03-21
Switchboard remote hosted

acmacalister:


acmacalister (commented):
Solid scaffolding for the multi-tenant hosted path — the new packages (audit, authn, tenant, integrationcache) are well-structured, the dual-mode local/hosted design is clean, and the per-tenant instance cache with TTL+LRU is a good foundation. A few things need attention before this ships: a data race in the cache, a non-deterministic hash that will break cache reuse under concurrency, and some coverage gaps for the new hosted-mode code paths.

PR #106, opened 2026-04-08
Add ACP integration for multi-agent communication

aleksclark:
## Summary

- Add ACP (Agent Communication Protocol) client integration that connects to remote ACP servers to discover and invoke AI agents
- Supports both pre-configured servers (via JSON config credential / `ACP_CONFIG` env var) and inline `server_url`/`server_headers` parameters for dynamic server targeting by the calling MCP client
- Three tools: `acp_list_agents`, `acp_run_agent`, `acp_resume_run` — stream-first with sync fallback, NDJSON event parsing, multi-turn sessions, and awaiting/resume flow

## Test plan

- [x] 46 unit tests covering all handlers, client HTTP, inline server URL, headers, stream/sync modes, awaiting/resume, multi-server, type helpers
- [x] Dispatch map parity tests (all tools covered, no orphans)
- [x] Configure success/failure tests including empty config (optional credentials)
- [x] `make build`, `make vet`, `make test-race`, `make lint`, `make security` all pass

🐙 Generated with [Crush](https://charm.sh/crush)

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -33,7 +33,7 @@ func TestLoad_CreatesDefaultWhenMissing(t *testing.T) {
 	assert.NoError(t, err)
 
 	assert.Len(t, m.cfg.Integrations, 31)
-	for _, name := range []string{"github", "datadog", "linear", "sentry", "slack", "metabase", "aws", "posthog", "postgres", "clickhouse", "elasticsearch", "pganalyze", "rwx", "gmail", "notion", "ynab", "gcp", "suno", "amazon", "jira", "confluence", "readarr", "salesforce", "cloudflare", "digitalocean", "fly", "snowflake", "web", "botidentity", "x"} {
+	for _, name := range []string{"github", "datadog", "linear", "sentry", "slack", "metabase", "aws", "posthog", "postgres", "clickhouse", "elasticsearch", "pganalyze", "rwx", "gmail", "notion", "ynab", "gcp", "suno", "amazon", "jira", "confluence", "readarr", "salesforce", "cloudflare", "digitalocean", "fly", "snowflake", "acp", "web", "botidentity", "x"} {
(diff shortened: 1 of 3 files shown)


acmacalister (commented):
Nice integration — clean architecture, solid test coverage (46 tests, all passing with `-race`), and the stream-first-with-sync-fallback design is well-executed. Two things worth looking at inline.

acmacalister (commented):
Build passes, tests pass (46 tests, all green), lint clean — solid foundation. Two things from the ACP files worth addressing before this ships widely.

In `integrations/acp/acp.go:209`, `session_id` is extracted with a standalone `mcp.ArgStr` call after the `r.Err()` guard — the previous reviewer flagged this already and it's still unaddressed. The fix is a one-liner: pull `sessionID := r.Str("session_id")` into the scanner block before the `r.Err()` check.

In `integrations/acp/client.go`, `parseStream` spawns a goroutine that has no context cancellation path. If the caller's context is cancelled, the goroutine can only exit once the HTTP transport closes the underlying connection — which for `streamClient` (Timeout: 0) means it depends entirely on the server or TCP layer. The previous reviewer suggested passing `ctx` into `parseStream` and closing `r` (or returning early from the scan loop) when `ctx.Done()` fires. There's also a related issue in `collectStreamResponse`: the `EventError` early return stops draining the channel but leaves the goroutine running and potentially blocked on a full channel until the body closes. Plumbing `ctx` into `parseStream` fixes both.

The config_test.go change itself (adding `"acp"` to the integration list) is correct — count goes to 31, entry is present in config defaults.

acmacalister (approved):
Both previous review issues are cleanly addressed — session_id now lives inside the NewArgs scanner block where it belongs, and parseStream properly propagates context. Regression test for the session_id fix is a nice touch. Build passes, all 46 tests pass with race detector, lint clean. LGTM.

PR #109, opened 2026-04-09
Add multi-postgres connection support

daltoniam:
## Summary

- Add multi-database connection support to the Postgres integration using the Slack multi-workspace pattern (internal multiplexing). A single `"postgres"` registry entry holds a `map[string]*pgConn` with alias-based routing — no framework or registry changes needed.
- All 24 existing tools gain an optional `database` parameter to target a specific connection, with fallback to the default. New `postgres_list_databases` tool returns configured aliases with metadata.
- Custom `postgres_setup.templ` web UI page with default connection form and dynamic add/remove for additional named connections, serialized as JSON on save.

## Changes

| Area | Files | What |
|------|-------|------|
| Adapter core | `postgres.go` | `pgConn` struct, `getConnForArgs()` router, `openConn()`/`buildConnStr()`, multi-conn `Configure()`/`Close()`/`Healthy()` |
| Tools | `tools.go` | `database` param on all tools, `postgres_list_databases` tool + dispatch |
| Handlers | `queries.go`, `databases.go`, `management.go` | All 25 handlers route via `getConnForArgs()` |
| Compaction | `compact_specs.go` | Spec for `postgres_list_databases` |
| Config | `config/config.go` | `connections` credential key |
| Web UI | `postgres_setup.templ`, `web.go` | Custom setup page, setup routes, setup-aware redirects |
| Tests | `postgres_test.go` | Multi-conn validation, routing, dispatch parity, `listDatabases` |

## Backward Compatibility

Fully backward compatible. Existing single-connection configs work unchanged — the primary connection becomes the `"default"` alias automatically.

## Test plan

- [x] `make vet` passes
- [x] `make lint` passes (0 issues)
- [x] `go test ./...` passes (all packages)
- [x] Dispatch parity tests pass (`TestDispatchMap_AllToolsCovered`, `TestDispatchMap_NoOrphanHandlers`)
- [x] Compaction spec parity tests pass
- [x] Multi-connection validation tests (invalid JSON, missing alias, reserved alias, duplicates)
- [x] Connection routing tests (`getConnForArgs` default/explicit/unknown)
- [x] `listDatabases` handler test
- [ ] Manual: configure two Postgres databases and verify routing via `database` param
- [ ] Manual: verify web UI add/remove connections and save round-trip

🐘 Generated with Crush


integrations/postgres/compact_specs.go
```diff
@@ -7,6 +7,9 @@ import (
 )
 
 var rawFieldCompactionSpecs = map[string][]string{
+	// ── Connection Management ────────────────────────────────────────
+	"postgres_list_databases": {"alias", "host", "database", "read_only", "is_default"},
```
acmacalister: The postgres adapter has `compact_specs.go` with 24 entries but no `compact_specs_test.go`. Both `pganalyze` and `snowflake` have dedicated compaction spec tests — postgres should too. Without them, a future spec added for the wrong tool name (or a typo) would be silently ignored.

A `compact_specs_test.go` alongside this file covering the four standard cases (`AllParse`, `NoDuplicateTools`, `NoOrphanSpecs`, `ReturnsFieldsForListTool`) would give the same coverage guarantee:

```go
func TestFieldCompactionSpecs_NoOrphanSpecs(t *testing.T) {
    for toolName := range fieldCompactionSpecs {
        _, ok := dispatch[toolName]
        assert.True(t, ok, "field compaction spec for %q has no dispatch handler", toolName)
    }
}

func TestFieldCompactionSpec_ReturnsFieldsForListDatabases(t *testing.T) {
    p := &postgres{}
    fields, ok := p.CompactSpec("postgres_list_databases")
    require.True(t, ok)
    assert.NotEmpty(t, fields)
}
```

web/templates/pages/postgres_setup.templ
```diff
@@ -0,0 +1,229 @@
...
+			</div>
+		</div>
+		<div class="form-group">
+			<label class="form-label">read_only</label>
+			<input class="form-input" type="text" data-field="read_only" value={ conn.ReadOnly } placeholder="true"/>
+		</div>
+	</div>
+}
+
+func pgConnCount(conns []PostgresConnection) string {
```
acmacalister, line 221: `pgConnCount` is defined here but never called anywhere in the template or its generated counterpart. If it was meant to render a summary (e.g. below the connections list), this is the right place to wire it in — otherwise it can be removed to keep the generated `postgres_setup_templ.go` clean.

web/web.go
```diff
@@ -854,6 +865,55 @@ func (w *WebServer) handleNotionSaveToken(rw http.ResponseWriter, r *http.Reques
...
 
+func (w *WebServer) handlePostgresSetup(rw http.ResponseWriter, r *http.Request) {
+	ic, exists := w.services.Config.GetIntegration("postgres")
+	enabled := exists && ic.Enabled
+
+	var healthy bool
+	if enabled {
+		integration, ok := w.services.Registry.Get("postgres")
+		if ok {
+			if err := integration.Configure(r.Context(), ic.Credentials); err == nil {
```
acmacalister, line 1163: Calling `Configure` on every GET request to the setup page is consistent with how Slack, Notion, and GitHub setup handlers work — so this isn't a new pattern. Worth keeping in mind that with multi-postgres, each page load now opens connections to *all* configured databases (not just one), including a 5-second ping per connection. If someone has several slow secondaries, the setup page could hang noticeably. Not a blocker given the established pattern, but something to revisit if users report slow setup page loads.

acmacalister (approved):
Solid work — the two-phase configure (validate all aliases before opening any connections, then atomically swap) is exactly the right approach, and the backward-compat story is clean. Two small things inline worth addressing before merge.

integrations/postgres/postgres_test.go
```diff
@@ -20,15 +20,18 @@ func TestNew(t *testing.T) {
 }
 
 func TestConfigure_ConnectionString(t *testing.T) {
-	p := &postgres{}
+	p := &postgres{conns: make(map[string]*pgConn)}
 	err := p.Configure(context.Background(), mcp.Credentials{"connection_string": "host=localhost port=5432 user=test dbname=testdb sslmode=disable"})
 	assert.Error(t, err)
 	assert.Contains(t, err.Error(), "failed to ping")
-	assert.True(t, p.readOnly)
+	conn, ok := p.conns["default"]
```
acmacalister, line 27: The `if ok` guard means the `conn.readOnly` assertion is never actually reached — after a failed configure, `p.conns` stays empty so `ok` is always false. The original test had `assert.True(t, p.readOnly)` which directly verified the safe-default behavior. Worth asserting this unconditionally:

```go
require.Error(t, err)
assert.Contains(t, err.Error(), "failed to ping")
assert.Empty(t, p.conns, "failed configure should leave conns empty")
```

Or if you want to verify the read-only default, open the conn struct manually rather than relying on a failed ping populating it.

integrations/postgres/tools.go
```diff
@@ -2,92 +2,113 @@ package postgres
 
 import mcp "github.com/daltoniam/switchboard"
 
+const databaseParamDesc = "Connection alias (omit to use default). Use postgres_list_databases to see available connections."
+
 var tools = []mcp.ToolDefinition{
+	// --- Connection Management ---
+	{
+		Name:        "postgres_list_databases",
+		Description: "List all configured database connections with their alias, host, database name, and read-only status",
```
acmacalister, line 11: Per the project convention, the primary tool for an integration should include "Start here" text for LLM wayfinding (see AGENTS.md). With multi-postgres, `postgres_list_databases` is now the natural first call — it tells an LLM what connections are available before issuing any schema or query tool. Consider updating the description:

```go
Description: "List all configured database connections with alias, host, and read-only status. Start here when multiple databases are configured to find the right connection alias.",
```

acmacalister (commented):
Solid implementation — the two-phase configure, per-connection readOnly enforcement, and test coverage are all well done. Two small things inline.

web/templates/pages/postgres_setup.templ
```diff
@@ -0,0 +1,229 @@
...
+
+import (
+	"fmt"
+	"github.com/daltoniam/switchboard/web/templates/layouts"
+	"github.com/daltoniam/switchboard/web/templates/components"
+)
+
+type PostgresConnection struct {
+	Alias            string
+	ConnectionString string
```
acmacalister, line 11: `PostgresConnection` has no JSON tags, so `json.Unmarshal` in `handlePostgresSetup` can't map `connection_string` → `ConnectionString` or `read_only` → `ReadOnly` (Go's default decoder is case-insensitive for single-word fields but won't match snake_case to CamelCase). In practice this means a user who saves additional connections and then reloads the setup page will see those two fields blank — they'd have to re-enter them on every edit.

Fixing it is one line per field:

```go
type PostgresConnection struct {
    Alias            string `json:"alias"`
    ConnectionString string `json:"connection_string"`
    Host             string `json:"host"`
    Port             string `json:"port"`
    User             string `json:"user"`
    Password         string `json:"password"`
    Database         string `json:"database"`
    SSLMode          string `json:"sslmode"`
    ReadOnly         string `json:"read_only"`
}
```

The generated `postgres_setup_templ.go` will also need to be regenerated via `templ generate` after the change.

integrations/postgres/postgres.go
```diff
@@ -127,10 +281,32 @@ func (p *postgres) Execute(ctx context.Context, toolName mcp.ToolName, args map[
 	return fn(ctx, p, args)
 }
 
+// getConnForArgs resolves the connection to use based on the optional "database" arg.
+func (p *postgres) getConnForArgs(args map[string]any) (*pgConn, error) {
```
acmacalister, line 285: `getConnForArgs` acquires `p.mu.RLock()`, reads the `*pgConn` pointer from the map, releases the lock, then returns the pointer. Every caller then uses `conn.db` outside the lock. If `Configure()` runs concurrently (e.g. via the hot-reload `PUT /api/integrations/postgres/credentials` endpoint), it swaps `p.conns` under the write lock and then closes the old `*pgConn.db` — but handlers that already received the old pointer will race against that close.

The Slack multi-workspace adapter uses the same pattern, so this is an inherited trade-off rather than something introduced here. It's worth noting in a comment (or a follow-up issue) that `pgConn.db` is assumed to outlive any in-flight requests against the old connection — which is typically true in practice because `sql.DB` is internally reference-counted, but it's not guaranteed and the race detector won't catch it unless both sides run concurrently in the same test.

No immediate action needed if the pattern is accepted project-wide, but worth documenting the assumption.

acmacalister (commented):
Solid approach — clean two-phase Configure, backward-compatible, good test coverage. Found one concrete UI bug with JSON tags on `PostgresConnection` and a concurrency note worth documenting.

PR #110, opened 2026-04-09
Add WASM modules management page to web UI

daltoniam:
## Summary

- Add a new **WASM Modules** page (`/wasm`) to the web config UI so users can add, view, and remove WASM modules without hand-editing `config.json`
- Add `SetWasmModules` method to `ConfigService` interface for atomic, mutex-protected persistence of the modules list
- Includes sidebar nav entry, dynamic credential rows via JS, delete-with-confirmation, and a restart-required banner (modules load at boot)

## Test plan

- [x] `TestSetWasmModules` — persists modules to disk and reads back correctly
- [x] `TestSetWasmModules_EmptySlice` — clearing modules works
- [x] `TestSetWasmModules_PreservedOnReload` — survives config reload from disk
- [x] `TestWasmModulesPage_Empty` — renders empty state
- [x] `TestWasmModulesPage_WithModules` — renders module list with credential counts
- [x] `TestWasmModuleAdd` — adds module with credentials
- [x] `TestWasmModuleAdd_EmptyPath` — rejects empty path
- [x] `TestWasmModuleAdd_NoCreds` — adds module without credentials
- [x] `TestWasmModuleDelete` — removes correct module by index
- [x] `TestWasmModuleDelete_InvalidIndex` — rejects out-of-range index
- [x] `TestWasmModuleDelete_BadIndex` — rejects non-numeric index

💘 Generated with Crush


web/web.go
```diff
@@ -313,6 +317,95 @@ func (w *WebServer) handleHealthRefresh(rw http.ResponseWriter, r *http.Request)
...
+
+func (w *WebServer) handleWasmModuleDelete(rw http.ResponseWriter, r *http.Request) {
+	if err := r.ParseForm(); err != nil {
+		http.Redirect(rw, r, "/wasm?error=Invalid+form+data", http.StatusSeeOther)
+		return
+	}
+
+	indexStr := r.FormValue("index")
+	var index int
+	if _, err := fmt.Sscanf(indexStr, "%d", &index); err != nil {
```
acmacalister: `fmt.Sscanf` only requires a prefix match, so `"1foo"` parses successfully with `index=1` and `err=nil`. The `TestWasmModuleDelete_BadIndex` test only covers pure-alpha input (`"abc"`), so this slips through. `strconv.Atoi` requires the entire string to be a valid integer:

```go
index, err := strconv.Atoi(indexStr)
if err != nil {
    http.Redirect(rw, r, "/wasm?error=Invalid+module+index", http.StatusSeeOther)
    return
}
```

Worth adding a `"1abc"` case to `TestWasmModuleDelete_BadIndex` to cover this.

acmacalister (commented):
Nice, clean addition — `SetWasmModules` slots in perfectly alongside `SetIntegration`, and the test coverage is thorough. One thing to fix inline.

acmacalister (approved):
Clean PR — builds, all tests pass, lint is 0. The previous Sscanf comment was addressed with strconv.Atoi, the 11 new tests cover the full handler surface, and the port interface + mocks are in sync. LGTM.

PR #111, opened 2026-04-10
Add App Platform management tools to DigitalOcean integration

PerishableDave:
## Summary

- Adds 12 new App Platform tools to the DigitalOcean integration using the REST API directly (raw HTTP, not godo SDK)
- Rewrites existing `list_apps` and `get_app` to use the REST API
- Adds HTTP helper infrastructure (`doRequest`, `doGet`, `doPost`, `doPut`, `doDel`) to the integration for direct API calls

### New tools
| Tool | Description |
|------|-------------|
| `digitalocean_create_app` | Create app from JSON spec |
| `digitalocean_update_app` | Update app configuration |
| `digitalocean_delete_app` | Delete an app |
| `digitalocean_restart_app` | Restart app (optionally specific components) |
| `digitalocean_list_app_deployments` | List deployment history |
| `digitalocean_get_app_deployment` | Get deployment details |
| `digitalocean_create_app_deployment` | Trigger new deployment |
| `digitalocean_cancel_app_deployment` | Cancel in-progress deployment |
| `digitalocean_get_app_logs` | Get deployment logs |
| `digitalocean_get_app_health` | Get app health status |
| `digitalocean_list_app_alerts` | List app alerts |
| `digitalocean_rollback_app` | Rollback to previous deployment |

## Test plan
- [x] `go test -race ./integrations/digitalocean/...` passes
- [x] `go vet ./integrations/digitalocean/...` passes
- [x] `golangci-lint run ./integrations/digitalocean/...` — 0 issues
- [x] Dispatch parity tests verify all 64 tools have handlers and no orphans
- [ ] Manual smoke test against a DigitalOcean account with App Platform apps

### DIFF
--- a/integrations/digitalocean/apps.go
+++ b/integrations/digitalocean/apps.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/url"
 
 	mcp "github.com/daltoniam/switchboard"
 )
@@ -180,13 +181,15 @@ func getAppLogs(ctx context.Context, d *integration, args map[string]any) (*mcp.
 	logType, _ := mcp.ArgStr(args, "log_type")
 
 	path := fmt.Sprintf("/v2/apps/%s/deployments/%s/logs", appID, deployID)
-	sep := "?"
+	params := url.Values{}
 	if component != "" {
-		path += sep + "component_name=" + component
-		sep = "&"
+		params.Set("component_name", component)
 	}
 	if logType != "" {
-		path += sep + "type=" + logType
+		params.Set("type", logType)
+	}
+	if len(params) > 0 {
+		path += "?" + params.Encode()
 	}
 
 	data, err := d.doGet(ctx, "%s", path)
--- a/integrations/digitalocean/apps.go
+++ b/integrations/digitalocean/apps.go
@@ -1,0 +1,238 @@
+package digitalocean
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	mcp "github.com/daltoniam/switchboard"
+)
+
+func listApps(ctx context.Context, d *integration, args map[string]any) (*mcp.ToolResult, error) {
+	page := mcp.OptInt(args, "page", 1)
+	perPage := mcp.OptInt(args, "per_page", 20)
+	data, err := d.doGet(ctx, "/v2/apps?page=%d&per_page=%d", page, perPage)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	return mcp.RawResult(data)
+}
+
+func getApp(ctx context.Context, d *integration, args map[string]any) (*mcp.ToolResult, error) {
+	id, err := mcp.ArgStr(args, "app_id")
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	data, err := d.doGet(ctx, "/v2/apps/%s", id)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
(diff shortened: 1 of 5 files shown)


integrations/digitalocean/digitalocean.go
```diff
@@ -89,6 +104,67 @@ func listOpts(args map[string]any) *godo.ListOptions {
 	}
 }
 
+// --- HTTP helpers (used by App Platform handlers) ---
+
+func (d *integration) doRequest(ctx context.Context, method, path string, body any) (json.RawMessage, error) {
```
acmacalister, line 109: The new `doRequest` stack handles auth, retries, size limits, and 204/empty body — all things that are easy to regress silently. The fly adapter covers exactly this with `httptest`-based tests for each case (`TestDoRequest_BearerAuth`, `TestDoRequest_APIError`, `TestDoRequest_204NoContent`, `TestDoRequest_RetryableOn429`, etc. in `fly/fly_test.go`). Worth adding the same coverage here so a future change to this function doesn't quietly break App Platform tools.

integrations/digitalocean/apps.go
```diff
@@ -0,0 +1,238 @@
...
+		return mcp.ErrResult(err)
+	}
+
+	component, _ := mcp.ArgStr(args, "component")
+	logType, _ := mcp.ArgStr(args, "log_type")
+
+	path := fmt.Sprintf("/v2/apps/%s/deployments/%s/logs", appID, deployID)
+	sep := "?"
+	if component != "" {
+		path += sep + "component_name=" + component
```
acmacalister: `component` and `log_type` are concatenated directly into the query string without escaping. A component name like `my component` produces an invalid URL; a value containing `&` could silently inject additional query parameters. `url.Values` handles both cleanly — the fly adapter has a small `queryEncode` helper that does exactly this:

```go
params := url.Values{}
if component != "" {
    params.Set("component_name", component)
}
if logType != "" {
    params.Set("type", logType)
}
path := fmt.Sprintf("/v2/apps/%s/deployments/%s/logs", appID, deployID)
if len(params) > 0 {
    path += "?" + params.Encode()
}
```

acmacalister (commented):
Clean work overall — HTTP infrastructure is solid, dispatch parity passes, and the arg/err plumbing is correct. Two things inline worth addressing before merge.

PerishableDave:
Eh, going to close this and redo it. I realize that you're using the DO package and this PR uses the API.

integrations/digitalocean/apps.go
```diff
@@ -0,0 +1,241 @@
...
+		params.Set("component_name", component)
+	}
+	if logType != "" {
+		params.Set("type", logType)
+	}
+	if len(params) > 0 {
+		path += "?" + params.Encode()
+	}
+
+	data, err := d.doGet(ctx, "%s", path)
```
acmacalister, line 195: Passing an already-built URL through `doGet(ctx, "%s", path)` works today, but it's a latent footgun: if `path` ever contains a literal `%` (e.g. a percent-encoded query value like `component_name=my%20service`), `fmt.Sprintf` will treat it as a format verb and silently corrupt the URL or panic. Since the path is already fully assembled at this point, calling `doRequest` directly makes the intent clearer and removes the risk:

```go
data, err := d.doRequest(ctx, "GET", path, nil)
```

acmacalister (approved):
Really solid addition — 12 new App Platform tools, clean httptest coverage for the HTTP helper stack, and the query-string escaping fix looks great. One small thing inline.

PR #112, opened 2026-04-10
Add App Platform support to DigitalOcean integration

PerishableDave:
## Summary
- Add 8 new App Platform tools to the DigitalOcean integration: `delete_app`, `restart_app`, `list_app_deployments`, `get_app_deployment`, `create_app_deployment`, `get_app_logs`, `get_app_health`, `list_app_alerts`
- Move existing `list_apps` and `get_app` handlers from `extras.go` into a dedicated `apps.go` handler file
- Total DigitalOcean tools: 45 → 53

## Test plan
- [x] `go build` passes
- [x] `go vet` passes
- [x] `golangci-lint` passes (0 issues)
- [x] All 16 DigitalOcean tests pass including dispatch map parity, tool name validation, and no-duplicate checks
- [x] `gofmt` applied

### DIFF
--- a/integrations/digitalocean/apps.go
+++ b/integrations/digitalocean/apps.go
@@ -103,11 +103,11 @@ func getAppLogs(ctx context.Context, d *integration, args map[string]any) (*mcp.
 	r := mcp.NewArgs(args)
 	appID := r.Str("app_id")
 	logType := r.Str("log_type")
+	deploymentID := r.Str("deployment_id")
+	component := r.Str("component")
 	if err := r.Err(); err != nil {
 		return mcp.ErrResult(err)
 	}
-	deploymentID, _ := mcp.ArgStr(args, "deployment_id")
-	component, _ := mcp.ArgStr(args, "component")
 	tailLines := mcp.OptInt(args, "tail_lines", 100)
 	logs, _, err := d.client.Apps.GetLogs(ctx, appID, deploymentID, component, godo.AppLogType(logType), false, tailLines)
 	if err != nil {
--- a/integrations/digitalocean/apps.go
+++ b/integrations/digitalocean/apps.go
@@ -1,0 +1,141 @@
+package digitalocean
+
+import (
+	"context"
+
+	"github.com/digitalocean/godo"
+
+	mcp "github.com/daltoniam/switchboard"
+)
+
+func listApps(ctx context.Context, d *integration, args map[string]any) (*mcp.ToolResult, error) {
+	apps, _, err := d.client.Apps.List(ctx, listOpts(args))
+	if err != nil {
+		return errResult(err)
+	}
+	return mcp.JSONResult(apps)
+}
+
+func getApp(ctx context.Context, d *integration, args map[string]any) (*mcp.ToolResult, error) {
+	id, err := mcp.ArgStr(args, "app_id")
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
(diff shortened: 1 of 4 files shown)


acmacalister (commented):
Solid expansion of the App Platform coverage — the new handlers are clean, dispatch parity tests pass, and the refactor of listApps/getApp into apps.go is a nice cleanup. Two small things inline.

integrations/digitalocean/apps.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func getAppLogs(ctx context.Context, d *integration, args map[string]any) (*mcp.ToolResult, error) {
+	r := mcp.NewArgs(args)
+	appID := r.Str("app_id")
+	logType := r.Str("log_type")
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	deploymentID, _ := mcp.ArgStr(args, "deployment_id")
```
acmacalister: `getAppLogs` mixes the scanner pattern with standalone `mcp.ArgStr` calls, which `docs/go-anti-patterns.md` (anti-pattern #3) flags as confusing — it's not obvious which parameters are covered by `r.Err()` and which have their own checks. Since `ArgStr` returns `("", nil)` for missing keys, the `_` discards aren't actually dangerous here, but the pattern is still inconsistent.

Cleaner to just use the scanner for all of them:

```go
func getAppLogs(ctx context.Context, d *integration, args map[string]any) (*mcp.ToolResult, error) {
	r := mcp.NewArgs(args)
	appID := r.Str("app_id")
	logType := r.Str("log_type")
	deploymentID := r.Str("deployment_id")
	component := r.Str("component")
		return mcp.ErrResult(err)
	}
	tailLines := mcp.OptInt(args, "tail_lines", 100)
	// ...
}
```

(`OptInt` stays standalone since there's no scanner equivalent that accepts a default.)

integrations/digitalocean/tools.go
```diff
@@ -147,7 +147,47 @@ var tools = []mcp.ToolDefinition{
...
+		Required:   []string{"app_id", "deployment_id"},
+	},
+	{
+		Name: "digitalocean_create_app_deployment", Description: "Trigger a new deployment for an App Platform app",
+		Parameters: map[string]string{"app_id": "App UUID", "force_build": "Force rebuild even if no source changes (true/false)"},
+		Required:   []string{"app_id"},
+	},
+	{
+		Name: "digitalocean_get_app_logs", Description: "Get logs for an App Platform app. Use log_type BUILD for build logs, DEPLOY for deploy logs, or RUN for runtime logs",
+		Parameters: map[string]string{"app_id": "App UUID", "deployment_id": "Deployment UUID (omit for active deployment)", "component": "Component name (omit for all components)", "log_type": "Log type: BUILD, DEPLOY, RUN, or RUN_RESTARTED", "tail_lines": "Number of log lines to return (default 100)"},
```
acmacalister: The `log_type` description lists BUILD, DEPLOY, RUN, and RUN_RESTARTED — but the godo SDK also defines `AUTOSCALE_EVENT`. An LLM won't know to try that value if it's not listed. Worth adding it so the tool description matches the full API surface:

```go
"log_type": "Log type: BUILD, DEPLOY, RUN, RUN_RESTARTED, or AUTOSCALE_EVENT",
```

acmacalister (approved):

PR #113, opened 2026-04-10
fix(sdk): coerce string args to bool/int in guest SDK

aleksclark:
## Summary

- Switchboard passes all tool arguments as JSON strings, but `arg_bool` and `arg_int` only accepted native JSON booleans/integers. This caused `enabled` and `timeout_minutes` to silently be ignored when set via `mcp_switchboard_execute`.
- `arg_int` now falls back to parsing the string representation, and `arg_bool` accepts `"true"`/`"false"`/`"1"`/`"0"`/`"yes"`/`"no"`.

## Test plan

- [ ] Verify `arg_bool` returns `Some(true)` for `Value::Bool(true)` and `Value::String("true")`
- [ ] Verify `arg_int` returns `Some(120)` for `Value::Number(120)` and `Value::String("120")`
- [ ] Verify `arg_bool` returns `None` for `Value::String("maybe")`
- [ ] Test with overmind flow create/update using `enabled=true` and `timeout_minutes=120` via switchboard

🐨 Generated with Crush

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -32,7 +32,7 @@ func TestLoad_CreatesDefaultWhenMissing(t *testing.T) {
 	_, err = os.Stat(path)
 	assert.NoError(t, err)
 
-	assert.Len(t, m.cfg.Integrations, 28)
+	assert.Len(t, m.cfg.Integrations, 29)
 	for _, name := range []string{"github", "datadog", "linear", "sentry", "slack", "metabase", "aws", "posthog", "postgres", "clickhouse", "elasticsearch", "pganalyze", "rwx", "gmail", "homeassistant", "notion", "ynab", "gcp", "suno", "amazon", "jira", "confluence", "readarr", "salesforce", "cloudflare", "digitalocean", "fly", "snowflake"} {
 		ic, ok := m.cfg.Integrations[name]
 		assert.True(t, ok, "missing default integration: %s", name)
@@ -134,7 +134,7 @@ func TestSave(t *testing.T) {
 
 	var cfg mcp.Config
 	require.NoError(t, json.Unmarshal(data, &cfg))
-	assert.Len(t, cfg.Integrations, 28)
+	assert.Len(t, cfg.Integrations, 29)
 }
 
 func TestGet(t *testing.T) {
@@ -143,7 +143,7 @@ func TestGet(t *testing.T) {
 
 	cfg := m.Get()
 	assert.NotNil(t, cfg)
-	assert.Len(t, cfg.Integrations, 28)
+	assert.Len(t, cfg.Integrations, 29)
 }
 
 func TestUpdate(t *testing.T) {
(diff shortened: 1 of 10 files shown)


acmacalister (approved):

integrations/acp/acp.go
```diff
@@ -0,0 +1,349 @@
...
+		case EventMessageCompleted:
+			if event.Message != nil {
+				text := TextContent([]Message{*event.Message})
+				if text != "" && len(parts) == 0 {
+					parts = append(parts, text)
+				}
+			}
+		case EventRunCompleted, EventRunFailed, EventRunAwaiting, EventRunCancelled:
+			lastRun = event.Run
+		case EventError:
```
acmacalister, line 292: This early return doesn't drain the `events` channel, which can leave the `parseStream` goroutine blocked trying to write to a full buffer. Since the channel is buffered at 16, any stream that emits more than ~15 events after the error line will cause the goroutine to hang until the underlying HTTP connection closes.

Quickest fix is to drain in a separate goroutine before returning:
```go
case EventError:
    if event.Error != nil {
        go func() {
            for range events {
            }
        }()
        return &mcp.ToolResult{Data: event.Error.Message, IsError: true}, nil
    }
```
Alternatively, pass a cancellable context into `parseStream` and have the goroutine check `ctx.Done()` — that handles the general case and is cleaner if you ever add more early-exit paths.

acmacalister (commented):
Nice integration — the ACP adapter is well-structured and thoroughly tested. Two things worth looking at inline.

config/config_test.go
```diff
@@ -32,7 +32,7 @@ func TestLoad_CreatesDefaultWhenMissing(t *testing.T) {
 	_, err = os.Stat(path)
 	assert.NoError(t, err)
 
-	assert.Len(t, m.cfg.Integrations, 28)
+	assert.Len(t, m.cfg.Integrations, 29)
 	for _, name := range []string{"github", "datadog", "linear", "sentry", "slack", "metabase", "aws", "posthog", "postgres", "clickhouse", "elasticsearch", "pganalyze", "rwx", "gmail", "homeassistant", "notion", "ynab", "gcp", "suno", "amazon", "jira", "confluence", "readarr", "salesforce", "cloudflare", "digitalocean", "fly", "snowflake"} {
```
acmacalister, line 36: The count jumped from 28 to 29 but `"acp"` isn't in this name slice, so if the `"acp"` entry were accidentally dropped from `defaultConfig()` this assertion would still pass (count would drop back to 28 and fail, but only because of the `assert.Len` two lines above — the loop itself wouldn't catch the specific integration being missing).

Worth adding `"acp"` here so there's a per-name assertion for it, the same way every other integration is covered:
```go
for _, name := range []string{..., "snowflake", "acp"} {
```
And similarly in `TestDefaultConfig`'s `expected` map:
```go
"acp": {"config"},
```

PR #114, opened 2026-04-10
Add WASM module edit, configurable name, and credential sync fixes

daltoniam:
## Summary

- **WASM module edit page**: Add `GET /wasm/{index}` and `POST /wasm/{index}` routes with a pre-filled edit form for path, name, and credentials — Edit button added to each module card
- **Configurable module name**: Add optional `Name` field to `WasmModuleConfig` so users can override the name reported by the `.wasm` binary's `name()` export, wired through the web UI add/edit forms
- **Credential sync fix**: `loadWasmModule` now always merges WASM inline credentials with existing integration config and persists the result — previously on second+ boot, if the integration entry already existed and was enabled, credential updates from the WASM config were silently dropped
- **Faster startup**: Defer `browser.New()` (Playwright + Chromium init) to a background goroutine so the HTTP server starts immediately instead of blocking ~60s

## Test plan

- [x] `TestWasmModuleEdit` — edit page renders with pre-filled values
- [x] `TestWasmModuleEdit_InvalidIndex` — bad indices redirect with error
- [x] `TestWasmModuleUpdate` — save replaces module at index, preserves others
- [x] `TestWasmModuleUpdate_EmptyPath` — empty path rejected
- [x] `TestWasmModuleUpdate_InvalidIndex` — bad indices rejected
- [x] `TestWasmModuleUpdate_NoCreds` — clearing credentials works
- [x] `TestWasmModuleAdd_WithName` — name saved on add
- [x] `TestWasmModuleEdit_ShowsName` — name rendered on edit page
- [x] `TestWasmModuleUpdate_WithName` — name saved on update
- [x] `TestWasmModulesPage_ShowsName` — name shown in module list
- [x] `TestSetName` / `TestSetName_Empty` — wasm module name override
- [x] All existing tests pass, vet + lint clean

💘 Generated with Crush


cmd/server/main.go
```diff
@@ -185,14 +185,17 @@ func runServer(stdioMode bool, port int, discoverAll bool) {
...
+	browserDone := make(chan struct{})
+	var browserSvc mcp.BrowserService
+	go func() {
+		defer close(browserDone)
+		svc, err := browser.New(true /* headless */)
+		if err != nil {
+			log.Printf("browser service unavailable (%v) — browser-based integrations disabled", err)
+			return
+		}
+		browserSvc = svc
```
acmacalister, line 197: This write to `browserSvc` races with the read at the `services` struct literal below (line ~241: `Browser: browserSvc`). The struct is built on the main goroutine while this goroutine is still running, so `services.Browser` always ends up `nil`. The race detector misses it in tests because `browser.New()` fails immediately there, so the goroutine never actually writes.

Since nothing currently reads `services.Browser` (Amazon uses `SetBrowserService` directly), there's no functional regression — but the field is now permanently nil, and the concurrent write/read is technically a data race. Consider using an `atomic.Value` or a mutex-protected setter, or just remove the `Browser` field from `services` since it's not actually used by any consumer.

cmd/server/main.go
```diff
@@ -332,33 +344,44 @@ func loadWasmModule(ctx context.Context, rt *wasmmod.Runtime, wc mcp.WasmModuleC
...
+		for k, v := range existing.Credentials {
+			mergedCreds[k] = v
+		}
 	}
-	if len(configureCreds) > 0 {
-		if err := mod.Configure(ctx, configureCreds); err != nil {
+
+	if len(mergedCreds) > 0 {
+		if err := mod.Configure(ctx, mergedCreds); err != nil {
 			log.Printf("WARN: failed to configure WASM module %q: %v", wc.Path, err)
```
acmacalister, line 370: If `mod.Configure()` fails, we return early here, but the module was already registered with `reg.Register()` a few lines above. That leaves an unconfigured module in the registry — its tools will appear in search results and calls against it will fail in unexpected ways.

Would be cleaner to move `reg.Register(mod)` to after the configure step succeeds, so a configure failure doesn't leave a half-initialized integration visible to the server.

acmacalister (commented):
Solid set of changes — the credential merge logic is clean, the edit page tests are thorough, and deferring browser init is a nice startup improvement. Two things worth looking at inline.

aleksclark (approved):

cmd/server/main.go
daltoniam: Good catch — removed `Browser` from the `Services` struct literal entirely since nothing reads it. The goroutine now only communicates through `SetBrowserService` which is already mutex-protected. Fixed in 20c03d5.

cmd/server/main.go
daltoniam: Agreed — moved `reg.Register(mod)` to after `mod.Configure()` succeeds so a configure failure doesn't leave a broken module in the registry. Fixed in 20c03d5.

PR #115, opened 2026-04-10
Fix browser service race and move reg.Register after Configure

daltoniam:
## Summary

Follow-up to #114 — the review feedback fixes were pushed after the PR was merged.

- **Remove `Browser: browserSvc` from Services struct**: The background goroutine writes to `browserSvc` while the main goroutine reads it to build the struct literal — a data race. Since nothing actually reads `services.Browser` (Amazon uses `SetBrowserService` directly), just drop the field assignment.
- **Move `reg.Register(mod)` after `mod.Configure()`**: Previously a configure failure returned early but left the unconfigured module registered in the registry, making its tools visible in search results with broken execution.

## Test plan

- [x] Build passes
- [x] All tests pass
- [x] Vet + lint clean

💘 Generated with Crush

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -240,7 +240,6 @@ func runServer(stdioMode bool, port int, discoverAll bool) {
 	services := &mcp.Services{
 		Config:   cfgMgr,
 		Registry: reg,
-		Browser:  browserSvc,
 		Metrics:  mcp.NewMetrics(),
 	}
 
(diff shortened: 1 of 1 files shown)


acmacalister (approved):
Clean, correct pair of fixes. Both changes are well-scoped and the PR description explains the reasoning clearly. `mcp.Services.Browser` in `mcp.go` (line 241) is now a dead field — nothing populates it after this PR. Could be a tidy follow-up to remove it and keep the DI container honest, but not worth blocking on.

PR #116, opened 2026-04-10
Add crate-type configuration to Cargo.toml

esjay:
this was needed to generate a `.wasm` file

### DIFF
--- a/wasm/guest-rust/sdk/Cargo.toml
+++ b/wasm/guest-rust/sdk/Cargo.toml
@@ -3,6 +3,9 @@ name = "switchboard-guest-sdk"
 version = "0.1.0"
 edition = "2021"
 
+[lib]
+crate-type = ["cdylib", "rlib"]
+
 [dependencies]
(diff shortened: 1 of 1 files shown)


daltoniam (approved):

daltoniam:
Not totally sure you need this change though, think it is trying to just build the SDK as a wasm module and this is just a dependency for an a proper wasm module to be built with a real entry. Full explanation from AI:
```
This PR adds  cdylib  to the SDK in the main switchboard repo, which:

• Does not break anything (verified:  make wasm-build  + all 16 WASM tests pass)
• Does not fix the overmind build problem (that repo needs the  sdk/  directory
copied/synced over)
• Only makes the SDK itself produce a standalone  .wasm  when built directly, which isn't
particularly useful since the SDK has no entry points ( name ,  tools ,  configure ,
execute )

If the contributor's actual goal is to get overmind building, the fix is syncing the
sdk/  directory into the agents repo, not changing crate-type here.
```

PR #117, opened 2026-04-10
fix: tool prompting

arjansingh:
Attacking this at a couple levels:
1. create a `search` and `execute` meta prompting, so that if an LLM tries to execute `search` or `execute` they get a useful error explaining that they are doing it wrong.
2. adjust synonyms to improve tf-idf scoring.
3. Add "start here" guidance for integrations missing it to give LLMs usage hints.

### DIFF
--- a/server/search.go
+++ b/server/search.go
@@ -129,7 +129,7 @@ var synonymGroups = [][]string{
 	{"channel", "conversation", "room"},
 	{"state", "status"},
 	{"cycle", "sprint", "iteration"},
-	{"dashboard", "dashboards", "board", "view"},
+	{"dashboard", "dashboards", "board"},
 	{"credential", "secret", "key", "token"},
 	{"member", "user", "participant"},
 	{"email", "mail", "gmail"},
@@ -144,6 +144,8 @@ var synonymGroups = [][]string{
 	{"find", "search", "query", "queries", "lookup", "discover", "filter"},
 	{"send", "post", "publish"},
 	{"execute", "run", "invoke", "trigger"},
+	{"get", "retrieve", "fetch", "read", "show", "view", "describe"},
+	{"list", "ls", "enumerate"},
 }
 
 // buildSynonymMap expands synonym groups into a self-inclusive lookup map.
--- a/server/search.go
+++ b/server/search.go
@@ -135,6 +135,14 @@ var synonymGroups = [][]string{
 	{"email", "mail", "gmail"},
 	{"break", "fail", "crash", "down"},
 	{"metric", "metrics"},
+	{"table", "tables"},
+	{"label", "labels", "tag", "tags"},
+	{"diff", "patch", "changes"},
+	{"column", "columns", "field", "fields"},
+	{"schema", "schemas"},
+	{"database", "databases", "db"},
+	{"branch", "branches"},
+	{"comment", "comments"},
 
 	// Verbs — action synonyms
 	{"create", "add", "new", "make", "generate"},
--- a/server/server.go
+++ b/server/server.go
@@ -549,6 +549,11 @@ func (s *Server) handleExecute(ctx context.Context, req *mcpsdk.CallToolRequest)
 	if args.ToolName == "" {
 		return errorResult("either tool_name or script is required"), nil
 	}
+	if args.ToolName == "search" || args.ToolName == "execute" {
+		return errorResult(fmt.Sprintf(
+			"tool %q is a meta-tool — use it directly as an MCP tool call, not through execute",
+			args.ToolName)), nil
+	}
 	if args.Arguments == nil {
 		args.Arguments = map[string]any{}
 	}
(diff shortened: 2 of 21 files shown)


daltoniam (approved):

server/server.go
```diff
@@ -913,6 +920,15 @@ type toolExecutor struct {
...
 
 func (te *toolExecutor) Execute(ctx context.Context, toolName string, args map[string]any) (*mcp.ToolResult, error) {
+	if toolName == "search" || toolName == "execute" {
+		return &mcp.ToolResult{
+			Data: fmt.Sprintf(
+				"tool %q is a meta-tool and cannot be called from scripts. "+
+					"Use the search MCP tool before writing a script to discover tool names.",
+				toolName),
+			IsError: true,
+		}, nil
```
acmacalister, line 933: This guard catches the script path perfectly. One edge case worth noting: if an LLM uses Mode 2 (passes `tool_name: "search"`) instead of a script, the message it gets is the generic "not found" from `findTool` rather than this specific one. The existing error is still actionable, but if we want full parity with the PR's stated goal, we could add the same check inside `handleExecute` before the `executeTool` call — just something to consider if you see LLMs still confused in that flow.

server/search.go
```diff
@@ -144,6 +144,8 @@ var synonymGroups = [][]string{
 	{"find", "search", "query", "queries", "lookup", "discover", "filter"},
 	{"send", "post", "publish"},
 	{"execute", "run", "invoke", "trigger"},
+	{"get", "retrieve", "fetch", "read", "show", "view", "describe"},
+	{"list", "ls", "enumerate"},
```
acmacalister, line 156: The comment at line 117 says "Never put the same word in two groups — groups must be disjoint" but nothing enforces it. This PR actually fixed a real violation (moving `"view"` from the `dashboard` group), so the risk is real. A small test would catch future accidents:

```go
func TestSynonymGroups_Disjoint(t *testing.T) {
	seen := make(map[string][]string)
	for _, group := range synonymGroups {
		for _, word := range group {
			if prev, ok := seen[word]; ok {
				t.Errorf("word %q in two groups: %v and %v", word, prev, group)
			}
			seen[word] = group
		}
	}
}
```

acmacalister (approved):
Clean PR — builds, all tests pass, lint clean. The Instructions addition is a nice touch and the meta-tool guard fills a real gap. Two optional ideas inline.

server/server.go
```diff
@@ -541,6 +546,11 @@ func (s *Server) handleExecute(ctx context.Context, req *mcpsdk.CallToolRequest)
 	if args.ToolName == "" {
 		return errorResult("either tool_name or script is required"), nil
 	}
+	if args.ToolName == "search" || args.ToolName == "execute" {
```
acmacalister, line 549: This guard is the right call to add Mode 2 parity, but it's currently untested — unlike the script path which has `TestScriptExecution_MetaToolReturnsSpecificError`. `executeRequest` is already a test helper in `server_test.go`, so the test is pretty small:

```go
func TestHandleExecute_MetaToolReturnsSpecificError(t *testing.T) {
	s := setupTestServer()
	for _, toolName := range []string{"search", "execute"} {
		t.Run(toolName, func(t *testing.T) {
			result, err := s.handleExecute(context.Background(),
				executeRequest(toolName, map[string]any{}))
			require.NoError(t, err)
			require.False(t, result.IsError == false, "expected error result")
			tc, ok := result.Content[0].(*mcpsdk.TextContent)
			require.True(t, ok)
			assert.Contains(t, tc.Text, "meta-tool")
		})
	}
}
```

acmacalister (approved):
Good PR — build passes, all 43 test packages pass with race detector, lint is clean. The Instructions field placement, disjoint synonym enforcement, and 'Start here' consistency are all well done. One small thing inline.

server/server.go
```diff
@@ -541,6 +546,11 @@ func (s *Server) handleExecute(ctx context.Context, req *mcpsdk.CallToolRequest)
...
+		return errorResult(fmt.Sprintf(
+			"tool %q is a meta-tool — use it directly as an MCP tool call, not through execute",
+			args.ToolName)), nil
+	}
```
acmacalister, line 553: The script path gets `TestScriptExecution_MetaToolReturnsSpecificError`, but Mode 2 (passing `tool_name: "search"` or `"execute"` directly to `handleExecute`) has no test. This guard could be removed or broken without a failing test to catch it. Something like:

```go
func TestHandleExecute_MetaToolReturnsSpecificError(t *testing.T) {
	s := setupTestServer()
	for _, toolName := range []string{"search", "execute"} {
		t.Run(toolName, func(t *testing.T) {
			result, err := s.handleExecute(context.Background(),
				executeRequest(toolName, nil))
			require.NoError(t, err)
			require.True(t, result.IsError)
			tc, ok := result.Content[0].(*mcpsdk.TextContent)
			require.True(t, ok)
			assert.Contains(t, tc.Text, "meta-tool")
		})
	}
}
```

acmacalister (commented):
Solid PR — clean build, all tests pass, lint is green. The layered prompting approach (Instructions + tool descriptions + runtime errors) is well thought out, and `TestSynonymGroups_Disjoint` is a great addition. One thing inline.

PR #118, opened 2026-04-11
Bump max response cap from 50KB to 256KB

PerishableDave:
## Summary
- Raise `maxResponseBytes` in `server/server.go` from 50KB to 256KB
- The previous cap was too small for content-rich integrations like Confluence, Notion, and Jira, where a single page response regularly exceeds 50KB and fails with `Response exceeded 50KB`
- 256KB (~64K tokens) accommodates these payloads while still leaving roughly two-thirds of a 200K-token context window for the rest of the conversation
- Updated the user-facing description text in the execute tool prompt and benchmark skill docs
- Updated the three byte-cap tests to use payload sizes that exceed the new cap

## Why 256KB?
| Cap | ~Tokens | % of 200K context |
|-----|---------|-------------------|
| 50KB (old) | ~12K | 6% |
| 256KB (new) | ~64K | 32% |

256KB is large enough for rich pages (Confluence/Notion bodies, GitHub issues with comments, Datadog log windows) but small enough that a single response can't dominate a conversation. Field projection and pagination remain available for further reduction.

## Test plan
- [x] `go build` passes
- [x] `go vet` passes
- [x] `golangci-lint` passes (0 issues)
- [x] All server tests pass, including `TestHandleExecute_ByteCapEnforced`, `TestHandleExecute_ByteCapSkippedOnError`, and `TestScriptExecution_OutputByteCapEnforced` updated to use 300KB / 100KB chunks

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #119, opened 2026-04-11
Add per-integration response size cap override

PerishableDave:
## Summary
- The 50KB default response cap is too small for content-heavy integrations like Confluence, where a single `confluence_get_page` response legitimately exceeds it and fails
- Instead of bumping the global default (closes #118), introduce an optional `MaxResponseBytesIntegration` interface — mirroring `FieldCompactionIntegration` — that lets specific integrations declare a higher cap without forcing every adapter to share the same ceiling
- Confluence now declares a 256KB cap; all other integrations continue to use the tight 50KB default and remain incentivized to use filters, pagination, and field compaction

## Design

| Path | Cap |
|------|-----|
| Direct tool call (`handleExecute`) | Per-integration override if declared, else 50KB default |
| Project-scoped tool call (`project_server.go`) | Same |
| Script output (`handleScriptExecute`) | 50KB default (scripts assemble across integrations; should project/trim) |

The override only loosens the cap — integrations can't declare a *lower* cap than the default. That's intentional: the default is already the floor for "nothing should exceed this."

## Why not a global bump?
Global ceiling tuning forces every integration to share the same limit. An integration listing 10 lightweight items shouldn't share a cap designed for a 200KB Confluence page — keeping the default tight keeps compaction discipline honest where it matters.

## Why not per-tool?
Too granular. The useful unit is "this integration publishes rich content" (Confluence, Notion, etc.), not "this specific tool call does." Integrations with mixed tool shapes can always split their declaration later if needed.

## Test plan
- [x] `go build` passes
- [x] `go vet` passes
- [x] `golangci-lint` passes (0 issues)
- [x] Existing `TestHandleExecute_ByteCapEnforced` / `ByteCapSkippedOnError` / `TestScriptExecution_OutputByteCapEnforced` still pass against the renamed `defaultMaxResponseBytes`
- [x] New `TestHandleExecute_PerIntegrationCapHonored` — 100KB payload from an integration declaring 256KB succeeds (would fail under default)
- [x] New `TestHandleExecute_PerIntegrationCapStillEnforced` — 300KB payload from the same integration is rejected with the integration's own cap in the error message

## Files
- `mcp.go` — new `MaxResponseBytesIntegration` interface
- `server/server.go` — `responseLimitFor()` helper; applied in `handleExecute`
- `server/project_server.go` — same helper applied to project tool-call path
- `integrations/confluence/confluence.go` — declares 256KB cap
- `server/server_test.go` — 2 new tests + `mockIntegrationWithCap` test helper

### DIFF
--- a/server/project_server.go
+++ b/server/project_server.go
@@ -331,13 +331,14 @@ func (pr *ProjectRouter) makeExecuteHandler(def *project.Definition, scopeRule *
 		if !result.IsError {
 			result.Data = processResult(integration, args.ToolName, result.Data, pr.services.Metrics)
 
-			if len(result.Data) > maxResponseBytes {
+			limit := responseLimitFor(integration)
+			if len(result.Data) > limit {
 				if pr.services.Metrics != nil {
 					pr.services.Metrics.RecordTruncation()
 				}
 				return errorResult(fmt.Sprintf(
 					"Response exceeded %dKB (actual: %dKB). Use more specific filters, lower limit/per_page, or fetch individual items.",
-					maxResponseBytes/1024, len(result.Data)/1024,
+					limit/1024, len(result.Data)/1024,
 				)), nil
 			}
 		}
--- a/server/server.go
+++ b/server/server.go
@@ -20,7 +20,19 @@ import (
 )
 
 const defaultSearchLimit = 20
-const maxResponseBytes = 50 * 1024 // 50KB
+const defaultMaxResponseBytes = 50 * 1024 // 50KB
+
+// responseLimitFor returns the effective response byte cap for an integration.
+// Integrations that implement MaxResponseBytesIntegration may declare a higher
+// cap; values at or below the default are ignored.
+func responseLimitFor(integration mcp.Integration) int {
+	if mri, ok := integration.(mcp.MaxResponseBytesIntegration); ok {
+		if v := mri.MaxResponseBytes(); v > defaultMaxResponseBytes {
+			return v
+		}
+	}
+	return defaultMaxResponseBytes
+}
 
 // searchToolInfo represents a tool in search results.
 type searchToolInfo struct {
@@ -566,13 +578,14 @@ func (s *Server) handleExecute(ctx context.Context, req *mcpsdk.CallToolRequest)
 		}, nil
 	}
 	result.Data = processResult(integration, args.ToolName, result.Data, s.services.Metrics)
-	if len(result.Data) > maxResponseBytes {
+	limit := responseLimitFor(integration)
+	if len(result.Data) > limit {
 		if s.services.Metrics != nil {
 			s.services.Metrics.RecordTruncation()
 		}
 		return errorResult(fmt.Sprintf(
 			"Response exceeded %dKB (actual: %dKB). Use more specific filters, lower limit/per_page, or fetch individual items.",
-			maxResponseBytes/1024,
+			limit/1024,
 			len(result.Data)/1024,
 		)), nil
 	}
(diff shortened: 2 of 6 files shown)


server/project_server.go
```diff
@@ -331,13 +331,14 @@ func (pr *ProjectRouter) makeExecuteHandler(def *project.Definition, scopeRule *
 		if !result.IsError {
 			result.Data = processResult(integration, args.ToolName, result.Data, pr.services.Metrics)
 
-			if len(result.Data) > maxResponseBytes {
+			limit := responseLimitFor(integration)
```
acmacalister, line 334: The `responseLimitFor` behavior is exercised by `TestHandleExecute_PerIntegrationCapHonored` and `TestHandleExecute_PerIntegrationCapStillEnforced`, but those tests go through `server.go`'s `handleExecute`. This code path — the project router's `makeExecuteHandler` — has no equivalent test coverage for the per-integration cap.

Worth adding a `TestProjectRouter_ExecutePerIntegrationCap` that wires a `mockIntegrationWithCap` through a `ProjectRouter` and confirms: (1) a payload between the default and the integration cap succeeds, (2) a payload exceeding the integration cap is rejected with the integration's cap in the error message. The two paths share `responseLimitFor` so they'll behave identically, but without a test here a future refactor could silently diverge.

acmacalister (approved):
Clean design — the optional interface pattern mirrors `FieldCompactionIntegration` exactly, `responseLimitFor` is simple and correct, and the Confluence constant is well-commented. One gap worth closing inline.

server/server.go
```diff
@@ -20,7 +20,19 @@ import (
...
 const defaultSearchLimit = 20
-const maxResponseBytes = 50 * 1024 // 50KB
+const defaultMaxResponseBytes = 50 * 1024 // 50KB
+
+// responseLimitFor returns the effective response byte cap for an integration.
+// Integrations that implement MaxResponseBytesIntegration may declare a higher
+// cap; values at or below the default are ignored.
+func responseLimitFor(integration mcp.Integration) int {
+	if mri, ok := integration.(mcp.MaxResponseBytesIntegration); ok {
+		if v := mri.MaxResponseBytes(); v > defaultMaxResponseBytes {
```
acmacalister, line 30: The `v > defaultMaxResponseBytes` guard is documented in both the comment above and the PR description — but there's no test that exercises it. Right now all three tests (`PerIntegrationCapHonored`, `PerIntegrationCapStillEnforced`, and the project router equivalents) use a 256KB override, which always wins. Worth adding a case where an integration returns a cap at or below the default (say, `maxBytes: 1024`) and asserting that `responseLimitFor` hands back `defaultMaxResponseBytes` instead.

```go
t.Run("cap_at_or_below_default_is_ignored", func(t *testing.T) {
    mi := &mockIntegrationWithCap{
        mockIntegration: &mockIntegration{name: "bigint", healthy: true,
            tools: []mcp.ToolDefinition{{Name: "bigint_tool", Description: "..."}},
            execFn: func(_ context.Context, _ string, _ map[string]any) (*mcp.ToolResult, error) {
                return &mcp.ToolResult{Data: strings.Repeat("x", 60*1024)}, nil
            }},
        maxBytes: 1024, // below default — should be ignored
    }
    limit := responseLimitFor(mi)
    assert.Equal(t, defaultMaxResponseBytes, limit)
})
```

acmacalister (approved):
Clean design — the optional interface mirrors FieldCompactionIntegration exactly, the helper is small and easily testable, and the two code paths that apply the cap are now both covered. One small gap inline.

PR #120, opened 2026-04-11
feat: add Markdown format support

arjansingh:
For documents, markdown is often much more efficient than JSON to return.

Adding a `MarkdownIntegration` interface and package to support this.

Also migrating confluence, jira, gmail, and notion integrations to return markdown for document requests.

Minor refactor: created `mcp.ToolName` type to enforce correct passing of tools at compile time.


mcp.go
```diff
@@ -191,7 +193,22 @@ type FieldCompactionIntegration interface {
...
+
+// MarkdownIntegration is an optional interface that integrations can implement
+// to render tool responses as Markdown instead of JSON. The server calls
+// RenderMarkdown in processResult before compaction — if it returns rendered
+// content, compaction and columnarization are skipped entirely.
+type MarkdownIntegration interface {
+	// RenderMarkdown converts a tool's JSON response to Markdown.
+	// Returns (markdown, true) if the tool supports rendering.
+	// Returns ("", false) for tools that return JSON normally.
+	RenderMarkdown(toolName ToolName, data []byte) (markdown.Markdown, bool)
```
acmacalister, line 211: `RenderMarkdown` returns `markdown.Markdown` — a type from the `markdown/` subpackage — which means `mcp.go` now imports its own subpackage to express this interface. The convention docs say to use `mcp.Markdown`, but that type doesn't exist yet.

Worth adding a type alias in `mcp.go` to make the docs accurate and let adapter authors reach this type without a second import:

```go
// Markdown is rendered markdown content returned by MarkdownIntegration.
type Markdown = markdown.Markdown
```

With that alias in place, `mcp.Markdown` becomes real and the AGENTS.md convention becomes self-consistent. Adapters that already import `mcp` can reference `mcp.Markdown` directly without pulling in `markdown` separately.

AGENTS.md
```diff
@@ -58,7 +59,9 @@ Co-Authored-By: <agent model name> <noreply@anthropic.com>
 - **Args parity test** (MUST): `TestNewArgs_ErrCheckParity` in `args_test.go` walks all adapters verifying every `NewArgs` call has a matching `.Err()` check — new adapters are covered automatically
 - **Parse at boundary, not throughout** (MUST): JSON is unmarshalled once at ingress and marshalled once at egress. Use `CompactAny`/`ColumnarizeAny` for already-parsed data — never re-serialize to `[]byte` just to call `CompactJSON`/`ColumnarizeJSON`. Redundant marshal/unmarshal cycles are the #1 performance regression to guard against in the response pipeline.
 - **Entry-point guidance** (MUST): Every integration's primary tool description must include "Start here" text for wayfinding. See `add-integration` skill for pattern.
-- **Script scope**: `api.call()` in scripts can only call integration tools — the `search` and `execute` meta-tools are not callable from scripts. The `toolExecutor` returns a specific error for this.
+- **Semantic types** (MUST): `mcp.ToolName` for all tool name parameters, `mcp.Markdown` for markdown return values — never cast back to `string` in the dispatch chain. `string()` only at I/O boundaries.
```
acmacalister, line 62: `mcp.Markdown` is referenced here as a convention, but this type doesn't exist in `mcp.go` — the actual type is `markdown.Markdown` from the subpackage. This will confuse future adapter authors following these docs. Either add `type Markdown = markdown.Markdown` to `mcp.go` (which makes this line accurate), or update the reference to `markdown.Markdown` to match what's actually in the codebase.

server/server.go
```diff
@@ -577,7 +577,7 @@ func (s *Server) handleExecute(ctx context.Context, req *mcpsdk.CallToolRequest)
 			IsError: true,
 		}, nil
 	}
-	result.Data = processResult(integration, args.ToolName, result.Data, s.services.Metrics)
+	result.Data = processResult(integration, mcp.ToolName(args.ToolName), result.Data, s.services.Metrics)
```
acmacalister, line 580: `mcp.ToolName(args.ToolName)` is cast twice here — once on line 570 for `executeTool` and again on line 580 for `processResult`. `project_server.go` (line 317) does this correctly with a single `tool := mcp.ToolName(args.ToolName)` variable. Worth aligning:

```go
toolName := mcp.ToolName(args.ToolName)
integration, result, err := s.executeTool(ctx, toolName, args.Arguments)
// ...
result.Data = processResult(integration, toolName, result.Data, s.services.Metrics)
```

server/server.go
```diff
@@ -631,13 +631,26 @@ func (s *Server) handleScriptExecute(ctx context.Context, source string) (*mcpsd
...
 		return data
 	}
+
+	// Try markdown rendering first — replaces JSON compaction entirely.
+	if mr, ok := integration.(mcp.MarkdownIntegration); ok {
+		if md, rendered := mr.RenderMarkdown(toolName, []byte(data)); rendered {
+			if metrics != nil {
+				metrics.RecordMarkdownRender(toolName, len(data), len(md))
+			}
+			return string(md)
```
acmacalister, line 650: The markdown rendering branch in `processResult` has no direct unit test — it's only exercised indirectly through the adapter integration tests. Since this is the critical path for all document tools going forward, a small `TestProcessResult_MarkdownIntegration` with a mock would make regressions here much easier to isolate. Just a suggestion, not blocking.

acmacalister (approved):
Solid PR. The MarkdownIntegration optional interface fits cleanly alongside FieldCompactionIntegration, parse-at-boundary discipline is consistent across all four adapters, and all the required parity tests are in place. A couple of small things inline.

daltoniam (approved):

PR #121, opened 2026-04-13
fix(slack): skip Chrome extraction and file persistence when token provided via config

daltoniam:
## Summary

- When a Slack token is provided via config credentials (e.g. OAuth in hosted environments), skip local file I/O, Chrome workspace extraction, and background refresh — none of which apply when tokens are externally managed.
- Local-only usage (no config token) follows the original code path unchanged.

## Test plan

- [ ] Verify hosted path: configure with `token` credential, confirm no file reads/writes or Chrome calls
- [ ] Verify local path: configure without `token`, confirm Chrome extraction and file persistence still run
- [ ] Verify re-configure: switch from local to config token, confirm background refresh goroutine is stopped

🐙 Generated with [Crush](https://crush.ai)

### DIFF
--- a/integrations/slack/slack.go
+++ b/integrations/slack/slack.go
@@ -39,14 +39,15 @@ func (s *slackIntegration) Configure(ctx context.Context, creds mcp.Credentials)
 	s.clients = make(map[string]*slack.Client)
 	s.mu.Unlock()
 
-	if t := creds["token"]; t != "" {
+	configToken := creds["token"]
+	if configToken != "" {
 		teamID := creds["team_id"]
 		if teamID == "" {
 			teamID = "_config"
 		}
 		s.store.setWorkspace(&workspace{
 			TeamID: teamID,
-			Token:  t,
+			Token:  configToken,
 			Cookie: creds["cookie"],
 			Source: "config",
 		})
@@ -55,16 +56,20 @@ func (s *slackIntegration) Configure(ctx context.Context, creds mcp.Credentials)
 		}
 	}
 
-	s.store.loadFromFile()
+	// When a token was provided via config (e.g. OAuth in hosted environments),
+	// skip local file and Chrome extraction — the config token is authoritative.
+	if configToken == "" {
+		s.store.loadFromFile()
 
-	if len(s.store.allWorkspaces()) == 0 {
-		wss, _ := listWorkspacesFromChrome()
-		for _, ws := range wss {
-			s.store.setWorkspace(&workspace{
-				TeamID:   ws.TeamID,
-				TeamName: ws.Name,
-				Source:   "chrome",
-			})
+		if len(s.store.allWorkspaces()) == 0 {
+			wss, _ := listWorkspacesFromChrome()
+			for _, ws := range wss {
+				s.store.setWorkspace(&workspace{
+					TeamID:   ws.TeamID,
+					TeamName: ws.Name,
+					Source:   "chrome",
+				})
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


integrations/slack/slack.go
```diff
@@ -79,13 +84,20 @@ func (s *slackIntegration) Configure(ctx context.Context, creds mcp.Credentials)
 		s.store.setDefault(tid)
 	}
 
-	_ = s.store.saveToFile()
+	// Only persist and run background refresh for locally-sourced tokens.
+	// OAuth tokens (xoxp-*, xoxb-*) are long-lived and managed externally.
```
acmacalister: The comment says "OAuth tokens (xoxp-*, xoxb-*)" but this branch fires for any config-provided token — including `xoxc-*` browser session tokens. If someone drops an `xoxc-*` token in their config, refresh will never run and the token will silently expire after ~8 hours.

The guard is really "was this token externally managed?", not "is it an OAuth token?" — worth either tightening the comment to say "config-provided tokens" or, if we want to keep `xoxc-*` refreshing even when supplied via config, gating the refresh skip on `strings.HasPrefix(configToken, "xoxp-") || strings.HasPrefix(configToken, "xoxb-")`.

integrations/slack/slack.go
```diff
@@ -79,13 +84,20 @@ func (s *slackIntegration) Configure(ctx context.Context, creds mcp.Credentials)
...
+	if configToken == "" {
+		_ = s.store.saveToFile()
 
-	if s.stopBg != nil {
+		if s.stopBg != nil {
+			close(s.stopBg)
+		}
+		s.stopBg = make(chan struct{})
+		go s.backgroundRefresh()
+	} else if s.stopBg != nil {
```
acmacalister, line 97: The three new paths added here — (1) Configure with a config token skips file/Chrome, (2) the `else if` branch stops a running background refresh when switching to a config token, and (3) re-configuring back to local clears stopBg — don't have unit tests yet.

The PR description's own test plan calls these out; it'd be great to see them in `slack_test.go`. Something like:

```go
func TestConfigure_ConfigTokenSkipsFileAndChrome(t *testing.T) {
    // construct a slackIntegration with a mock store
    // call Configure with creds["token"] = "xoxb-test"
    // assert store has exactly 1 workspace with Source=="config"
    // assert stopBg is nil (no background refresh started)
}

func TestConfigure_SwitchToConfigTokenStopsBackgroundRefresh(t *testing.T) {
    // set up integration with stopBg already running
    // call Configure with a config token
    // assert the original stopBg channel is closed and s.stopBg == nil
}
```

Without these, a future refactor could accidentally re-enable the Chrome path for config tokens and there'd be nothing catching it.

acmacalister (commented):
Approach is solid — gating file I/O and Chrome extraction on the presence of a config token is the right way to keep hosted deployments clean. Two things worth looking at inline.

daltoniam:
Addressed both review comments in 43a98f0:

**Comment on misleading OAuth comment (line 88):** Good catch — updated the comment to say "config-provided tokens" which is what the gate actually checks. If we later need `xoxc-*` tokens to refresh even when config-supplied, we can add a prefix check at that point.

**Comment on missing tests (line 97):** Added three tests covering the new paths:
- `TestConfigure_ConfigTokenSkipsFileAndChrome` — verifies exactly one workspace with `Source=="config"` and `stopBg==nil`
- `TestConfigure_ConfigTokenDefaultTeamID` — verifies fallback to `_config` team ID
- `TestConfigure_SwitchToConfigTokenStopsBackgroundRefresh` — verifies pre-existing stopBg channel is closed and set to nil

integrations/slack/slack_test.go
```diff
@@ -347,6 +347,60 @@ func TestGetClientForArgs_UnknownWorkspace(t *testing.T) {
 	assert.Contains(t, err.Error(), "unknown workspace")
 }
 
+// --- Configure token-gating tests ---
+
+func TestConfigure_ConfigTokenSkipsFileAndChrome(t *testing.T) {
+	s := &slackIntegration{}
+	s.Configure(t.Context(), mcp.Credentials{
```
acmacalister, line 354: All three new test functions call `Configure(...)` but discard the error return. The linter excludes `errcheck` in test files so CI won't catch this, but it's worth fixing — if `Configure` ever starts returning an error on the config-token path, these tests would silently continue and the assertions would fail with confusing messages instead of pointing at the root cause.

```go
require.NoError(t, s.Configure(t.Context(), mcp.Credentials{
	"token":   "xoxb-test-token",
	"team_id": "T_CONFIG",
}))
```

Same pattern applies to the `ConfigTokenDefaultTeamID` and `SwitchToConfigTokenStopsBackgroundRefresh` tests.

acmacalister (approved):
Both previous review comments are addressed — the comment wording is accurate and all three test cases are in place. One small thing inline.

PR #122, opened 2026-04-13
fix: markdown rendering and types

arjansingh:
Address #120 feedback.

I also added some more semantic types and did some whitespace trimming so that ` tool_name` searches will return `tool_name`.

### DIFF
--- a/mcp.go
+++ b/mcp.go
@@ -200,6 +200,10 @@ type FieldCompactionIntegration interface {
 // Semantic type prevents mixing with arbitrary strings at interface boundaries.
 type ToolName string
 
+// Markdown is rendered markdown content returned by MarkdownIntegration.
+// Alias lets adapter authors use mcp.Markdown without importing the markdown subpackage.
+type Markdown = markdown.Markdown
+
 // MarkdownIntegration is an optional interface that integrations can implement
 // to render tool responses as Markdown instead of JSON. The server calls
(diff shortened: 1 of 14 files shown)


daltoniam (approved):

mcp.go
```diff
@@ -63,6 +64,15 @@ func ParseRetryAfter(header string) time.Duration {
 // Credentials holds key-value credential pairs for an integration.
 type Credentials map[string]string
 
+// Credential key constants for OAuth/infrastructure keys that are not
+// real secrets. Centralized here to prevent hasCredentials and web handlers
```
acmacalister, line 68: The phrase "not real secrets" here is a bit misleading — `client_secret` is a real secret. What the comment means to say is that these keys aren't treated as *active credentials* for the purpose of `hasCredentials`: the server needs an access token or API key to be functional, and `client_id`/`client_secret`/`token_source` alone don't grant access. Worth tightening the wording to avoid confusion later:

```go
// Credential key constants for OAuth plumbing metadata.
// These keys are excluded from hasCredentials — the server requires an
// access token or API key to be functional; these fields alone don't grant access.
```

acmacalister (approved):
Solid work — `ToolName.UnmarshalJSON` cleanly handles the whitespace-trimming bug, the `resultProcessor` refactor makes the processing pipeline much easier to extend, and the test coverage is thorough. One small comment on the constant block doc.

PR #123, opened 2026-04-13
feat: add session context and breadcrumb trail meta-tools

acmacalister:
## Summary

- **Session context**: New `session` meta-tool lets LLMs set key-value context (e.g. `owner`, `repo`) once per session. All subsequent `execute` calls auto-inject these as default args — explicit args always override. Eliminates repetitive parameter passing across multi-call interactions.
- **Breadcrumb trail**: New `history` meta-tool returns a compact log of prior tool calls in the session, with optional tool-name filter. Enables LLMs to recover context after mid-conversation compression without re-executing calls.
- **Session infrastructure**: `SessionStore` with configurable TTL (default 1h), lazy eviction, breadcrumb cap (200 entries FIFO), and `WithSessionTTL` server option. Both new meta-tools are blocked from `execute` and scripts.

This is Phase 1 of the [stateful features plan](SESSION-TODO.md) — session context was identified as the highest-impact feature since it reduces token waste on every multi-call interaction.

## Test plan

- [x] 19 unit tests: session CRUD, merge precedence (explicit > session), breadcrumb cap, TTL expiry, concurrent access
- [x] 11 integration tests: handleSession set/get/clear, handleHistory with filter, execute with session injection, explicit override, breadcrumb recording, meta-tool blocking
- [x] `go vet ./...` clean

🐘 Generated with Crush


server/server.go
```diff
@@ -558,7 +613,7 @@ func (s *Server) handleExecute(ctx context.Context, req *mcpsdk.CallToolRequest)
 	if args.ToolName == "" {
 		return errorResult("either tool_name or script is required"), nil
 	}
-	if args.ToolName == "search" || args.ToolName == "execute" {
+	if args.ToolName == "search" || args.ToolName == "execute" || args.ToolName == "session" || args.ToolName == "history" {
```
arjansingh, line 625: haha we finally have more than two tools.

arjansingh (approved):

server/session.go
```diff
@@ -0,0 +1,216 @@
...
+	for k, v := range pairs {
+		s.Context[k] = v
+	}
+	s.LastUsed = time.Now()
+}
+
+func (s *Session) GetContext() map[string]any {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	s.touch()
```
acmacalister: `touch()` writes `s.LastUsed` but `GetContext()` holds only an `RLock` — that's a write under a read lock, which is a data race. Two goroutines calling `GetContext()` concurrently will race on `LastUsed`. The race detector didn't catch it in the test suite likely because the concurrent-`GetContext` path isn't exercised in parallel.

Simplest fix is to drop the `touch()` call from `GetContext()` — reads probably shouldn't reset TTL anyway. If you want reads to refresh the TTL, upgrade to a full `Lock()`:

```go
func (s *Session) GetContext() map[string]any {
	s.mu.Lock()
	defer s.mu.Unlock()
	s.LastUsed = time.Now()
	cp := make(map[string]any, len(s.Context))
	for k, v := range s.Context {
		cp[k] = v
	}
	return cp
}
```

server/session_handlers.go
```diff
@@ -0,0 +1,96 @@
...
+	if sess == nil {
+		sess = s.sessionStore.GetOrCreate("default")
+	}
+
+	bcs := sess.RecentBreadcrumbs(args.LastN, mcp.ToolName(args.Tool))
+
+	result, err := mcp.JSONResult(map[string]any{
+		"session_id":       sess.ID,
+		"breadcrumbs":      bcs,
+		"total_in_session": len(sess.Breadcrumbs),
```
acmacalister: `sess.Breadcrumbs` is read here without holding `sess.mu`, but `AddBreadcrumb` writes to it under a full `Lock()`. This is a data race — a concurrent `execute` call recording a breadcrumb while `handleHistory` reads `len(sess.Breadcrumbs)` will be flagged by the race detector if both run in parallel.

The cleanest fix is to expose a `TotalBreadcrumbs()` method on `Session` that reads under `RLock`:

```go
func (s *Session) TotalBreadcrumbs() int {
	s.mu.RLock()
	defer s.mu.RUnlock()
	return len(s.Breadcrumbs)
}
```

Then use `sess.TotalBreadcrumbs()` here instead.

server/server.go
```diff
@@ -567,17 +622,26 @@ func (s *Server) handleExecute(ctx context.Context, req *mcpsdk.CallToolRequest)
 		args.Arguments = map[string]any{}
 	}
 
-	integration, result, err := s.executeTool(ctx, mcp.ToolName(args.ToolName), args.Arguments)
+	sess := sessionFromCtx(ctx)
+	if sess == nil {
+		sess = s.sessionStore.GetOrCreate("default")
```
acmacalister: Every HTTP client — and every call to `handleSession`/`handleHistory` — falls back to the `"default"` session when `sessionFromCtx` returns nil. That means two simultaneous HTTP MCP clients share a single session: client A's `owner/repo` bleeds into client B's calls.

The MCP SDK surfaces a per-connection session ID via `req.Session.ID()` (the `CallToolRequest` embeds `*ServerSession`). Wiring that in would give each HTTP connection its own isolated bucket:

```go
sessID := "default"
if req.Session != nil {
	sessID = req.Session.ID()
}
sess := s.sessionStore.GetOrCreate(sessID)
```

Same change needed in `handleSession` and `handleHistory`. For stdio the single-connection model means `"default"` is fine, but for HTTP this is a correctness issue as soon as two clients connect.

docs/stateful-features.md
```diff
@@ -0,0 +1,499 @@
+# Stateful Features for Switchboard (Open Source)
```
acmacalister, line 1: This file is a 499-line internal roadmap — it's great planning material but probably doesn't belong committed to the OSS repo. It exposes implementation timelines and unfinished feature designs to external contributors, and it'll go stale quickly as features land.

Would recommend moving this to a GitHub Project / issue with the same content, or at minimum adding it to `.gitignore`. The PR description already captures the Phase 1 summary well.

acmacalister (commented):
Solid feature — the design is clean, the test coverage is thorough, and the breadcrumb FIFO cap and copy-on-read semantics are done right. Three concurrency issues worth fixing before merge, and a minor note on the planning doc.

arjansingh (approved):

server/server.go
```diff
@@ -216,8 +230,55 @@ List issues with server-side projection (only id, title, labels — no manual .m
 		}, nil),
 	}
 
+	sessionTool := &mcpsdk.Tool{
```
arjansingh, line 236: i think the global prompt also needs to explain session as well. i don't see it anywhere.

docs/stateful-features.md
arjansingh, line 1: can we throw this in /docs?

server/session.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+func (s *Session) TotalBreadcrumbs() int {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	return len(s.Breadcrumbs)
+}
+
+func (s *Session) touch() {
+	s.LastUsed = time.Now()
```
acmacalister, line 134: `touch()` writes `s.LastUsed` without holding `s.mu`, but every other `LastUsed` write (`SetContext`, `ClearContext`, `AddBreadcrumb`) happens under a full `s.mu.Lock()`. If two goroutines hold the same `*Session` and one calls `GetOrCreate` (store lock only, calls `touch`) while the other calls `SetContext` (session lock), they race on `s.LastUsed`.

This is the same class of bug the earlier review caught on `GetContext`. The fix is to make `touch` acquire the session lock:

```go
func (s *Session) touch() {
	s.mu.Lock()
	s.LastUsed = time.Now()
	s.mu.Unlock()
}
```

Then `GetOrCreate` can drop its own explicit lock around the call — it just needs the store lock for the map.

server/session_file.go
```diff
@@ -0,0 +1,240 @@
...
+func (fs *FileSessionStore) Get(id string) (*Session, bool) {
+	fs.mu.RLock()
+	defer fs.mu.RUnlock()
+	if s, ok := fs.mem[id]; ok {
+		if time.Since(s.LastUsed) <= fs.ttl {
+			return s, true
+		}
+		return nil, false
+	}
+	s := fs.loadFromDisk(id)
```
acmacalister, line 67: `Get` loads from disk when the session isn't in the in-memory cache, but never populates `fs.mem` with the result. A caller that does `Get` repeatedly will re-read the file every time. `GetOrCreate` does cache the disk result at line 49 — `Get` could do the same for consistency, though it would need a lock upgrade (RLock → Lock). Alternatively, callers in `handleSession`/`handleHistory` could consistently use `GetOrCreate` instead of `Get`, which already has the caching logic.

acmacalister (commented):
Solid implementation — the session store interface is clean, the file format is human-readable (nice touch), and the follow-up commit addressed the earlier review feedback thoroughly. Two small concurrency items inline.

PR #124, opened 2026-04-13
feat: ollama integration

arjansingh:
1. add support for ollama's built in API
2. adds a rendered output option so tools can pass around output rendered from our render pipeline directly to each other during execution scripts.

### DIFF
--- a/integrations/ollama/compact_specs_test.go
+++ b/integrations/ollama/compact_specs_test.go
@@ -35,7 +35,10 @@ func TestFieldCompactionSpecs_OnlyReadTools(t *testing.T) {
 }
 
 func TestFieldCompactionSpecs_ExpectedTools(t *testing.T) {
-	expected := []mcp.ToolName{"ollama_list_models", "ollama_list_running"}
+	expected := []mcp.ToolName{
+		"ollama_list_models", "ollama_list_running",
+		"ollama_chat", "ollama_generate", "ollama_embed",
+	}
 	for _, name := range expected {
 		_, ok := fieldCompactionSpecs[name]
 		assert.True(t, ok, "expected compaction spec for %s", name)
--- a/integrations/ollama/compact_specs_test.go
+++ b/integrations/ollama/compact_specs_test.go
@@ -1,0 +1,43 @@
+package ollama
+
+import (
+	"testing"
+
+	mcp "github.com/daltoniam/switchboard"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestFieldCompactionSpecs_AllParse(t *testing.T) {
+	require.NotEmpty(t, fieldCompactionSpecs, "fieldCompactionSpecs should not be empty")
+}
+
+func TestFieldCompactionSpecs_NoDuplicateTools(t *testing.T) {
+	assert.Equal(t, len(rawFieldCompactionSpecs), len(fieldCompactionSpecs))
+}
+
+func TestFieldCompactionSpecs_NoOrphanSpecs(t *testing.T) {
+	for toolName := range fieldCompactionSpecs {
+		_, ok := dispatch[toolName]
+		assert.True(t, ok, "field compaction spec for %q has no dispatch handler", toolName)
+	}
+}
+
+func TestFieldCompactionSpecs_OnlyReadTools(t *testing.T) {
+	mutationPrefixes := []string{"create", "update", "delete", "pull", "push", "copy"}
+	for toolName := range fieldCompactionSpecs {
+		name := string(toolName)
+		for _, prefix := range mutationPrefixes {
+			assert.NotContains(t, name, "_"+prefix+"_",
+				"mutation tool %q should not have compaction specs", toolName)
+		}
+	}
+}
+
+func TestFieldCompactionSpecs_ExpectedTools(t *testing.T) {
+	expected := []mcp.ToolName{"ollama_list_models", "ollama_list_running"}
+	for _, name := range expected {
+		_, ok := fieldCompactionSpecs[name]
+		assert.True(t, ok, "expected compaction spec for %s", name)
+	}
+}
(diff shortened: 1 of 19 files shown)


script/engine.go
```diff
@@ -242,6 +251,65 @@ func (e *Engine) Run(ctx context.Context, source string) (*mcp.ToolResult, error
...
+		return nil, fmt.Errorf("failed to set api.callRendered: %w", err)
+	}
+
+	if err := apiObj.Set("tryCallRendered", func(call goja.FunctionCall) goja.Value {
+		if err := ctx.Err(); err != nil {
+			panic(vm.NewGoError(fmt.Errorf("script cancelled: %w", err)))
+		}
+		callCount++
+		if callCount > e.maxCalls {
+			return vm.ToValue(map[string]any{"ok": false, "error": fmt.Sprintf("exceeded maximum of %d api.call() invocations", e.maxCalls)})
```
acmacalister: The error message says `api.call() invocations` but this is inside `tryCallRendered` — a script that burns through the limit via `callRendered` calls will get a confusing error pointing at the wrong method. The message should say `api calls` (or similar) since the limit is shared across all four call variants.

```go
return vm.ToValue(map[string]any{"ok": false, "error": fmt.Sprintf("exceeded maximum of %d api calls per script", e.maxCalls)})
```

Same copy appears in `tryCall` at line 222 — could update both while here.

server/server.go
```diff
@@ -987,6 +1000,30 @@ func (te *toolExecutor) Execute(ctx context.Context, toolName mcp.ToolName, args
 	return result, err
 }
 
+// ExecuteRendered applies the same markdown/compaction/columnarization pipeline
+// as the execute meta-tool, returning LLM-readable output. Used by api.callRendered().
+func (te *toolExecutor) ExecuteRendered(ctx context.Context, toolName mcp.ToolName, args map[string]any) (*mcp.ToolResult, error) {
+	if toolName == "search" || toolName == "execute" {
```
acmacalister: The meta-tool guard is duplicated verbatim between `Execute` and `ExecuteRendered`. If we ever add a third meta-tool or change the error wording we'll need to update it in two places. A small helper would keep this DRY:

```go
func (te *toolExecutor) checkMetaTool(toolName mcp.ToolName) *mcp.ToolResult {
	if toolName == "search" || toolName == "execute" {
		return &mcp.ToolResult{
			Data: fmt.Sprintf(
				"tool %q is a meta-tool and cannot be called from scripts. "+
					"Use the search MCP tool before writing a script to discover tool names.",
				toolName),
			IsError: true,
		}
	}
	return nil
}
```

integrations/ollama/types.go
```diff
@@ -0,0 +1,201 @@
...
+	// context (token ID array) intentionally omitted.
+	timingStats
+}
+
+// --- /api/embed ---
+
+type embedRequest struct {
+	Model      ModelName       `json:"model"`
+	Input      any             `json:"input"`
+	Stream     bool            `json:"stream"`
```
acmacalister: `Stream` is always serialized as `"stream":false` in the embed request, but the Ollama `/api/embed` endpoint doesn't support streaming at all — the field is a no-op and just adds noise to the wire payload. Adding `omitempty` (or removing it entirely) would keep the request body clean:

```go
Stream bool `json:"stream,omitempty"`
```

Same note applies to `chatRequest` and `generateRequest`, but for those it's explicitly documenting intent (force non-streaming), so removing it there would reduce clarity.

integrations/ollama/compact_specs_test.go
```diff
@@ -0,0 +1,43 @@
...
+		name := string(toolName)
+		for _, prefix := range mutationPrefixes {
+			assert.NotContains(t, name, "_"+prefix+"_",
+				"mutation tool %q should not have compaction specs", toolName)
+		}
+	}
+}
+
+func TestFieldCompactionSpecs_ExpectedTools(t *testing.T) {
+	expected := []mcp.ToolName{"ollama_list_models", "ollama_list_running"}
```
acmacalister: `TestFieldCompactionSpecs_ExpectedTools` only checks that the two list tools have compaction specs, but `ollama_chat`, `ollama_generate`, and `ollama_embed` are also in `rawFieldCompactionSpecs` and worth asserting here. The `NoOrphanSpecs` test already catches specs without handlers, but this test is the only place where we verify the *intent* ("inference tools should strip timing stats"). Adding the three inference tools to `expected` would lock that in.

acmacalister (approved):
Solid work on both the Ollama integration and the callRendered pipeline. The interface-as-optional-capability pattern for RenderedExecutor is clean, the test coverage is thorough, and the tool descriptions/compaction specs are well-thought-out. Few small things inline.

integrations/ollama/types.go
```diff
@@ -0,0 +1,200 @@
...
+	Role      ChatRole        `json:"role"`
+	Content   string          `json:"content"`
+	Thinking  string          `json:"thinking,omitempty"`
+	ToolCalls json.RawMessage `json:"tool_calls,omitempty"`
+}
+
+// --- /api/generate ---
+
+type generateRequest struct {
+	Model     ModelName       `json:"model"`
```
acmacalister, line 152: `Prompt` serializes as `"prompt":""` on the wire whenever the caller doesn't pass a prompt — `generate()` only sets it when non-empty, but the struct tag lacks `omitempty`. `Suffix`, `System`, and all the other optional string fields in this struct consistently use `omitempty`, so this is the odd one out.

```go
Prompt    string          `json:"prompt,omitempty"`
```

Worth adding a test case alongside `TestGenerate` that verifies an empty `prompt` field isn't sent (similar to how `TestEmbed_Success` at line 116 asserts `assert.Nil(t, body["stream"])`).

acmacalister (approved):
Clean PR — builds, tests pass, lint clean, all three previous review comments addressed. One small thing inline.

PR #125, opened 2026-04-13
Migrate jira_search_issues to /search/jql endpoint

PerishableDave:
## Summary
- Migrates `jira_search_issues` from deprecated `POST /rest/api/3/search` to `POST /rest/api/3/search/jql` ([Atlassian deprecation notice](https://developer.atlassian.com/changelog/#CHANGE-2046))
- Replaces offset-based pagination (`start_at` / `startAt`) with cursor-based pagination (`next_page_token` / `nextPageToken`)
- Bumps default `maxResults` from 50 to 200 (new endpoint supports up to 5000)
- Adds `nextPageToken` and `isLast` to compaction spec so pagination fields survive response compaction

## Test plan
- [x] Added `TestSearchIssues` table-driven test verifying new endpoint path, cursor parameter, and default max
- [x] Updated compact spec shape parity test payload to match new response format
- [x] All existing tests pass (`go test ./integrations/jira/`)
- [ ] Live smoke test against a Jira Cloud instance

### DIFF
--- a/integrations/jira/jira_test.go
+++ b/integrations/jira/jira_test.go
@@ -345,14 +345,12 @@ func TestSearchIssues(t *testing.T) {
 		wantPath  string
 		wantToken string
 		wantMax   float64
-		noStartAt bool
 	}{
 		{
-			name:      "basic search hits /search/jql",
-			args:      map[string]any{"jql": "project = PROJ"},
-			wantPath:  "/search/jql",
-			wantMax:   200,
-			noStartAt: true,
+			name:     "basic search hits /search/jql",
+			args:     map[string]any{"jql": "project = PROJ"},
+			wantPath: "/search/jql",
+			wantMax:  200,
 		},
 		{
 			name:      "with next_page_token",
@@ -360,14 +358,12 @@ func TestSearchIssues(t *testing.T) {
 			wantPath:  "/search/jql",
 			wantToken: "abc123",
 			wantMax:   200,
-			noStartAt: true,
 		},
 		{
-			name:      "custom max_results",
-			args:      map[string]any{"jql": "project = PROJ", "max_results": float64(500)},
-			wantPath:  "/search/jql",
-			wantMax:   500,
-			noStartAt: true,
+			name:     "custom max_results",
+			args:     map[string]any{"jql": "project = PROJ", "max_results": float64(500)},
+			wantPath: "/search/jql",
+			wantMax:  500,
 		},
 	}
 
@@ -380,12 +376,14 @@ func TestSearchIssues(t *testing.T) {
 				var body map[string]any
 				require.NoError(t, json.NewDecoder(r.Body).Decode(&body))
 
-				if tt.noStartAt {
-					_, hasStartAt := body["startAt"]
-					assert.False(t, hasStartAt, "request body should not contain startAt")
-				}
+				_, hasStartAt := body["startAt"]
+				assert.False(t, hasStartAt, "request body should not contain startAt")
+
 				if tt.wantToken != "" {
 					assert.Equal(t, tt.wantToken, body["nextPageToken"])
+				} else {
+					_, hasToken := body["nextPageToken"]
+					assert.False(t, hasToken, "nextPageToken should not be sent on first page")
 				}
 				assert.Equal(t, tt.wantMax, body["maxResults"])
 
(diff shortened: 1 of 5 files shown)


integrations/jira/jira_test.go
```diff
@@ -338,6 +338,70 @@ func TestQueryEncode(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				assert.Equal(t, tt.wantPath, r.URL.Path)
+				assert.Equal(t, "POST", r.Method)
+
+				var body map[string]any
+				require.NoError(t, json.NewDecoder(r.Body).Decode(&body))
+
+				if tt.noStartAt {
```
acmacalister: The `noStartAt` field is `true` in every test case, so this branch always runs — the field isn't doing any work. The `startAt`-absent assertion is actually the key invariant of this migration (it proves the old offset param is gone); it should be unconditional rather than opt-in.

Simplest fix is to drop the `noStartAt` field from the struct and remove the `if tt.noStartAt` guard:

```go
_, hasStartAt := body["startAt"]
assert.False(t, hasStartAt, "request body should not contain startAt")
```

integrations/jira/jira_test.go
```diff
@@ -338,6 +338,70 @@ func TestQueryEncode(t *testing.T) {
...
+					_, hasStartAt := body["startAt"]
+					assert.False(t, hasStartAt, "request body should not contain startAt")
+				}
+				if tt.wantToken != "" {
```
acmacalister, line 382: We check that `nextPageToken` is present when passed, but there's no assertion that it's *absent* on the initial (no-token) requests. Worth adding to the "basic search" and "custom max_results" cases so a future regression (e.g., accidentally defaulting the token) would be caught:

```go
if tt.wantToken != "" {
    assert.Equal(t, tt.wantToken, body["nextPageToken"])
} else {
    _, hasToken := body["nextPageToken"]
    assert.False(t, hasToken, "nextPageToken should not be sent on first page")
}
```

integrations/jira/tools.go
```diff
@@ -5,8 +5,8 @@ import mcp "github.com/daltoniam/switchboard"
 var tools = []mcp.ToolDefinition{
 	// ── Issues ───────────────────────────────────────────────────────
 	{
-		Name: mcp.ToolName("jira_search_issues"), Description: "Search issues using JQL (Jira Query Language). Start here for most issue workflows. Returns paginated results",
-		Parameters: map[string]string{"jql": "JQL query (e.g., 'project = PROJ AND status = Open')", "fields": "Comma-separated fields to return (default: summary,status,assignee,priority,issuetype)", "start_at": "Pagination offset (0-based)", "max_results": "Max results per page (default 50, max 100)"},
+		Name: mcp.ToolName("jira_search_issues"), Description: "Search issues using JQL (Jira Query Language). Start here for most issue workflows. Returns paginated results; pass nextPageToken from response to fetch next page",
+		Parameters: map[string]string{"jql": "JQL query (e.g., 'project = PROJ AND status = Open')", "fields": "Comma-separated fields to return (default: summary,status,assignee,priority,issuetype)", "next_page_token": "Cursor for next page (from previous response's nextPageToken field)", "max_results": "Max results per page (default 200, max 5000)"},
```
acmacalister: The Atlassian docs for `/search/jql` don't explicitly publish a maximum for `maxResults` — the old `/search` endpoint had a 100-item cap and the new one is higher, but the server-enforced ceiling isn't clearly documented. Advertising `max 5000` is probably fine in practice (the server will just clamp), but if LLMs take this literally and request 5000 on every call, they'll get silently truncated results whenever the actual cap is lower. Might be safer as `max results per page (default 200, server may cap)`.

acmacalister (approved):
Clean migration — correct endpoint, cursor pagination wired end-to-end, compaction spec updated, tests pass with the race detector. A couple of small test improvements inline.

acmacalister (approved):
All three comments from the previous round are addressed — the dead noStartAt guard is gone, both negative token assertions are in place, and the max_results description is softened. Build passes, full test suite green with -race, lint clean. LGTM.

PR #126, opened 2026-04-14
feat: add result pinning with handle references (Phase 2)

acmacalister:
## Summary

- **Auto-pinning**: Every successful `execute` call pins its result with a session-scoped handle (`$1`, `$2`, ...). The handle is surfaced as a second content item in the response so LLMs know it's available.
- **Arg references**: LLMs can pass `"$1.owner.login"` as an argument value — the server resolves it from the pinned result before execution, eliminating re-fetch chains like "get PR, then get its owner, then get their repos."
- **`pin` meta-tool**: `list` shows all handles with tool name and size, `get` retrieves a pinned result with optional dot-path extraction, `unpin` frees memory.
- **Memory cap**: 5MB total pinned data per session with LRU eviction of oldest pins when exceeded.

This is Phase 2 of the [stateful features plan](docs/stateful-features.md), building on Phase 1's session infrastructure (PR #123).

## Test plan

- [x] 16 unit tests: pin/unpin, handle increment, eviction, ResolveRef with full object/dot path/array index, splitRef, extractPath
- [x] 12 integration tests: auto-pin in execute, ref resolution in args, pin list/get/unpin handlers, meta-tool blocking

🐘 Generated with Crush

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -110,7 +110,9 @@ func New(services *mcp.Services, opts ...Option) *Server {
 				"(e.g., {\"query\": \"get page\"} finds notion_retrieve_page). " +
 				"Use the session tool to set context (e.g., owner/repo) once — " +
 				"subsequent execute calls auto-inject those values as defaults. " +
-				"Use the history tool to review prior calls after context compression.",
+				"Use the history tool to review prior calls after context compression. " +
+				"Every execute result is auto-pinned ($1, $2, ...) — use pin to retrieve or " +
+				"pass handles like $1.field in execute arguments to reference previous results.",
 		},
 	)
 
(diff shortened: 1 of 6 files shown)


server/session_pin.go
```diff
@@ -0,0 +1,185 @@
...
+
+	var parsed any
+	if err := json.Unmarshal(pr.Data, &parsed); err != nil {
+		return nil, fmt.Errorf("unmarshal pinned data: %w", err)
+	}
+	return extractPath(parsed, path)
+}
+
+func (s *Session) evictPinnedIfNeeded(incoming int) {
+	for s.pinnedSize+incoming > MaxPinnedBytes && len(s.pinned) > 0 {
```
acmacalister, line 117: When the pin map is empty and `incoming` is larger than `MaxPinnedBytes`, the loop condition is true but `len(s.pinned) == 0` so the loop body never runs — the oversized item gets stored anyway, blowing past the cap. In practice the 50KB response limit keeps most results small, but since `PinResult` is called with raw (pre-compaction) data, a pathological result could still slip through. A cheap guard at the top of `PinResult` would close this:

```go
if size > MaxPinnedBytes {
    return "" // too large to pin; skip silently
}
```

server/server.go
```diff
@@ -643,6 +676,10 @@ func (s *Server) handleExecute(ctx context.Context, req *mcpsdk.CallToolRequest)
 		_ = s.sessionStore.Save(sess)
 		return errorResult(err.Error()), nil
 	}
+	var handle string
+	if !result.IsError {
+		handle = sess.PinResult(args.ToolName, result.Data)
```
acmacalister, line 681: `PinResult` is called here with the raw integration response, before `processResult` (line 691) and the response-size check (lines 693–701). That means if the processed result still exceeds the 50KB limit, the LLM gets an error response — but `$1` is already pinned and fully resolvable. That's a subtle contract: the handle exists even when the caller never saw the data.

Pinning pre-compaction is correct for path resolution (raw JSON preserves all field names), but it's worth a comment here explaining the intentional ordering, and potentially clearing/not emitting the handle in the truncation error path so LLMs don't get confused by a `$1` they never received output for.

server/session_pin.go
```diff
@@ -0,0 +1,185 @@
...
+			continue
+		}
+		if len(s) < 2 {
+			continue
+		}
+		if s[1] < '0' || s[1] > '9' {
+			continue
+		}
+		resolved, err := sess.ResolveRef(s)
+		if err == nil {
```
acmacalister, line 181: Resolution errors are silently swallowed here — if `$1.owner.login` doesn't resolve (expired session, wrong path, non-JSON data), the argument is left as the raw string `"$1.owner.login"` and passed to the downstream tool. The tool then gets a confusing string value it doesn't expect.

Depending on the desired UX, we could either surface the error directly:

```go
resolved, err := sess.ResolveRef(s)
if err != nil {
    // leave arg as-is and let the tool surface a cleaner error,
    // or propagate: return fmt.Errorf("resolve ref %q: %w", s, err)
}
args[k] = resolved
```

At minimum, a comment explaining "silent fail intentional" would help future readers.

server/session_handlers.go
```diff
@@ -96,3 +96,89 @@ func (s *Server) handleHistory(ctx context.Context, req *mcpsdk.CallToolRequest)
...
 	}, nil
 }
+
+func (s *Server) handlePin(ctx context.Context, req *mcpsdk.CallToolRequest) (*mcpsdk.CallToolResult, error) {
+	var args struct {
+		Action string `json:"action"`
+		Handle string `json:"handle"`
+		Path   string `json:"path"`
+	}
+	if err := json.Unmarshal(req.Params.Arguments, &args); err != nil {
```
acmacalister, line 106: `handleHistory` (line 65) guards against nil `req.Params.Arguments` before unmarshalling, but `handlePin` here does not. The MCP SDK will likely always populate arguments when `action` is a required enum field, but for consistency — and to avoid a panic if the SDK ever sends a nil body — worth adding the same guard:

```go
if req.Params.Arguments == nil {
    return errorResult("arguments required"), nil
}
```

acmacalister (commented):
Solid implementation — the auto-pin + handle reference model is a clean UX win and the session-scoped memory management is well thought out. A few non-blocking things to look at inline.

PR #127, opened 2026-04-14
feat: add web_fetch integration for URL content retrieval

acmacalister:
## Summary

- Adds a `web_fetch` tool that fetches URLs and returns readable text, closing the knowledge gap for local/smaller models that lack built-in web browsing
- Security hardened: HTTPS-only, private/localhost address blocking, 5-redirect cap, 40K char truncation, HTML noise stripping with code block preservation
- Adds `url`/`docs` synonym groups to the search engine so natural queries like "fetch url", "read docs", "get webpage" surface the tool

## Details

New adapter at `integrations/webfetch/` with a single tool (`web_fetch`) designed for minimal tool-calling overhead — flat two-parameter schema (url + timeout) that weak tool-callers can reliably invoke. Integration name is `web`, requires no credentials, disabled by default.

Design doc: `docs/web-fetch.md`

## Test plan

- [x] Dispatch map parity tests (AllToolsCovered + NoOrphanHandlers)
- [x] HTML extraction test (strips script/style/nav/footer, preserves code blocks)
- [x] Private address rejection (localhost, 127.x, 10.x, 172.16.x, 192.168.x, ::1)
- [x] HTTPS-only enforcement
- [x] Truncation at 40K chars
- [x] HTTP error handling (404 → model-visible error)
- [x] Synonym group disjointness (TestSynonymGroups_Disjoint)
- [x] Config integration count updated (29→30)
- [x] `go build`, `go test ./...`, `go vet`, `golangci-lint` all pass
- [x] 88.2% test coverage on new code

🐘 Generated with Crush

### DIFF
--- a/docs/web-fetch.md
+++ b/docs/web-fetch.md
@@ -1,0 +1,201 @@
+# Spec: `fetch_url` Tool for Switchboard
+
+## Motivation
+
+Frontier models (Sonnet, Opus) have strong library and API knowledge baked into training and can
+fall back to web search. Open-source models like Gemma 4 26B don't. The knowledge gap shows up most
+clearly when a model needs to reason about a specific library version, an API reference, or a doc
+page that postdates or was underrepresented in its training data — it will hallucinate method
+signatures or miss version-specific behavior.
+
+`fetch_url` closes that gap by letting the model pull the actual source of truth into context at
+call time. It is also the minimum viable tool needed to make local models useful for the kind of
+work switchboard is designed for: referencing runbooks, API docs, internal wikis, GitHub raw files,
+and package documentation without copy-pasting.
+
+---
+
+## Tool Definition
+
+**Name:** `fetch_url`
+
+**Description (shown to model verbatim — keep this precise):**
+> Fetches the content of a URL and returns it as plain text. Use this to read documentation, API
+> references, README files, GitHub raw content, or any web page whose content you need to reason
+> about. Returns extracted readable text, not raw HTML.
+
+**Why the description matters:** Weak tool-callers like Gemma 4 rely heavily on the tool
+description to decide when and how to call it. Vague descriptions ("get web content") cause missed
+calls. Overly broad descriptions ("browse the internet") cause spurious ones. The description above
+scopes it to reference/doc lookup, which is the actual use case.
+
+---
+
+## Input Schema
+
+```go
+type FetchURLArgs struct {
+    URL     string `json:"url"      jsonschema:"required,description=The full URL to fetch (https only)"`
+    Timeout int    `json:"timeout"  jsonschema:"description=Request timeout in seconds (default: 10 max: 30)"`
+}
+```
+
+**Keep the schema flat and minimal.** Two fields max. Every additional parameter is a chance for a
+weak model to get the call wrong. No headers, no auth, no output format selector. If those are
+needed later, add a v2 tool rather than complicating this one.
+
+---
+
+## Output
+
+Return a single plain-text string. The model receives this as the tool result content.
+
+Structure:
+
+```
+Source: <url>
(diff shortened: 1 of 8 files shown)


integrations/webfetch/webfetch.go
```diff
@@ -0,0 +1,269 @@
...
+	allowPrivateAddr bool // testing only: bypass private address check
+}
+
+func New() mcp.Integration {
+	return &webfetch{
+		client: &http.Client{
+			Timeout: time.Duration(defTimeout) * time.Second,
+			CheckRedirect: func(req *http.Request, via []*http.Request) error {
+				if len(via) >= maxRedirect {
+					return fmt.Errorf("too many redirects (max %d)", maxRedirect)
```
acmacalister, line 38: The private address and HTTPS checks only run against the initial URL, not redirect targets. A server at `https://attacker.com/redirect` returning `302 → http://192.168.1.1/admin` (or `http://169.254.169.254/latest/meta-data` on AWS) would bypass both guards — the redirect is followed because `CheckRedirect` only enforces the count limit.

We should add host + scheme validation here:

```go
CheckRedirect: func(req *http.Request, via []*http.Request) error {
    if len(via) >= maxRedirect {
        return fmt.Errorf("too many redirects (max %d)", maxRedirect)
    }
    if req.URL.Scheme != "https" {
        return fmt.Errorf("redirect to non-https URL not allowed")
    }
    if isPrivateHost(req.URL.Hostname()) {
        return fmt.Errorf("redirect to private/local address not allowed")
    }
    return nil
},
```

And a test case like `TestFetchURL_RedirectToPrivateRejected` using `httptest.NewTLSServer` that responds with a `302` to a private IP would lock this in.

acmacalister (commented):
Good work overall — the security hardening, test coverage, and clean architecture here are solid. Two things worth addressing before merge: a redirect path that bypasses the SSRF guards, and a UTF-8 boundary issue in the truncation logic. One doc nit inline too.

integrations/webfetch/webfetch.go
```diff
@@ -0,0 +1,269 @@
...
+
+	body, err := io.ReadAll(io.LimitReader(resp.Body, maxBody+1))
+	if err != nil {
+		return &mcp.ToolResult{Data: "error: reading response: " + err.Error(), IsError: true}, nil
+	}
+
+	content := string(body)
+	truncated := false
+	if len(body) > maxBody {
+		content = content[:maxBody]
```
acmacalister, line 131: `content[:maxBody]` slices by byte index. For pages with multibyte UTF-8 characters (Chinese, Japanese, Arabic docs are exactly the kind of reference material this tool is meant to fetch), cutting at a byte boundary mid-rune produces an invalid UTF-8 string, which can cause JSON marshalling issues downstream and sends garbled output to the model.

A clean fix:

```go
if len(body) > maxBody {
    // find a valid UTF-8 boundary so we don't cut mid-rune
    n := maxBody
    for n > 0 && !utf8.RuneStart(content[n]) {
        n--
    }
    content = content[:n]
    truncated = true
}
```

Or, if rune-count semantics are preferred over byte-count:

```go
runes := []rune(content)
if len(runes) > maxBody {
    content = string(runes[:maxBody])
    truncated = true
}
```

Needs `"unicode/utf8"` in the import block for the first approach.

docs/web-fetch.md
```diff
@@ -0,0 +1,201 @@
+# Spec: `fetch_url` Tool for Switchboard
```
acmacalister, line 1: The doc still uses `fetch_url` throughout (the original design name) while the implemented tool is `web_fetch`. Since this is a spec doc rather than reference docs it's low-priority, but if it's intended to remain as living documentation the tool name, user-agent string, and config snippet examples will need updating to match what actually shipped.

PR #128, opened 2026-04-14
fix: remove hardcoded .rwx/ci.yml filter from RWX integration

aleksclark:
## Summary

- **`rwx_get_recent_runs`** was hardcoded to filter runs to `.rwx/ci.yml` only, making other workflow runs (e.g. `auto-deploy.yml`) invisible. Now returns all workflows by default with an optional `definition_path` parameter to filter.
- **`rwx_launch_ci_run`** was hardcoded to launch `.rwx/ci.yml`. Now accepts an optional `workflow` parameter (default: `.rwx/ci.yml`).
- Converted `rwxAPIBase` const to a `baseURL` struct field, matching the pattern used by other integrations (YNAB, Suno, Snowflake), enabling proper httptest coverage.
- Each run in `rwx_get_recent_runs` response now includes `definition_path` so callers can distinguish workflow types.

## Changes

| File | Change |
|------|--------|
| `runs.go` | Added `workflow` param to `launchCIRun`, `definition_path` param to `getRecentRuns`, replaced `rwxAPIBase` const with `r.baseURL` |
| `rwx.go` | Added `baseURL` field to struct, renamed const to `defaultRWXAPIBase` |
| `tools.go` | Updated tool descriptions and parameters for both tools |
| `compact_specs.go` | Added `definition_path` to compaction spec |
| `extras.go` | Updated `rwxAPIBase` → `r.baseURL` |
| `rwx_test.go` | Tests now use `baseURL: ts.URL` for real HTTP assertions; added subtests for definition_path filtering |

## Test plan

- [x] `make ci` passes (build, vet, test-race, lint, security, govulncheck)
- [x] New subtests verify: no filter returns all workflows, `definition_path=.rwx/ci.yml` filters correctly, `definition_path=.rwx/auto-deploy.yml` filters correctly
- [x] Existing tests upgraded from no-ops to real HTTP assertions (Healthy, GetRecentRuns, FetchRunStatus)

🐾 Generated with Crush

### DIFF
--- a/integrations/rwx/compact_specs.go
+++ b/integrations/rwx/compact_specs.go
@@ -8,8 +8,8 @@ import (
 
 var rawFieldCompactionSpecs = map[mcp.ToolName][]string{
 	// ── Runs ────────────────────────────────────────────────────────
-	// Handler: jsonResult(map{ref, count, runs: [{run_id, status, commit_sha, title, url}]})
-	mcp.ToolName("rwx_get_recent_runs"): {"ref", "count", "runs[].run_id", "runs[].status", "runs[].commit_sha", "runs[].title", "runs[].url"},
+	// Handler: jsonResult(map{ref, count, runs: [{run_id, status, commit_sha, title, definition_path, url}]})
+	mcp.ToolName("rwx_get_recent_runs"): {"ref", "count", "runs[].run_id", "runs[].status", "runs[].commit_sha", "runs[].title", "runs[].definition_path", "runs[].url"},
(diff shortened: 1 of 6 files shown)


acmacalister (approved):
Clean PR — builds, all tests pass with race detector, lint is clean. The refactor from const to struct field is the right call and the test upgrades (Healthy, GetRecentRuns, FetchRunStatus now hit real HTTP) are a genuine improvement over the previous no-op stubs. LGTM.

PR #129, opened 2026-04-14
fix: correct github_create_update_file content param description

daltoniam:
## Summary

- The `github_create_update_file` tool description said content should be "Base64-encoded file content", but `go-github`'s `RepositoryContentFileOptions.Content` field (`[]byte`) expects **raw unencoded bytes** — Go's `encoding/json` marshaler handles base64 encoding automatically when serializing `[]byte` to JSON.
- This misleading description caused AI agents to pre-encode file content to base64, resulting in **double-encoding** — files were written containing raw base64 text instead of source code (e.g. [teamcurri/curri#20000](https://github.com/teamcurri/curri/pull/20000)).

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./integrations/github/...` passes
- [ ] Verify an agent using `github_create_update_file` writes plain text correctly

🐰 Generated with Crush

### DIFF
--- a/integrations/github/tools.go
+++ b/integrations/github/tools.go
@@ -76,7 +76,7 @@ var tools = []mcp.ToolDefinition{
 	},
 	{
 		Name: mcp.ToolName("github_create_update_file"), Description: "Create or update a file in a repository",
-		Parameters: map[string]string{"owner": "Repository owner", "repo": "Repository name", "path": "File path", "message": "Commit message", "content": "Base64-encoded file content", "sha": "SHA of file being replaced (required for update)", "branch": "Target branch"},
+		Parameters: map[string]string{"owner": "Repository owner", "repo": "Repository name", "path": "File path", "message": "Commit message", "content": "Raw file content (plain text, not base64 — encoding is handled automatically)", "sha": "SHA of file being replaced (required for update)", "branch": "Target branch"},
 		Required:   []string{"owner", "repo", "path", "message", "content"},
 	},
 	{


integrations/github/tools.go
```diff
@@ -76,7 +76,7 @@ var tools = []mcp.ToolDefinition{
 	},
 	{
 		Name: mcp.ToolName("github_create_update_file"), Description: "Create or update a file in a repository",
-		Parameters: map[string]string{"owner": "Repository owner", "repo": "Repository name", "path": "File path", "message": "Commit message", "content": "Base64-encoded file content", "sha": "SHA of file being replaced (required for update)", "branch": "Target branch"},
+		Parameters: map[string]string{"owner": "Repository owner", "repo": "Repository name", "path": "File path", "message": "Commit message", "content": "Raw file content (plain text, not base64 — encoding is handled automatically)", "sha": "SHA of file being replaced (required for update)", "branch": "Target branch"},
```
acmacalister, line 79: Description fix is correct — `RepositoryContentFileOptions.Content` is tagged `// unencoded` in go-github and `encoding/json` handles the base64 automatically, so the old description was causing real double-encoding bugs.

One thing worth adding while we're here: there's no unit test that exercises `createOrUpdateFile` and asserts the `Content` bytes sent to the GitHub API. A small `httptest` test that captures the request body and checks `content` is the plain-text value (not double-encoded) would lock in the correct behavior permanently and prevent regressions. Not blocking — the fix itself is right.

acmacalister (approved):
Correct fix — the old description was actively causing double-encoding. One small suggestion inline about adding a unit test to lock in the behavior.

PR #131, opened 2026-04-14
feat: add botidentity integration for managing bot identities

aleksclark:
## Summary

- Adds a `botidentity` integration that manages bot identities across GitHub and Slack as a unified entity
- `create_bot` orchestrates both platforms: creates Slack app via manifest API, kicks off GitHub App manifest flow with auto-opening local web server
- Interactive browser-based logo generator using AWS Bedrock (Stability AI), with prompt iteration and auto-compression under 900KB
- Persistent bot inventory at `~/.config/switchboard/bots.json` with credential storage, logo paths, and platform-aware routing
- Browser auto-opens for Slack install, GitHub App approval, GitHub App installation, and logo preview
- 30+ tools covering bot CRUD, GitHub App management (installations, tokens, webhooks), Slack app management (manifests, token rotation), and logo generation

## Test plan

- [x] `make ci` passes (build, test-race, lint, gosec, govulncheck)
- [x] Dispatch parity tests (all tools covered, no orphan handlers)
- [x] Inventory persistence, file permissions (0600), path traversal protection
- [x] HTTP handler tests for GitHub and Slack APIs with httptest servers
- [x] End-to-end: created PizzaBot on both Slack and GitHub via teamcurri org
- [x] Logo generation via Bedrock Stable Image Core in us-west-2
- [ ] Verify deploy restart works with updated Makefile

🐘 Generated with Crush


integrations/botidentity/botidentity.go
```diff
@@ -0,0 +1,297 @@
...
+	if ak := creds["aws_access_key_id"]; ak != "" {
+		if sk := creds["aws_secret_access_key"]; sk != "" {
+			opts = append(opts, awsconfig.WithCredentialsProvider(
+				credentials.NewStaticCredentialsProvider(ak, sk, creds["aws_session_token"]),
+			))
+		}
+	}
+	cfg, err := awsconfig.LoadDefaultConfig(ctx, opts...)
+	if err != nil {
+		return nil
```
acmacalister, line 83: `awsconfig.LoadDefaultConfig` failing silently here means Bedrock will be unavailable with no indication why. The current code discards the error and leaves `bedrockClient` nil, so the caller only discovers the problem when they try to generate a logo — and gets a generic "AWS credentials not configured" message that doesn't reflect the actual failure.

Consider logging or surfacing the error, even as a warning, so the misconfiguration is visible at startup:

```go
cfg, err := awsconfig.LoadDefaultConfig(ctx, opts...)
if err != nil {
    // Bedrock unavailable — logo generation will be disabled
    return nil
}
```

Or, if the intent is always to treat AWS as optional, drop the error check entirely and call `LoadDefaultConfig` unconditionally (it won't fail unless the chain itself is broken). As written, the comment in Configure is implicit and the behavior surprising.

integrations/botidentity/botidentity.go
```diff
@@ -0,0 +1,297 @@
...
+			return nil, err
+		}
+		bodyReader = bytes.NewReader(data)
+	}
+
+	req, err := http.NewRequestWithContext(ctx, method, b.githubBaseURL+path, bodyReader)
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Authorization", "Bearer "+b.githubToken)
```
acmacalister, line 194: `b.githubToken` is read here without holding `b.mu`. The mutex is only taken in `slackPost` and `slackRotateToken`, but the GitHub token and other credentials are set in `Configure` and read throughout `githubDo`, `Healthy`, and the inventory handlers without synchronization.

If `Configure` can be called concurrently with `Execute` (e.g., on a credential refresh), this is a data race. Either extend `b.mu` to cover all credential fields and use it consistently, or document that `Configure` is only called before any requests.

integrations/botidentity/botidentity.go
```diff
@@ -0,0 +1,297 @@
...
+
+	data, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode == 429 || resp.StatusCode >= 500 {
+		re := &mcp.RetryableError{StatusCode: resp.StatusCode, Err: fmt.Errorf("slack API error (%d): %s", resp.StatusCode, string(data))}
+		re.RetryAfter = mcp.ParseRetryAfter(resp.Header.Get("Retry-After"))
+		return nil, re
+	}
```
acmacalister, line 280: `githubAppPEM` and `githubAppID` are stored in Configure but never read anywhere in the codebase. The GitHub App REST API endpoints under `/app/*` require JWT authentication (signed with the app's private key), not a user PAT — but `githubDo` always uses `githubToken` (a user PAT or installation token) for every request.

This means tools like `botidentity_gh_get_app`, `botidentity_gh_list_installations`, and `botidentity_gh_create_install_token` will return 401 for any caller authenticating as a GitHub App rather than a user. The PEM and App ID are accepted but silently unused.

Either add JWT signing support using the PEM (the `golang-jwt/jwt` package is already in go.mod), or update the tool descriptions to clarify that a user PAT with app-level scopes is required instead of a GitHub App private key.

integrations/botidentity/inventory_handlers.go
```diff
@@ -0,0 +1,496 @@
...
+		}
+		if v, _ := mcp.ArgStr(args, "github_app_slug"); v != "" {
+			e.GitHub.AppSlug = v
+		}
+	})
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+
+	return invGetBot(context.Background(), b, args)
```
acmacalister, line 306: `invUpdateBot` passes `context.Background()` to `invGetBot` instead of the original `ctx`. Since `invGetBot` ignores its context (`_ context.Context`), this is harmless today — but it's a subtle inconsistency. If `invGetBot` ever starts doing I/O, the caller's cancellation won't propagate. Worth threading the context through for correctness:

```go
return invGetBot(ctx, b, args)
```

integrations/botidentity/slack_apps.go
```diff
@@ -0,0 +1,258 @@
...
+		token = b.slackBotToken
+	}
+	if token == "" {
+		return mcp.ErrResult(fmt.Errorf("token is required — provide as argument or configure slack_bot_token (needs users.profile:write scope)"))
+	}
+
+	var imageData []byte
+
+	if p, _ := mcp.ArgStr(args, "path"); p != "" {
+		data, err := os.ReadFile(p)
```
acmacalister, line 198: The `path` argument to `slackSetBotIcon` is passed directly to `os.ReadFile` without any validation or restriction. An LLM (or a script) calling this tool with a path like `/etc/passwd` or `~/.config/switchboard/config.json` would happily read and upload that file to Slack.

We should validate that the path points to a regular image file in an expected location, or at least reject obvious out-of-bounds paths:

```go
cleanPath := filepath.Clean(p)
// Only allow paths within known safe directories, e.g. the logos dir or /tmp
if !strings.HasPrefix(cleanPath, b.inv.logoDir()) && !strings.HasPrefix(cleanPath, os.TempDir()) {
    return mcp.ErrResult(fmt.Errorf("path must be within the logos directory or temp dir"))
}
```

This is the same kind of path-traversal protection already implemented in `inventory.saveLogo`.

acmacalister (commented):
Solid integration — the dispatch parity tests, inventory mutex hygiene, and end-to-end browser flow are all well done. A few things worth looking at inline, the main one being that `githubAppPEM`/`githubAppID` are stored but never used for JWT signing.

integrations/botidentity/slack_apps.go
```diff
@@ -0,0 +1,258 @@
...
+		b.mu.Unlock()
+	}
+
+	return mcp.RawResult(data)
+}
+
+func slackSetBotIcon(ctx context.Context, b *botidentity, args map[string]any) (*mcp.ToolResult, error) {
+	token, _ := mcp.ArgStr(args, "token")
+	if token == "" {
+		token = b.slackBotToken
```
acmacalister, line 189: The existing comment on `githubDo` (line 194 of botidentity.go) flags the same issue for the GitHub token — same thing here with `b.slackBotToken`. It's read without holding `b.mu`, while `slackRefreshToken` and `slackConfigToken` are protected. Worth extending the mutex coverage to all credential fields for consistency:

```go
b.mu.Lock()
token = b.slackBotToken
b.mu.Unlock()
```

integrations/botidentity/botidentity.go
```diff
@@ -0,0 +1,297 @@
...
+		req.Header.Set("Content-Type", "application/json")
+	}
+
+	resp, err := b.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer func() { _ = resp.Body.Close() }()
+
+	data, err := io.ReadAll(resp.Body)
```
acmacalister, line 206: `io.ReadAll` on the GitHub API response body has no size cap. The existing S3 adapter sets a 10MB limit with `io.LimitReader` to prevent a misbehaving or malicious server from consuming unbounded memory. Same guard would be appropriate here:

```go
const maxRespBytes = 10 * 1024 * 1024
data, err := io.ReadAll(io.LimitReader(resp.Body, maxRespBytes))
```

Same applies to the Slack helper in `slackPost` and `slackSetBotIcon`.

config/config.go
```diff
@@ -118,6 +118,11 @@ var envMapping = map[string]map[string]string{
 	"acp": {
 		"config": "ACP_CONFIG",
 	},
+	"botidentity": {
+		"github_token":        "BOTIDENTITY_GITHUB_TOKEN",
```
acmacalister, line 122: The `envMapping` table only covers three of the ten credentials (`github_token`, `slack_config_token`, `slack_refresh_token`). Several important ones — `github_app_pem`, `github_app_id`, `slack_bot_token`, `aws_access_key_id`, `aws_secret_access_key`, `aws_session_token`, `aws_region` — have no env var mappings, so users can't inject them via environment variables. The convention in other integrations (e.g., `aws`) is to map all keys that a user might reasonably set via env. Worth rounding this out:

```go
"botidentity": {
    "github_token":          "BOTIDENTITY_GITHUB_TOKEN",
    "github_app_pem":        "BOTIDENTITY_GITHUB_APP_PEM",
    "github_app_id":         "BOTIDENTITY_GITHUB_APP_ID",
    "slack_config_token":    "BOTIDENTITY_SLACK_CONFIG_TOKEN",
    "slack_refresh_token":   "BOTIDENTITY_SLACK_REFRESH_TOKEN",
    "slack_bot_token":       "BOTIDENTITY_SLACK_BOT_TOKEN",
    "aws_access_key_id":     "BOTIDENTITY_AWS_ACCESS_KEY_ID",
    "aws_secret_access_key": "BOTIDENTITY_AWS_SECRET_ACCESS_KEY",
    "aws_session_token":     "BOTIDENTITY_AWS_SESSION_TOKEN",
    "aws_region":            "BOTIDENTITY_AWS_REGION",
},
```

acmacalister (commented):
Solid PR — dispatch parity, file-permission hardening, and credential masking are all done well. Few things to look at inline, mostly extensions of the mutex/read-limit patterns already flagged in the other review.

PR #133, opened 2026-04-15
fix(slack): resolve false EXPIRED badge and add multi-browser token extraction

acmacalister:
## Summary

- **Fix EXPIRED badge**: The Slack Token Setup page showed "EXPIRED" even when the token was connected and working. The badge now uses the actual `auth.test` result instead of a stale age-based heuristic.
- **Multi-browser extraction**: Auto-extraction now supports **Chrome, Brave, and the Slack desktop app** (all Chromium/Electron-based with the same LevelDB + encrypted SQLite cookie format). Introduces a `browserSource` abstraction so adding more browsers is a one-liner.
- **Slack desktop app support**: Reads tokens directly from `~/Library/Application Support/Slack/` using "Slack Safe Storage" Keychain entry — no browser needed at all.

## Test plan

- [x] `go test ./integrations/slack/` passes (new tests for `browserSources`, `listWorkspacesWithTokensFromBrowser`)
- [x] `go test ./...` passes
- [x] `go vet ./...` clean
- [x] `go build ./...` clean
- [ ] Manual: verify web UI shows "Healthy" badge when token is connected
- [ ] Manual: verify "Extract All Workspaces" pulls from Slack app when Chrome is not installed

🐰 Generated with Crush

### DIFF
--- a/integrations/slack/extract.go
+++ b/integrations/slack/extract.go
@@ -94,7 +94,7 @@ func ExtractAllFromBrowsersForWeb() (int, error) {
 	}
 
 	type extractedWorkspace struct {
-		chromeWorkspace
+		browserWorkspace
 		cookie string
 		source string
 	}
@@ -131,9 +131,9 @@ func ExtractAllFromBrowsersForWeb() (int, error) {
 				}
 				seen[ws.TeamID] = true
 				result = append(result, extractedWorkspace{
-					chromeWorkspace: ws,
-					cookie:          cookie,
-					source:          strings.ToLower(src.name),
+					browserWorkspace: ws,
+					cookie:           cookie,
+					source:           strings.ToLower(src.name),
 				})
 			}
 		}
--- a/integrations/slack/extract.go
+++ b/integrations/slack/extract.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"time"
 )
 
@@ -37,27 +38,33 @@ type TokenInfo struct {
 	Status         string    `json:"status"`
 }
 
-// ListWorkspacesFromChrome returns all Slack workspaces found in Chrome's localStorage.
+// ListWorkspacesFromBrowsers returns all Slack workspaces found in any supported
+// browser's localStorage (Chrome, Brave, Slack desktop app).
 // Exported for use by the web UI to let the user pick which workspace to extract.
-func ListWorkspacesFromChrome() ([]WorkspaceInfo, error) {
+func ListWorkspacesFromBrowsers() ([]WorkspaceInfo, error) {
 	if runtime.GOOS != "darwin" {
-		return nil, fmt.Errorf("chrome extraction is only available on macOS")
+		return nil, fmt.Errorf("browser extraction is only available on macOS")
 	}
-	return listWorkspacesFromChrome()
+	return listWorkspacesFromAllBrowsers()
 }
 
-// ExtractFromChromeForWeb triggers Chrome extraction and returns the result.
-// If teamID is non-empty, only the token for that workspace is extracted.
+// ListWorkspacesFromChrome is an alias kept for backward compatibility.
+func ListWorkspacesFromChrome() ([]WorkspaceInfo, error) {
+	return ListWorkspacesFromBrowsers()
+}
+
+// ExtractFromBrowserForWeb triggers extraction from all supported browsers and
+// returns the result. If teamID is non-empty, only that workspace's token is extracted.
 // This is exported for use by the web UI server.
-func ExtractFromChromeForWeb(teamID string) *ExtractResult {
+func ExtractFromBrowserForWeb(teamID string) *ExtractResult {
 	if runtime.GOOS != "darwin" {
 		return &ExtractResult{
 			Success: false,
-			Error:   "Chrome extraction is only available on macOS. Use the manual method below.",
+			Error:   "Browser extraction is only available on macOS. Use the manual method below.",
 		}
 	}
 
-	extracted, err := extractFromChromeWithError(teamID)
+	extracted, err := extractFromBrowserWithError(teamID)
 	if err != nil {
 		return &ExtractResult{
 			Success: false,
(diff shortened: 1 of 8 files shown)


integrations/slack/extract.go
```diff
@@ -68,39 +75,70 @@ func ExtractFromChromeForWeb(teamID string) *ExtractResult {
...
+			seen[ws.TeamID] = true
+			allWorkspaces = append(allWorkspaces, extractedWorkspace{
+				chromeWorkspace: ws,
+				cookie:          cookie,
+				source:          strings.ToLower(src.name),
+			})
 		}
 	}
-	workspaces := listWorkspacesWithTokensFromChrome(profiles)
 	extractMu.Unlock()
```
acmacalister: `extractMu.Unlock()` is called manually here rather than being deferred. If any of the inner calls (`findBrowserProfiles`, `browserKeychainPassword`, `extractCookieFromBrowser`, `listWorkspacesWithTokensFromBrowser`) panic, the mutex leaks and the next extraction deadlocks. Every other lock site in this file uses `defer extractMu.Unlock()`.

The tricky part is that `defer` can't be used mid-function when you want to release before subsequent work. The cleanest fix is to extract the locked section into a closure:

```go
allWorkspaces := func() []extractedWorkspace {
    extractMu.Lock()
    defer extractMu.Unlock()
    // ... loop body ...
    return result
}()
```

integrations/slack/extract_test.go
```diff
@@ -35,6 +35,57 @@ func TestCanExtractFromChrome(t *testing.T) {
...
+	assert.True(t, names["Chrome"], "should include Chrome")
+	assert.True(t, names["Brave"], "should include Brave")
+	assert.True(t, names["Slack"], "should include Slack")
+}
+
+func TestFindBrowserProfiles_NonProfileSource(t *testing.T) {
+	dir := t.TempDir()
+	src := browserSource{name: "Test", dataDir: "nonexistent", hasProfiles: false}
+	// Point to a real temp dir for the non-profile case
+	srcWithRealDir := browserSource{name: "Test", dataDir: "", hasProfiles: false}
```
acmacalister: `TestFindBrowserProfiles_NonProfileSource` doesn't assert anything — every variable is immediately suppressed with `_ = ...`. The `hasProfiles: false` path through `findBrowserProfiles` is genuinely testable: a non-existent directory should return an error, and an existing directory should return a single-element slice. This test registers as passing but provides zero coverage of the actual logic.

acmacalister (commented):
Nice refactor — the `browserSource` abstraction is clean and backward-compat aliases keep callsites untouched. Build passes, all tests green, lint clean.

One thing not in the diff but worth a follow-up: `handleSlackExtractChrome` in `web/web.go` still hardcodes `"chrome"` as the `CredKeyTokenSource` credential (line 670) and the success redirect still says "from Chrome" (line 673). Neither was part of this PR's diff, but they'll misattribute extractions from Brave or the Slack desktop app. Could be a quick cleanup in a follow-up commit.

integrations/slack/tokens.go
```diff
@@ -331,42 +369,36 @@ func extractFromChromeWithError(teamID string) (*chromeTokens, error) {
...
-		msg := "found token but no xoxd-* cookie in Chrome cookie store."
-		if lastCookieErr != nil {
-			msg += " " + lastCookieErr.Error()
-		}
 		return nil, fmt.Errorf("%s", msg)
 	}
 
-	return &chromeTokens{token: token, cookie: cookie}, nil
+	sourceName := strings.ToLower(src.name)
+	return &browserTokens{token: token, cookie: cookie, source: sourceName}, nil
```
acmacalister, line 383: The old code returned an error if either token or cookie was missing — now we return success whenever at least one is present. That changes the behavior in `tryRefreshWorkspace`: if a browser has the cookie but no matching xoxc-* token, `updateTokens` gets called with `token=""`, which silently overwrites the stored token with an empty string.

Since the Slack desktop app source (`hasProfiles: false`) can't produce a cookie but can produce a token, and Chrome can produce both, this could manifest as a workspace losing its token if extraction order ever produces a partial result. Worth guarding in `tryRefreshWorkspace` before calling `updateTokens`:

```go
if extracted.token == "" {
    return false
}
```

acmacalister (commented):
Good work overall — the `browserSource` abstraction is clean and the backward-compat aliases make this easy to review. One thing caught my eye inline.

integrations/slack/slack.go
```diff
@@ -245,8 +245,8 @@ func (s *slackIntegration) tryRefreshWorkspace(teamID string) bool {
 	if s.tryRefreshViaCookieForTeam(teamID) {
 		return true
 	}
-	extracted := extractFromChrome(teamID)
-	if extracted == nil {
+	extracted := extractFromBrowser(teamID)
+	if extracted == nil || extracted.token == "" {
 		return false
 	}
 	s.store.updateTokens(teamID, extracted.token, extracted.cookie)
```
acmacalister: `updateTokens` unconditionally overwrites `ws.Cookie` with whatever `extracted.cookie` is — which can be `""`. This matters for the Slack desktop app source: the Slack app's data directory may not have a `Cookies` SQLite file in the same location that browsers do, so `extractCookieFromBrowser` returns an error and `extracted.cookie` stays empty. If a background refresh reaches the Slack source (because Chrome and Brave both failed), the stored cookie gets erased, breaking future cookie-injected API calls even though a valid cookie was originally set.

Worth preserving the existing cookie when the fresh extraction didn't produce one:

```go
existingCookie := extracted.cookie
if existingCookie == "" {
    if ws := s.store.getWorkspace(teamID); ws != nil {
        existingCookie = ws.Cookie
    }
}
s.store.updateTokens(teamID, extracted.token, existingCookie)
```

web/web.go
```diff
@@ -600,11 +600,16 @@ func (w *WebServer) handleSlackSetup(rw http.ResponseWriter, r *http.Request) {
...
+
 	page := w.pageData(r, "Slack Setup", "/integrations")
 	data := pages.SlackSetupData{
 		HasToken:       info.HasToken,
 		HasCookie:      info.HasCookie,
-		TokenStatus:    info.Status,
+		TokenStatus:    tokenStatus,
 		TokenAge:       info.AgeHours,
 		TokenSource:    tokenSource,
 		CanAutoExtract: slackInt.CanExtractFromChrome(),
```
acmacalister: This callsite still uses `CanExtractFromChrome()` (the old alias) while the rest of `handleSlackSetup` was updated. Line 644 in `handleSlackListWorkspaces` also calls `ListWorkspacesFromChrome()`. Both aliases work fine, but updating them to `CanExtractFromBrowser()` and `ListWorkspacesFromBrowsers()` would keep the file consistent with the intent of this PR.

acmacalister (commented):
Clean PR overall — the `browserSource` abstraction is well-designed and the backward-compat aliases make the rename safe. Build passes, all tests green, lint clean. Two things inline.

web/templates/pages/slack_setup.templ
```diff
@@ -122,12 +122,12 @@ templ SlackSetup(page layouts.PageData, data SlackSetupData) {
...
 			<div class="card">
-				<div class="card-title" style="margin-bottom: 0.75rem;">Option 1: Auto-Extract from Chrome</div>
+				<div class="card-title" style="margin-bottom: 0.75rem;">Option 1: Auto-Extract from Browser</div>
 				<p class="slack-desc">
-					Automatically extracts session tokens and cookies for <strong>all workspaces</strong> from a running Chrome browser with Slack open. Requires macOS and Chrome with a Slack tab at <code>app.slack.com</code>.
+					Automatically extracts session tokens and cookies for <strong>all workspaces</strong> from Chrome, Brave, or the Slack desktop app. Requires macOS with at least one of these installed and signed in to Slack.
 				</p>
 				<form method="POST" action="/api/slack/extract-chrome" style="margin-top: 0.75rem;">
-					<button type="submit" class="btn">Extract All Workspaces from Chrome</button>
+					<button type="submit" class="btn">Extract All Workspaces</button>
```
acmacalister, line 130: The form action one line above still posts to `/api/slack/extract-chrome`, and the route registration in `web.go:61` and handler name `handleSlackExtractChrome` weren't updated either. The callsites and method names were all migrated in this PR, but the URL itself got left behind. Worth renaming the route and handler to `extract-browsers`/`handleSlackExtractBrowsers` to match — it's a path that shows up in browser history and server logs, so the Chrome specificity there will keep looking wrong.

integrations/slack/tokens.go
```diff
@@ -508,11 +539,9 @@ type chromeWorkspace struct {
...
 }
 
-// listWorkspacesWithTokensFromChrome scans all Chrome profiles and returns
-// every Slack workspace with its xoxc-* token. Unlike listWorkspacesFromChrome,
-// this includes the token so callers don't need to re-read the LevelDB.
-// Must be called with extractMu held.
-func listWorkspacesWithTokensFromChrome(profiles []string) []chromeWorkspace {
+// listWorkspacesWithTokensFromBrowser scans profiles and returns every Slack
+// workspace with its xoxc-* token. Must be called with extractMu held.
+func listWorkspacesWithTokensFromBrowser(profiles []string) []chromeWorkspace {
```
acmacalister: The return type here is still `[]chromeWorkspace`, and the `chromeWorkspace` type itself (a few lines up) wasn't renamed. Since it's used generically across Chrome, Brave, and the Slack app now, `browserWorkspace` would be more accurate. Same for the stale “Chrome’s localStorage” in the doc comments on `extractTokenFromLevelDB` and `readSlackLocalConfig`.

acmacalister (commented):
Solid work — all four previous review items are addressed cleanly, and the `browserSource` abstraction is a genuine improvement. Build, tests, and lint all pass. Two small consistency nits inline; nothing blocking.

daltoniam (approved):

web/web.go
```diff
@@ -600,14 +600,19 @@ func (w *WebServer) handleSlackSetup(rw http.ResponseWriter, r *http.Request) {
 		tokenSource = info.Source
 	}
 
+	tokenStatus := info.Status
+	if healthy && tokenStatus != "healthy" {
```
acmacalister, line 604: This correctly un-expires a token when `auth.test` passes, but the reverse isn't handled: if `healthy == false` but the token is less than 6 hours old, `info.Status` is already `"healthy"` from the age heuristic, so the badge stays green even though the live check just failed.

Since we're doing a real `auth.test` call here now, trusting it in both directions makes sense:

```go
tokenStatus := info.Status
if info.HasToken {
    if healthy {
        tokenStatus = "healthy"
    } else if tokenStatus == "healthy" {
        tokenStatus = "warning"
    }
}
```

acmacalister (commented):
Good work overall — the mutex-with-defer closure, cookie-preservation, and empty-token guard are all solid improvements. One inline thing on the badge fix, plus one stale string outside the diff: `slack.go:75` still hardcodes `Source: "chrome"` when seeding workspaces from `listWorkspacesFromAllBrowsers()` — those workspaces could have come from Brave or the Slack app, so `"browser"` would be more accurate (and consistent with how `ExtractAllFromBrowsersForWeb` labels them).

PR #134, opened 2026-04-16
feat: add plugin marketplace with discovery, install, and auto-update

aleksclark:
## Summary

- Add a full plugin marketplace system for discovering, installing, and managing WASM plugins from the web UI
- Implement versioned plugin ABI (v1) with **required** `metadata()` export for self-describing plugins
- Add manifest format spec, official plugin registry, auto-update system, and comprehensive e2e test suite

### Plugin ABI
- `metadata()` is now a **required** WASM export (alongside `name`, `tools`, `configure`, `execute`, `healthy`)
- Returns JSON with name, version, ABI version, description, author, homepage, license, capabilities
- Rust guest SDK updated with `PluginMetadata` struct and `leaked_metadata()` helper
- Example plugin rebuilt with metadata export

### Manifest Format (Schema v1)
- JSON manifest with plugin listings, version ranges, download URLs, SHA256 checksums
- ABI min/max compatibility filtering
- Official manifest committed at `plugins/manifest.json`

### Marketplace Manager (`marketplace/`)
- Fetch and parse manifests from multiple sources
- Install plugins from manifest, URL, or browser upload
- SHA256 verification on every download
- Periodic auto-update with configurable interval (default 6h)
- Per-plugin and global auto-update toggles

### Web UI (`/plugins`)
- Browse available plugins from configured manifests
- One-click install, update, and uninstall
- Install from arbitrary URL
- Upload `.wasm` directly from browser
- Add/remove third-party manifest sources
- Auto-update toggle with manual "Check Now" button

### Config
- New `marketplace` section in `config.json` with manifest sources, installed plugins, auto-update settings
- Persisted through config manager round-trips

### Tests
- 25 unit tests in `marketplace_test.go`
- 28 e2e tests in `marketplace_e2e_test.go` using:
  - Real `example.wasm` binary (no fakes)
  - Real wazero WASM runtime (load + execute)
  - Real filesystem (install/uninstall writes)
  - Real web handlers via httptest
  - Real config persistence (JSON round-trip)
  - Only mock: `httptest.Server` for manifest/WASM HTTP serving

## Test plan

- [x] All 53 marketplace tests pass
- [x] WASM binary rebuilt with required `metadata()` export
- [x] Existing tests unaffected

🐻 Generated with [Claude Code](https://claude.ai/code)


acmacalister (commented):
Solid work — the e2e test suite with a real wazero runtime is especially good. Found a few things to look at inline, the most important being an unguarded map access in `CheckForUpdates`/`UpdatePlugin` and a lexicographic version comparison that breaks for plugins with double-digit minor versions.

marketplace/marketplace.go
```diff
@@ -0,0 +1,774 @@
...
+	m.mu.RUnlock()
+
+	lookup := make(map[string]*PluginVersion)
+	lookupURL := make(map[string]string) // name -> manifest URL
+	for _, mf := range manifests {
+		for _, pl := range mf.Plugins {
+			if best := bestCompatibleVersion(pl.Versions, ABIVersion); best != nil {
+				lookup[pl.Name] = best
+				for _, src := range m.cfg.ManifestSources {
+					if cached, ok := m.manifests[src.URL]; ok && cached == mf {
```
acmacalister: `m.cfg.ManifestSources` and `m.manifests` are read here without holding the lock — the `RUnlock()` on line ~403 released it before this loop, so a concurrent `AddManifestSource` or `RemoveManifestSource` call would race.

The simplest fix is to snapshot the map before releasing the lock:

```go
m.mu.RLock()
installed := make([]InstalledPlugin, len(m.cfg.InstalledPlugins))
copy(installed, m.cfg.InstalledPlugins)
// snapshot manifest cache while still holding the read lock
cachedManifests := make(map[string]*Manifest, len(m.manifests))
for k, v := range m.manifests {
    cachedManifests[k] = v
}
sourcesSnap := make([]ManifestSource, len(m.cfg.ManifestSources))
copy(sourcesSnap, m.cfg.ManifestSources)
m.mu.RUnlock()
```

The same issue appears in `UpdatePlugin` around line 477–478.

marketplace/marketplace.go
```diff
@@ -0,0 +1,774 @@
...
+	}
+	return sanitizeFilename(name)
+}
+
+func bestCompatibleVersion(versions []PluginVersion, abi int) *PluginVersion {
+	var best *PluginVersion
+	for i := range versions {
+		v := &versions[i]
+		if v.ABIMin <= abi && v.ABIMax >= abi {
+			if best == nil || v.Version > best.Version {
```
acmacalister: Version comparison uses plain `>` on strings, which is lexicographic, not semantic. That means `"1.9.0" > "1.10.0"` evaluates to true, so `bestCompatibleVersion` would pick `1.9.0` over `1.10.0` — the wrong winner once a plugin crosses a minor version boundary.

The project already uses `github.com/Masterminds/semver/v3` (visible in `go.sum`). Using `semver.MustParse` here would give correct ordering with minimal change:

```go
import semver "github.com/Masterminds/semver/v3"

if best == nil {
    best = v
    continue
}
bv, err1 := semver.NewVersion(best.Version)
vv, err2 := semver.NewVersion(v.Version)
if err1 != nil || err2 != nil {
    // fall back to lexicographic on parse error
    if v.Version > best.Version {
        best = v
    }
    continue
}
if vv.GreaterThan(bv) {
    best = v
}
```

web/web_marketplace.go
```diff
@@ -0,0 +1,310 @@
...
+	if err := w.marketplace.RemoveManifestSource(url); err != nil {
+		http.Redirect(rw, r, "/plugins?error="+urlEncode(err.Error()), http.StatusSeeOther)
+		return
+	}
+
+	http.Redirect(rw, r, "/plugins?success=Manifest+source+removed.", http.StatusSeeOther)
+}
+
+func urlEncode(s string) string {
+	return strings.ReplaceAll(s, " ", "+")
```
acmacalister, line 335: `urlEncode` only replaces spaces — it leaves `&`, `=`, `#`, `?`, and other reserved URL characters untouched. An error message like `"connection refused & retry=false"` would produce a query string with a second parameter, silently truncating the error text shown to the user (and potentially introducing unexpected query params).

The standard library has this covered:

```go
import "net/url"

// in each handler:
http.Redirect(rw, r, "/plugins?error="+url.QueryEscape(err.Error()), http.StatusSeeOther)
```

Or keep the helper but use `url.QueryEscape` inside it.

wasm/module.go
```diff
@@ -6,6 +6,7 @@ import (
 	"fmt"
 
 	mcp "github.com/daltoniam/switchboard"
+	"github.com/daltoniam/switchboard/marketplace"
```
acmacalister, line 9: The `wasm` package is a low-level WASM runtime adapter; importing `marketplace` here pulls in application-layer concerns (manifests, installs, config persistence) as a transitive dependency of every binary that uses the runtime — and it creates a layering issue where the infrastructure layer knows about the application layer.

`PluginMetadata` is really a domain type, so it belongs in `mcp.go` alongside `WasmModuleConfig` and the other shared config structs. That way `wasm` imports only the root `mcp` package (which it already does) and `marketplace` can import `mcp` for the type — keeping the dependency arrows pointing the right way.

marketplace/marketplace.go
```diff
@@ -0,0 +1,774 @@
...
+	m.mu.Lock()
+	m.cfg.LastCheck = time.Now().UTC().Format(time.RFC3339)
+	for i, ip := range m.cfg.InstalledPlugins {
+		if latest, ok := lookup[ip.Name]; ok {
+			m.cfg.InstalledPlugins[i].LatestVersion = latest.Version
+		}
+	}
+	m.mu.Unlock()
+
+	_ = m.save()
```
acmacalister: The save error after updating `LastCheck` and `LatestVersion` is silently discarded with `_ = m.save()`. If the config write fails the user won't know, and the next restart will show stale timestamps.

Even just logging it keeps the behavior consistent with the rest of the file:

```go
if err := m.save(); err != nil {
    log.Printf("WARN: failed to persist update-check state: %v", err)
}
```

cmd/server/main.go
```diff
@@ -306,7 +307,72 @@ func runServer(stdioMode bool, port int, discoverAll bool) {
...
+				AutoUpdate:    ip.AutoUpdate,
+				LatestVersion: ip.LatestVersion,
+			})
+		}
+		cfgNow := cfgMgr.Get()
+		cfgNow.Marketplace = mc
+		return cfgMgr.Update(cfgNow)
+	})
+
+	if mpCfg.AutoUpdate {
```
acmacalister: The auto-update loop is only started if `mpCfg.AutoUpdate` is `true` at server startup (line 370). That means toggling auto-update on via the web UI takes effect in the config file, but the background goroutine never gets launched until the next restart — the UI gives no indication of this.

Worth calling out in the docs at minimum. A more robust fix would be to always start the loop and let `StartAutoUpdateLoop` read `m.cfg.AutoUpdate` dynamically on each tick — which it already does at line 616 — so the goroutine itself correctly no-ops when disabled. Starting it unconditionally would make the toggle actually work at runtime.

marketplace/marketplace.go
```diff
@@ -0,0 +1,805 @@
...
+		Name:        name,
+		Version:     "unknown",
+		InstalledAt: time.Now().UTC().Format(time.RFC3339),
+		Path:        destPath,
+		SHA256:      hash,
+		AutoUpdate:  false,
+	}
+
+	m.mu.Lock()
+	m.cfg.InstalledPlugins = append(m.cfg.InstalledPlugins, ip)
```
acmacalister, line 332: `InstallFromURL` appends unconditionally — if someone installs the same URL twice (or a URL that resolves to a plugin with the same name as an existing one), `InstalledPlugins` ends up with duplicate entries. `downloadAndInstall` does the right thing with an upsert; this path should too:

```go
m.mu.Lock()
var updated bool
for i, existing := range m.cfg.InstalledPlugins {
    if existing.Name == ip.Name {
        m.cfg.InstalledPlugins[i] = ip
        updated = true
        break
    }
}
if !updated {
    m.cfg.InstalledPlugins = append(m.cfg.InstalledPlugins, ip)
}
m.mu.Unlock()
```

Same issue at line 364 for `InstallFromBytes`.

wasm/module.go
acmacalister, line 9: `wasm` importing `marketplace` to type the return value of `Metadata()` inverts the natural layering — the WASM execution layer shouldn't depend on the distribution/marketplace layer.

`PluginMetadata` has no marketplace-specific logic; it's a domain type describing a plugin. It already fits in the root `mcp` package alongside `WasmModuleConfig` and `MarketplaceConfig`, which both packages import today. Moving it there would let `Metadata()` return `*mcp.PluginMetadata` and remove the `wasm` → `marketplace` coupling.

web/web_marketplace.go
```diff
@@ -0,0 +1,310 @@
...
+			Path:          ip.Path,
+			InstalledAt:   ip.InstalledAt,
+			AutoUpdate:    ip.AutoUpdate,
+			LatestVersion: ip.LatestVersion,
+			UpdateAvail:   ip.LatestVersion != "" && ip.LatestVersion != ip.Version,
+		}
+		data.Installed = append(data.Installed, entry)
+	}
+
+	if len(cfg.ManifestSources) > 0 {
```
acmacalister, line 51: Every `GET /plugins` page load fetches all configured manifests synchronously, blocking for up to 15 seconds. With two slow or unreliable sources the page will feel broken to users. The `Manager.manifests` cache map already exists — consider giving it a TTL so repeated page loads return cached data and trigger a background refresh, or at minimum skip the network fetch when `LastCheck` is recent enough (the field is already there).

acmacalister (commented):
Solid feature — well-structured, great test coverage with real WASM execution, and both issues from the last review round are fixed. Three things inline worth a look before merging.

acmacalister (commented):
Solid PR — builds clean, all 45 test packages pass with the race detector, lint is clear, and the e2e suite using real wazero + real WASM is the right way to test this. Three things inline worth a look.

marketplace/marketplace.go
```diff
@@ -0,0 +1,809 @@
...
+
+	m.mu.RLock()
+	installed := make([]InstalledPlugin, len(m.cfg.InstalledPlugins))
+	copy(installed, m.cfg.InstalledPlugins)
+	m.mu.RUnlock()
+
+	m.mu.RLock()
+	sources := make([]ManifestSource, len(m.cfg.ManifestSources))
+	copy(sources, m.cfg.ManifestSources)
+	manifestCache := make(map[string]*Manifest, len(m.manifests))
```
acmacalister, line 414: There are two back-to-back `mu.RLock` / `mu.RUnlock` pairs here — one for `installed` (lines 402–405) and a second for `sources` + `manifestCache` (lines 407–414). Both are immediately released before any shared work begins, so they could just as well be a single lock scope, and having two separate acquisitions is easy to misread as intentional double-check locking.

Merging them removes ambiguity:

```go
m.mu.RLock()
installed := make([]InstalledPlugin, len(m.cfg.InstalledPlugins))
copy(installed, m.cfg.InstalledPlugins)
sources := make([]ManifestSource, len(m.cfg.ManifestSources))
copy(sources, m.cfg.ManifestSources)
manifestCache := make(map[string]*Manifest, len(m.manifests))
for k, v := range m.manifests {
    manifestCache[k] = v
}
m.mu.RUnlock()
```

web/web_marketplace.go
acmacalister, line 335: `urlEncode` only swaps spaces for `+`, but error strings from the marketplace can contain `&`, `=`, `?`, `#`, or other characters that would corrupt the query parameter and silently truncate or misparse the error message in the template. `url.QueryEscape` handles all of this correctly:

```go
func urlEncode(s string) string {
    return url.QueryEscape(s)
}
```

(just add `"net/url"` to the imports)

cmd/server/main.go
```diff
@@ -306,7 +305,70 @@ func runServer(stdioMode bool, port int, discoverAll bool) {
...
+				Path:          ip.Path,
+				SHA256:        ip.SHA256,
+				AutoUpdate:    ip.AutoUpdate,
+				LatestVersion: ip.LatestVersion,
+			})
+		}
+		cfgNow := cfgMgr.Get()
+		cfgNow.Marketplace = mc
+		return cfgMgr.Update(cfgNow)
+	})
```
acmacalister, line 390: This save callback is ~30 lines of field-by-field translation between `marketplace.Config` and `mcp.MarketplaceConfig`, mirroring the ~25-line init block above it (lines 310–336). Three parallel struct pairs with identical fields.

The root cause is having two type sets for the same data. If `mcp.MarketplaceConfig` (and its sub-types) can't be collapsed into `marketplace.Config` directly, at minimum wrapping this in a pair of helpers (`toMCPMarketplaceConfig` / `fromMCPMarketplaceConfig`) would keep `main.go` readable and make the translation testable in isolation.

wasm/runtime.go
```diff
@@ -113,6 +114,10 @@ func (m *Module) resolveExports() error {
 	if m.fnHealthy == nil {
 		return fmt.Errorf("wasm: module does not export 'healthy'")
 	}
+	m.fnMetadata = m.mod.ExportedFunction("metadata")
+	if m.fnMetadata == nil {
+		return fmt.Errorf("wasm: module does not export 'metadata'")
```
acmacalister, line 119: Making `metadata()` a hard required export is a breaking ABI change — any WASM plugin compiled before this PR will now fail to load silently (the `WARN` log in `loadWasmModule` is easy to miss). There's no migration note in the docs or a graceful fallback.

Since `Module.Metadata()` already returns `nil` when `fnMetadata == nil`, it seems safe to treat the export as optional during `resolveExports` and just log a hint if it's absent:

```go
m.fnMetadata = m.mod.ExportedFunction("metadata")
if m.fnMetadata == nil {
    // metadata() is optional; older plugins without it still work,
    // credential_keys discovery and PlainTextKeys just won't be available.
}
```

If the intent is to require it going forward, the docs in `docs/plugin-marketplace.md` should call out the migration step (rebuild with the updated SDK) and ideally the WARN log should become something more prominent so users know why their plugin stopped loading.

marketplace/marketplace.go
```diff
@@ -0,0 +1,809 @@
...
+	copy(sources, m.cfg.ManifestSources)
+	m.mu.RUnlock()
+
+	var results []*Manifest
+	var errs []string
+	for _, src := range sources {
+		if !src.Enabled {
+			continue
+		}
+		manifest, err := m.FetchManifest(ctx, src.URL)
```
acmacalister, line 196: The `manifests` cache map (written in `FetchManifest` at line 177) is never consulted here before making the request — `FetchAllManifests` always fires a network call for every enabled source. The existing comment from another reviewer flags the synchronous page-load latency, and adding a TTL would fix both that and this. A simple approach:

```go
type cachedManifest struct {
    manifest  *Manifest
    fetchedAt time.Time
}
// replace manifests map[string]*Manifest with:
// manifests map[string]cachedManifest
```

Then in `FetchAllManifests`, check `if c, ok := m.manifests[src.URL]; ok && time.Since(c.fetchedAt) < cacheTTL { results = append(results, c.manifest); continue }`. Even a 60-second TTL would eliminate the latency spike on repeated page loads.

acmacalister (commented):
Solid work — the marketplace architecture is well-designed, and 53 tests with real WASM execution is great coverage. Two things the existing comments haven't flagged inline.

web/web_marketplace.go
acmacalister, line 335: `urlEncode` only swaps spaces for `+` — characters like `&`, `=`, and `#` pass through unchanged. An error message containing `&` would silently inject extra query parameters into the redirect URL, and `#` would truncate it entirely. `url.QueryEscape` (or `url.Values`) is the right tool here:

```go
func urlEncode(s string) string {
    return url.QueryEscape(s)
}
```

marketplace/marketplace.go
```diff
@@ -0,0 +1,809 @@
...
+	m.mu.RUnlock()
+
+	m.mu.RLock()
+	sources := make([]ManifestSource, len(m.cfg.ManifestSources))
+	copy(sources, m.cfg.ManifestSources)
+	manifestCache := make(map[string]*Manifest, len(m.manifests))
+	for k, v := range m.manifests {
+		manifestCache[k] = v
+	}
+	m.mu.RUnlock()
```
acmacalister, line 418: There are two back-to-back `RLock/RUnlock` pairs here that read from the same protected fields. This is fine for correctness but slightly wasteful — they can be merged into a single lock section, which is also cleaner to reason about under data race analysis:

```go
m.mu.RLock()
installed := make([]InstalledPlugin, len(m.cfg.InstalledPlugins))
copy(installed, m.cfg.InstalledPlugins)
sources := make([]ManifestSource, len(m.cfg.ManifestSources))
copy(sources, m.cfg.ManifestSources)
manifestCache := make(map[string]*Manifest, len(m.manifests))
for k, v := range m.manifests {
    manifestCache[k] = v
}
m.mu.RUnlock()
```

wasm/module.go
acmacalister, line 9: The `wasm` package now imports `marketplace` so that `Metadata()` can return `*marketplace.PluginMetadata`. This inverts the intended dependency direction: `wasm` is a runtime adapter, `marketplace` is a domain package — and `marketplace` imports nothing from `wasm`. Introducing this edge means any future attempt to use `wasm` types from `marketplace` would create a cycle.

The cleaner home for `PluginMetadata` is `mcp.go` (the root domain package), where the other shared types like `MarketplaceConfig` already live. `wasm.Module.Metadata()` would then return `*mcp.PluginMetadata`, and `marketplace` could reference it from the root package the same way it would any other `mcp.*` type.

acmacalister (commented):
Solid PR — the architecture is well thought out, tests are thorough, and SHA256 verification on every download is a nice touch. Three things worth a look inline.

wasm/loader.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+	}
+
+	l.mu.Lock()
+	if old, ok := l.modules[mod.Name()]; ok {
+		l.reg.Unregister(mod.Name())
+		old.Close(ctx) //nolint:errcheck
+		delete(l.modules, mod.Name())
+	}
+	l.mu.Unlock()
```
acmacalister, line 90: There's a TOCTOU window here: the lock is released on line 90 after removing the old module, but `reg.Register` (line 92) and the second `l.mu.Lock` / `l.modules[mod.Name()] = mod` (lines 106-108) happen outside it. Two concurrent `LoadPlugin` calls for the same plugin name can both pass the old-module check, then race to register — the second caller will get "already registered" and its module will be closed, but the winning caller's module ends up tracked while the registry state may be inconsistent.

Holding `l.mu` across the register+track sequence fixes this, though `reg.Register` and the module's own lock require care. One approach: move the entire replace-and-register sequence inside a single held section, or use a per-plugin lock/channel to serialize installs:

```go
// Hold the lock across the full replace-register-track sequence:
l.mu.Lock()
if old, ok := l.modules[mod.Name()]; ok {
    l.reg.Unregister(mod.Name())
    old.Close(ctx) //nolint:errcheck
}
if err := l.reg.Register(mod); err != nil {
    l.mu.Unlock()
    mod.Close(ctx) //nolint:errcheck
    return fmt.Errorf("register WASM module %q: %w", path, err)
}
l.modules[mod.Name()] = mod
l.mu.Unlock()
```

marketplace/marketplace.go
```diff
@@ -0,0 +1,809 @@
...
+	}
+
+	ip := InstalledPlugin{
+		Name:        name,
+		Version:     ver.Version,
+		ManifestURL: manifestURL,
+		InstalledAt: time.Now().UTC().Format(time.RFC3339),
+		Path:        destPath,
+		SHA256:      hash,
+		AutoUpdate:  m.cfg.AutoUpdate,
```
acmacalister, line 698: `m.cfg.AutoUpdate` is read here before the mutex is acquired (the lock comes three lines later at the `m.mu.Lock()` on line 701). Under the race detector this is a data race: any goroutine in `StartAutoUpdateLoop` can be reading or writing `m.cfg.AutoUpdate` concurrently.

The simplest fix is to read it inside the lock that's already being acquired for the slice update:

```go
m.mu.Lock()
autoUpdate := m.cfg.AutoUpdate
var updated bool
for i, existing := range m.cfg.InstalledPlugins {
    if existing.Name == name {
        ip.AutoUpdate = autoUpdate
        m.cfg.InstalledPlugins[i] = ip
        updated = true
        break
    }
}
if !updated {
    ip.AutoUpdate = autoUpdate
    m.cfg.InstalledPlugins = append(m.cfg.InstalledPlugins, ip)
}
m.mu.Unlock()
```

web/web_marketplace.go
```diff
@@ -0,0 +1,336 @@
...
+
+	ctx, cancel := context.WithTimeout(r.Context(), 2*time.Minute)
+	defer cancel()
+	ip, err := w.marketplace.InstallFromURL(ctx, url)
+	if err != nil {
+		http.Redirect(rw, r, "/plugins?error=Install+failed:+"+urlEncode(err.Error()), http.StatusSeeOther)
+		return
+	}
+
+	w.liveLoadPlugin(r.Context(), ip.Path, "")
```
acmacalister, line 130: `liveLoadPlugin` is called with `r.Context()` here (and on lines 102, 165, 215) after the preceding operation used a bounded `ctx` with a 2-minute timeout. If the network op took nearly the full timeout and the HTTP client disconnected, `r.Context()` may already be cancelled — the WASM load would fail silently (it's just logged as a WARN). It would be more consistent to pass `ctx` here so the load shares the same deadline as the download that preceded it:

```go
w.liveLoadPlugin(ctx, ip.Path, "")
```

web/web_marketplace.go
```diff
@@ -0,0 +1,336 @@
...
+func (w *WebServer) liveUnloadPlugin(ctx context.Context, name string) {
+	if w.wasmLoader == nil {
+		return
+	}
+	if err := w.wasmLoader.UnloadPlugin(ctx, name); err != nil {
+		log.Printf("WARN: live-unload plugin %q failed: %v", name, err)
+	}
+}
+
+func urlEncode(s string) string {
```
acmacalister, line 334: `urlEncode` only replaces spaces with `+`, so error messages containing `&`, `=`, `#`, or `%` will produce malformed query strings in the redirect URLs. For example, an error like `SHA256 mismatch: got abc&want xyz` would inject a second query param into the URL and truncate the visible error message.

`url.QueryEscape` from the standard library handles all of this correctly:

```go
func urlEncode(s string) string {
    return url.QueryEscape(s)
}
```

(You'd also need to add `"net/url"` to the import block.)

acmacalister (commented):
Solid feature — the architecture is well-structured, the test suite is genuinely comprehensive (real WASM binary, real filesystem, real runtime), and the SHA256 verification + size limits show good security hygiene. Four things worth a look before merging, all inline.

PR #135, opened 2026-04-16
feat: add Docker image publishing to release pipeline

daltoniam:
## Summary

- Add multi-arch Docker images (amd64/arm64) to the release pipeline, published to `ghcr.io/daltoniam/switchboard` via goreleaser buildx
- Minimal `scratch`-based Dockerfile with CA certs for HTTPS connectivity
- Example `docker-compose.yml` for quick local setup with env var passthrough and config volume persistence

## Changes

| File | What |
|------|------|
| `Dockerfile` | Multi-stage: Alpine for CA certs → `scratch` final with static binary |
| `.goreleaser.yml` | `dockers` (amd64 + arm64 buildx) + `docker_manifests` (versioned + `latest`) |
| `.github/workflows/release.yml` | `packages: write` perm, `setup-buildx-action`, `login-action` for GHCR |
| `docker-compose.yml` | Example compose with env vars and named config volume |

## Usage

```bash
# Env vars (simplest)
docker run -d -p 3847:3847 \
  -e GITHUB_TOKEN=ghp_xxx \
  ghcr.io/daltoniam/switchboard:latest

# With persistent config
docker run -d -p 3847:3847 \
  -v ~/.config/switchboard:/root/.config/switchboard \
  ghcr.io/daltoniam/switchboard:latest

# Docker Compose
docker compose up -d
```

## Test plan

- [ ] Verify `RELEASE_TOKEN` has `write:packages` scope for GHCR push
- [ ] Trigger a test release and confirm images appear at `ghcr.io/daltoniam/switchboard`
- [ ] Pull and run the image locally to verify the server starts and responds on `:3847`

🐘 Generated with Crush

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,6 +7,7 @@ on:
 
 permissions:
   contents: write
+  packages: write
 
 jobs:
   release:
@@ -20,6 +21,14 @@ jobs:
       - uses: actions/setup-go@v5
         with:
           go-version-file: go.mod
+
+      - uses: docker/setup-buildx-action@v3
+
+      - uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.RELEASE_TOKEN }}
 
       - uses: goreleaser/goreleaser-action@v6
         with:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM alpine:3.21 AS certs
+RUN apk add --no-cache ca-certificates
+
+FROM scratch
+COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+COPY switchboard /switchboard
+
+EXPOSE 3847
+ENTRYPOINT ["/switchboard"]
+CMD ["--port", "3847"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,7 @@ FROM alpine:3.21 AS certs
 RUN apk add --no-cache ca-certificates
 
 FROM scratch
+ENV HOME=/root
 COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
 COPY switchboard /switchboard
 
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -105,6 +105,34 @@ nfpms:
       postinstall: scripts/postinstall.sh
       preremove: scripts/preremove.sh
 
+dockers:
+  - image_templates:
+      - "ghcr.io/daltoniam/switchboard:{{ .Version }}-amd64"
+    use: buildx
+    goos: linux
+    goarch: amd64
+    dockerfile: Dockerfile
+    build_flag_templates:
+      - "--platform=linux/amd64"
+  - image_templates:
+      - "ghcr.io/daltoniam/switchboard:{{ .Version }}-arm64"
+    use: buildx
+    goos: linux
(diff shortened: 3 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM alpine:3.21 AS certs
+RUN apk add --no-cache ca-certificates
+
+FROM scratch
```
acmacalister, line 4: The `scratch` base has no `/etc/passwd` and no default `$HOME`, which causes `os.UserHomeDir()` to fail at startup — the server calls `config.NewManager()` unconditionally, which calls `os.UserHomeDir()`, and without a fallback it returns an error that causes a fatal log.

Adding `ENV HOME=/root` here fixes it and also makes the config volume path (`/root/.config/switchboard`) consistent with the `docker-compose.yml` mount:

```dockerfile
FROM scratch
ENV HOME=/root
COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
COPY switchboard /switchboard
```

Dockerfile
```diff
@@ -0,0 +1,10 @@
...
+COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+COPY switchboard /switchboard
+
+EXPOSE 3847
+ENTRYPOINT ["/switchboard"]
```
acmacalister, line 10: `--port 3847` is the server's default, so this `CMD` is a no-op. It's not wrong, but it does imply you can override it with `docker run ... --port 9000` — which actually works fine since `CMD` is replaced by anything passed after the image name. Just worth noting in case someone expects `docker run ... 9000` (positional) to work; they'd need `docker run ... --port 9000`.

acmacalister (commented):
Solid setup — multi-arch buildx with a minimal scratch image is exactly the right approach here. One thing to fix before merging: the scratch base will cause the container to crash at startup without `ENV HOME=/root`. One small note inline too.

Dockerfile
daltoniam: Good catch — fixed in 7419cbb. Added `ENV HOME=/root` so `os.UserHomeDir()` resolves correctly in the `scratch` image.

Dockerfile
daltoniam: Yeah, it's intentionally explicit so the `EXPOSE` and default behavior are self-documenting. Overriding works as expected with `docker run ... --port 9000`.

.github/workflows/release.yml
```diff
@@ -21,6 +22,14 @@ jobs:
         with:
           go-version-file: go.mod
 
+      - uses: docker/setup-buildx-action@v3
+
+      - uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.RELEASE_TOKEN }}
```
acmacalister, line 31: The workflow already declares `packages: write`, so the built-in `GITHUB_TOKEN` has everything it needs to push to GHCR. Using a PAT here is broader than necessary — PATs don't expire with the job and carry whatever other scopes were granted at creation. `RELEASE_TOKEN` is still needed for the goreleaser step (GitHub release creation), but splitting the two uses is cleaner:

```yaml
- uses: docker/login-action@v3
  with:
    registry: ghcr.io
    username: ${{ github.repository_owner }}
    password: ${{ secrets.GITHUB_TOKEN }}
```

acmacalister (approved):
Solid — builds clean, all tests pass, lint is clear. The scratch-based image with explicit `ENV HOME=/root` is the right approach. One small thing inline about the GHCR login token.

PR #136, opened 2026-04-17
Add custom field support to Jira create/update issue

PerishableDave:
## Summary
- Adds a `custom_fields` JSON parameter to `jira_create_issue` and `jira_update_issue` that accepts field values keyed by Jira field ID (e.g. `{"customfield_10001": "value"}`)
- Custom field values are merged into the fields payload sent to Jira's REST API, supporting all field types (text, select, multi-select, date, user, etc.)
- Users discover field IDs and types via the existing `jira_list_fields` tool — no client-side validation, Jira API returns clear errors for invalid fields

## Test plan
- [x] Table-driven tests for `updateIssue`: standard+custom merge, custom-only, complex values (select/multi-select objects), invalid JSON error
- [x] Table-driven tests for `createIssue`: custom fields in creation request, invalid JSON error
- [x] All existing tests pass
- [x] `go vet` clean
- [ ] Live smoke test: `jira_list_fields` → pick a custom field → `jira_update_issue` with `custom_fields`

### DIFF
--- a/integrations/jira/issues.go
+++ b/integrations/jira/issues.go
@@ -90,6 +90,15 @@ func createIssue(ctx context.Context, j *jira, args map[string]any) (*mcp.ToolRe
 	if v := r.Str("parent_key"); v != "" {
 		fields["parent"] = map[string]string{"key": v}
 	}
+	if cfStr := r.Str("custom_fields"); cfStr != "" {
+		var cf map[string]any
+		if err := json.Unmarshal([]byte(cfStr), &cf); err != nil {
+			return mcp.ErrResult(fmt.Errorf("invalid JSON for custom_fields: %w", err))
+		}
+		for k, v := range cf {
+			fields[k] = v
+		}
+	}
 	if err := r.Err(); err != nil {
 		return mcp.ErrResult(err)
 	}
@@ -127,6 +136,15 @@ func updateIssue(ctx context.Context, j *jira, args map[string]any) (*mcp.ToolRe
 			rawLabels[i] = strings.TrimSpace(l)
 		}
 		fields["labels"] = rawLabels
+	}
+	if cfStr := r.Str("custom_fields"); cfStr != "" {
+		var cf map[string]any
+		if err := json.Unmarshal([]byte(cfStr), &cf); err != nil {
+			return mcp.ErrResult(fmt.Errorf("invalid JSON for custom_fields: %w", err))
+		}
+		for k, v := range cf {
+			fields[k] = v
+		}
 	}
 	issueKey := r.Str("issue_key")
 	if err := r.Err(); err != nil {
--- a/integrations/jira/issues.go
+++ b/integrations/jira/issues.go
@@ -10,6 +10,15 @@ import (
 	mcp "github.com/daltoniam/switchboard"
 )
 
+// standardFields are the Jira field keys managed by explicit tool parameters.
+// Custom fields passed via custom_fields must not collide with these to prevent
+// overwriting already-formatted values (e.g. priority as {"name":"High"}).
+var standardFields = map[string]bool{
+	"project": true, "issuetype": true, "summary": true,
+	"description": true, "priority": true, "assignee": true,
+	"labels": true, "parent": true,
+}
+
 func searchIssues(ctx context.Context, j *jira, args map[string]any) (*mcp.ToolResult, error) {
 	r := mcp.NewArgs(args)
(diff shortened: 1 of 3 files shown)


integrations/jira/issues.go
```diff
@@ -90,6 +90,15 @@ func createIssue(ctx context.Context, j *jira, args map[string]any) (*mcp.ToolRe
 	if v := r.Str("parent_key"); v != "" {
 		fields["parent"] = map[string]string{"key": v}
 	}
+	if cfStr := r.Str("custom_fields"); cfStr != "" {
+		var cf map[string]any
+		if err := json.Unmarshal([]byte(cfStr), &cf); err != nil {
+			return mcp.ErrResult(fmt.Errorf("invalid JSON for custom_fields: %w", err))
+		}
+		for k, v := range cf {
+			fields[k] = v
```
acmacalister: The merge loop runs after standard fields are set, so a caller passing `{"priority": "High"}` in `custom_fields` would silently replace the already-formatted `{"name": "High"}` object with a plain string — causing a Jira 400 that's hard to trace back to this code.

Same applies in `updateIssue`. Worth guarding against known standard field keys so the explicit params always win:

```go
var standardFields = map[string]bool{
    "project": true, "issuetype": true, "summary": true,
    "description": true, "priority": true, "assignee": true,
    "labels": true, "parent": true,
}
for k, v := range cf {
    if !standardFields[k] {
        fields[k] = v
    }
}
```

acmacalister (commented):
Solid feature — the `map[string]any` approach is exactly right for opaque Jira fields and the test coverage is thorough. One thing worth looking at inline.

acmacalister (commented):
Nice PR — the `standardFields` guard is exactly the right fix for the override risk flagged in review, and the test coverage is thorough. Two small things inline.

integrations/jira/jira_test.go
```diff
@@ -400,6 +400,171 @@ func TestSearchIssues(t *testing.T) {
...
+			checkBody: func(t *testing.T, body map[string]any) {
+				fields := body["fields"].(map[string]any)
+				assert.Equal(t, "New issue", fields["summary"])
+				assert.Equal(t, "story points", fields["customfield_10001"])
+				selectField := fields["customfield_10002"].(map[string]any)
+				assert.Equal(t, "10100", selectField["id"])
+			},
+		},
+		{
+			name: "invalid JSON returns error",
```
acmacalister, line 529: `TestUpdateIssue_CustomFields` has a "standard field keys in custom_fields are ignored" case, but `TestCreateIssue_CustomFields` doesn't. The `standardFields` guard runs in both handlers — worth adding the same coverage here so a future refactor can't accidentally break it undetected:

```go
{
    name: "standard field keys in custom_fields are ignored",
    args: map[string]any{
        "project_key":   "PROJ",
        "issue_type":    "Task",
        "summary":       "New issue",
        "custom_fields": `{"project": "OVERRIDE", "customfield_10001": "value"}`,
    },
    checkBody: func(t *testing.T, body map[string]any) {
        fields := body["fields"].(map[string]any)
        // project from explicit param wins, not the custom_fields value
        project := fields["project"].(map[string]string)
        assert.Equal(t, "PROJ", project["key"])
        assert.Equal(t, "value", fields["customfield_10001"])
    },
},
```

integrations/jira/issues.go
```diff
@@ -90,6 +99,17 @@ func createIssue(ctx context.Context, j *jira, args map[string]any) (*mcp.ToolRe
 	if v := r.Str("parent_key"); v != "" {
 		fields["parent"] = map[string]string{"key": v}
 	}
+	if cfStr := r.Str("custom_fields"); cfStr != "" {
+		var cf map[string]any
+		if err := json.Unmarshal([]byte(cfStr), &cf); err != nil {
+			return mcp.ErrResult(fmt.Errorf("invalid JSON for custom_fields: %w", err))
+		}
+		for k, v := range cf {
+			if !standardFields[k] {
```
acmacalister, line 108: When a caller passes a standard key like `"summary"` in `custom_fields`, we silently drop it — no error, no indication it was ignored. That's probably fine since the explicit params win and the intent is clear, but it might surprise someone who fat-fingers `"priority"` instead of using the top-level param and wonders why their value didn't apply. Worth noting in the `custom_fields` parameter description that standard field keys are ignored, so the LLM surfaces that to the user.

PR #137, opened 2026-04-20
Add calver release publishing with cross-platform binaries

aleksclark:
## Summary

- Switch version scheme from semver to **calver** (`vYYYY.MM.DD[.N]`) for automatic release tagging on main push
- Add `.goreleaser.yaml` to build **linux/darwin × amd64/arm64** binaries with version injected via ldflags
- Introduce shared `version` package replacing hardcoded `"0.2.0"` across server, project_server, and remotemcp
- Display version in the **web UI sidebar footer** and log it to **stderr on startup**

## Changes

| File | What |
|------|------|
| `version/version.go` | New shared version package (ldflags-injected `Version`, `Commit`, `Date`) |
| `.goreleaser.yaml` | GoReleaser v2 config — 4 binaries (linux/darwin × amd64/arm64) |
| `.github/workflows/auto-release.yml` | Calver tagging instead of semver bump |
| `Makefile` | `build` target now passes `-ldflags` |
| `cmd/server/main.go` | Uses version pkg, logs version on startup |
| `server/server.go` | `version.String()` instead of `"0.2.0"` |
| `server/project_server.go` | `version.String()` instead of `"0.2.0"` |
| `remotemcp/remotemcp.go` | `version.String()` instead of `"0.2.0"` |
| `web/templates/layouts/base.templ` | Version in sidebar footer |

## Test plan

- [x] `./dist/switchboard --version` prints injected version
- [ ] CI passes on this PR

💖 Generated with Crush


version/version.go
```diff
@@ -0,0 +1,15 @@
+package version
+
+var (
+	Version = "dev"
```
acmacalister, line 4: These vars being exported means any package can mutate them at runtime — `version.Version = "evil"` compiles and runs fine. The ldflags injection mechanism works equally well on unexported identifiers, and `String()`/`Full()` are already the right public API.

```go
var (
    version = "dev"
    commit  = "none"
    date    = "unknown"
)

func String() string { return version }
func Full() string   { return version + " (commit: " + commit + ", built: " + date + ")" }
```

The ldflags paths just change to `-X github.com/daltoniam/switchboard/version.version=...` etc.

web/templates/layouts/base.templ
```diff
@@ -1,12 +1,17 @@
...
+
+	"github.com/daltoniam/switchboard/version"
+)
 
 type PageData struct {
 	Title         string
 	CurrentPath   string
 	FlashSuccess  string
 	FlashError    string
+	Version       string
```
acmacalister, line 14: `PageData.Version` is never populated — `pageData()` in `web/web.go` doesn't set it, and the template calls `version.String()` directly instead of `data.Version`. The field is dead weight right now.

Either remove the field and keep the direct `version.String()` call in the template (simpler), or populate it in `pageData()` and use `data.Version` in the template (makes the template testable). As-is it's confusing to future readers.

.github/workflows/auto-release.yml
```diff
@@ -45,73 +42,49 @@ jobs:
...
-              ;;
-            minor)
-              MINOR=$((MINOR + 1))
-              PATCH=0
-              ;;
-            patch)
-              PATCH=$((PATCH + 1))
-              ;;
-          esac
+          EXISTING=$(git tag -l "v${TODAY}*" --sort=-v:refname | head -1)
```
acmacalister, line 66: `--sort=-v:refname` sorts the `.N` suffix lexicographically, so on a day where you tag `v2026.04.20.9` and then `v2026.04.20.10`, `git tag -l ... --sort=-v:refname | head -1` returns `v2026.04.20.9` (since `'1' < '9'`). The next tag would then become `v2026.04.20.10` again and `git push` would fail on the duplicate.

Swapping to `--sort=-version:refname` or piping through `sort -V | tail -1` makes it handle double-digit N correctly. Unlikely to matter in practice, but worth a one-liner fix.

version/version.go
```diff
@@ -0,0 +1,15 @@
...
+	Commit  = "none"
+	Date    = "unknown"
+)
+
+func String() string {
```
acmacalister, line 9: No tests for this package. It's tiny, but we require tests for all new code. A couple of lines covering the defaults would do it:

```go
func TestString_default(t *testing.T) {
    require.Equal(t, "dev", String())
}

func TestFull_default(t *testing.T) {
    require.Contains(t, Full(), "dev")
    require.Contains(t, Full(), "none")
}
```

acmacalister (commented):
Nice work — clean extraction of the version package, solid goreleaser config, and the calver workflow is much simpler than the old semver bump. A few things inline worth a look.

PR #138, opened 2026-04-22
feat: add field compaction support to WASM module ABI

daltoniam:
## Summary

- Adds optional `compact_specs()` export to the WASM plugin ABI, enabling WASM modules to declare per-tool field compaction specs using the same dot-notation syntax as native integrations
- Specs are parsed once at module load via `ParseCompactSpecs` and cached for the module lifetime — `Module` now implements `FieldCompactionIntegration` so the server pipeline picks it up automatically
- Fully backward compatible: modules without the export simply skip compaction (no ABI version bump needed)

## Changes

**Go host** (`wasm/runtime.go`, `wasm/module.go`):
- `Module` gains `fnCompactSpecs` (optional function ref) and `compactSpecs` (parsed cache)
- `loadCompactSpecs()` called once after `LoadModule` resolves exports
- `CompactSpec(toolName)` implements `mcp.FieldCompactionIntegration`
- Compile-time interface assertion

**Rust SDK** (`wasm/guest-rust/sdk/src/lib.rs`):
- `leaked_compact_specs(&HashMap<String, Vec<String>>)` — serializes specs map and leaks to WASM memory

**Example plugin** (`wasm/guest-rust/example/src/lib.rs`):
- Exports `compact_specs` with `example_list_items → ["id", "name"]`
- Rebuilt `example.wasm`

## Test plan

- [x] `TestFieldCompactionInterface` — compile-time interface check
- [x] `TestCompactSpec_HasSpecs` — specs present for declared tools
- [x] `TestCompactSpec_NoSpecs` — returns false for undeclared tools
- [x] `TestCompactSpec_AppliesCorrectly` — end-to-end compaction strips extra fields
- [x] `TestCompactSpec_AllToolsCovered` — at least one tool has specs
- [x] All existing WASM tests pass (backward compatible)
- [x] Full test suite passes (`go test ./...`)

💘 Generated with Crush


wasm/runtime.go
```diff
@@ -118,9 +125,67 @@ func (m *Module) resolveExports() error {
...
 }
 
+// loadCompactSpecs calls the guest compact_specs() export (if present),
+// parses the returned JSON map into pre-compiled CompactField slices,
+// and caches them for the lifetime of the module.
+func (m *Module) loadCompactSpecs() {
+	if m.fnCompactSpecs == nil {
+		return
+	}
+	ctx := context.Background()
```
acmacalister: `loadCompactSpecs` creates its own `context.Background()` even though `LoadModule` receives a caller context that may carry a deadline or cancellation — and passes it to every other call in the same function. A slow or hung WASM module can't be interrupted here.

Threading the context through is a one-line fix:

```go
func (m *Module) loadCompactSpecs(ctx context.Context) { ... }
// call site:
m.loadCompactSpecs(ctx)
```

wasm/wasm_test.go
```diff
@@ -303,3 +303,80 @@ func TestIntegrationInterface(t *testing.T) {
...
+	}
+	if _, exists := first["extra"]; exists {
+		t.Error("extra field should have been compacted out")
+	}
+	if _, exists := first["price"]; exists {
+		t.Error("price field should have been compacted out")
+	}
+}
+
+func TestCompactSpec_AllToolsCovered(t *testing.T) {
```
acmacalister: The test name `TestCompactSpec_AllToolsCovered` implies every tool must have a compaction spec, but the assertion only checks that at least one does. As this plugin grows with new tools that intentionally have no specs, the name will confuse future readers (and maybe mislead them into thinking the assertion is weaker than it should be).

Something like `TestCompactSpec_AtLeastOneTool` or a clearer failure message like `"expected at least one tool to have compact specs (not all tools need specs)"` would make the intent unambiguous.

acmacalister (approved):
Solid PR — clean implementation, the parse-once caching and optional-export ABI design are both right. Two small things inline worth a look before merge.

acmacalister (approved):
Clean implementation — build passes, full test suite green, lint clean. The backward-compatible optional export design is exactly right, and parsing/caching specs once at load time rather than per-call is the correct approach. The open context-threading comment from the previous review is still worth addressing, but nothing here is a blocker.

PR #140, opened 2026-04-22
feat: add dry-run mode for execute (Phase 3)

acmacalister:
## Summary

- **`dry_run` flag on execute**: Pass `dry_run: true` to validate arguments and check integration health without executing. LLMs can preview mutation operations (create issue, merge PR, post comment) before committing.
- **Tiered approach**: Integrations implementing the new `DryRunIntegration` interface return native previews. All others get a simulated dry-run that validates args, checks the circuit breaker, and confirms health.
- **No side effects**: Dry-run never calls `Execute`, never pins results, never records breadcrumbs — purely a validation + preview path.

This is Phase 3 of the [stateful features plan](docs/stateful-features.md). Independent of Phases 1-2 but most useful alongside session context.

## Test plan

- [x] 8 tests: simulated fallback, validation failure, native integration, native-declines-falls-back, unhealthy integration, tool not found, does-not-execute, no-pinning

🐘 Generated with Crush

### DIFF
--- a/mcp.go
+++ b/mcp.go
@@ -267,6 +267,15 @@ func (tn *ToolName) UnmarshalJSON(b []byte) error {
 // Alias lets adapter authors use mcp.Markdown without importing the markdown subpackage.
 type Markdown = markdown.Markdown
 
+// DryRunIntegration is an optional interface that integrations can implement
+// to provide native dry-run previews for mutation tools. When dry_run is set
+// on an execute call, the server checks this interface first. If the integration
+// returns (result, true), that result is used directly. Otherwise the server
+// falls back to a simulated dry-run (arg validation + preview).
+type DryRunIntegration interface {
+	DryRun(ctx context.Context, toolName ToolName, args map[string]any) (*ToolResult, bool)
+}
+
 // MarkdownIntegration is an optional interface that integrations can implement
 // to render tool responses as Markdown instead of JSON. The server calls
 // RenderMarkdown in processResult before compaction — if it returns rendered
--- a/server/dryrun.go
+++ b/server/dryrun.go
@@ -1,0 +1,65 @@
+package server
+
+import (
+	"context"
+	"fmt"
+
+	mcp "github.com/daltoniam/switchboard"
+	mcpsdk "github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+func (s *Server) handleDryRun(ctx context.Context, toolName mcp.ToolName, args map[string]any) (*mcpsdk.CallToolResult, error) {
+	integration, toolDef, err := s.findTool(toolName)
+	if err != nil {
+		return errorResult(err.Error()), nil
+	}
+
+	if err := validateArgs(toolDef, args); err != nil {
+		return errorResult(fmt.Sprintf("dry-run validation failed: %s", err)), nil
+	}
+
+	cb := s.getBreaker(integration.Name())
+	if !cb.allow() {
+		cb.recordSuccess()
+		return errorResult(fmt.Sprintf(
+			"dry-run: integration %q temporarily unavailable (circuit breaker open)",
+			integration.Name(),
+		)), nil
+	}
+	cb.recordSuccess()
+
+	if !integration.Healthy(ctx) {
+		return errorResult(fmt.Sprintf(
+			"dry-run: integration %q is unhealthy — call would likely fail",
+			integration.Name(),
+		)), nil
+	}
+
+	if dri, ok := integration.(mcp.DryRunIntegration); ok {
+		if result, handled := dri.DryRun(ctx, toolName, args); handled {
+			return &mcpsdk.CallToolResult{
+				Content: []mcpsdk.Content{
+					&mcpsdk.TextContent{Text: result.Data},
+				},
+				IsError: result.IsError,
+			}, nil
+		}
+	}
+
+	result, err := mcp.JSONResult(map[string]any{
+		"dry_run":        true,
+		"tool":           toolName,
+		"integration":    integration.Name(),
+		"validated_args": args,
+		"status":         "ok",
+		"note":           "Simulated dry-run: arguments are valid, integration is healthy. This tool does not support native dry-run preview.",
+	})
+	if err != nil {
+		return errorResult(err.Error()), nil
+	}
+	return &mcpsdk.CallToolResult{
+		Content: []mcpsdk.Content{
+			&mcpsdk.TextContent{Text: result.Data},
+		},
+	}, nil
+}
--- a/server/dryrun_test.go
+++ b/server/dryrun_test.go
@@ -1,0 +1,234 @@
+package server
+
+import (
+	"context"
+	"encoding/json"
+	"testing"
+
+	mcp "github.com/daltoniam/switchboard"
+	mcpsdk "github.com/modelcontextprotocol/go-sdk/mcp"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type mockDryRunIntegration struct {
+	mockIntegration
+	dryRunFn func(ctx context.Context, toolName mcp.ToolName, args map[string]any) (*mcp.ToolResult, bool)
+}
+
+func (m *mockDryRunIntegration) DryRun(ctx context.Context, toolName mcp.ToolName, args map[string]any) (*mcp.ToolResult, bool) {
+	if m.dryRunFn != nil {
+		return m.dryRunFn(ctx, toolName, args)
+	}
+	return nil, false
+}
+
+func dryRunRequest(toolName string, args map[string]any) *mcpsdk.CallToolRequest {
+	data, _ := json.Marshal(map[string]any{
+		"tool_name": toolName,
+		"arguments": args,
+		"dry_run":   true,
+	})
+	return &mcpsdk.CallToolRequest{
+		Params: &mcpsdk.CallToolParamsRaw{
+			Name:      "execute",
+			Arguments: json.RawMessage(data),
+		},
+	}
+}
+
+func TestDryRun_SimulatedFallback(t *testing.T) {
+	mi := &mockIntegration{
+		name:    "github",
+		healthy: true,
+		tools: []mcp.ToolDefinition{
+			{
+				Name:        "github_create_issue",
+				Description: "Create issue",
+				Parameters:  map[string]string{"owner": "Owner", "repo": "Repo", "title": "Title"},
+				Required:    []string{"owner", "repo", "title"},
+			},
+		},
+	}
+	s := setupTestServer(mi)
+	ctx := context.Background()
+
+	result, err := s.handleExecute(ctx, dryRunRequest("github_create_issue", map[string]any{
+		"owner": "daltoniam",
+		"repo":  "switchboard",
+		"title": "Test issue",
+	}))
+	require.NoError(t, err)
+	require.False(t, result.IsError)
+
+	var resp map[string]any
+	require.NoError(t, json.Unmarshal([]byte(result.Content[0].(*mcpsdk.TextContent).Text), &resp))
+	assert.Equal(t, true, resp["dry_run"])
+	assert.Equal(t, "github_create_issue", resp["tool"])
+	assert.Equal(t, "github", resp["integration"])
+	assert.Equal(t, "ok", resp["status"])
+	args := resp["validated_args"].(map[string]any)
+	assert.Equal(t, "daltoniam", args["owner"])
+}
+
+func TestDryRun_ValidationFails(t *testing.T) {
+	mi := &mockIntegration{
+		name:    "github",
+		healthy: true,
+		tools: []mcp.ToolDefinition{
+			{
+				Name:       "github_create_issue",
+				Parameters: map[string]string{"owner": "Owner", "repo": "Repo", "title": "Title"},
+				Required:   []string{"owner", "repo", "title"},
+			},
+		},
+	}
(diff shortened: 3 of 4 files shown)


server/dryrun.go
```diff
@@ -0,0 +1,65 @@
...
+		return errorResult(err.Error()), nil
+	}
+
+	if err := validateArgs(toolDef, args); err != nil {
+		return errorResult(fmt.Sprintf("dry-run validation failed: %s", err)), nil
+	}
+
+	cb := s.getBreaker(integration.Name())
+	if !cb.allow() {
+		cb.recordSuccess()
```
acmacalister: `cb.recordSuccess()` is called here when `cb.allow()` returned `false` — meaning the circuit is open and we're about to reject the request. Calling `recordSuccess()` at this point resets `failures` to zero, healing the breaker mid-rejection. The next call will pass through as if nothing happened.

Should just return without touching the breaker state:

```go
if !cb.allow() {
    return errorResult(fmt.Sprintf(
        "dry-run: integration %q temporarily unavailable (circuit breaker open)",
        integration.Name(),
    )), nil
}
cb.recordSuccess()
```

The `recordSuccess()` that follows the `if` block (line 29) is the right call — that one fires when the breaker is in half-open state and we're about to let the request continue.

server/dryrun.go
```diff
@@ -0,0 +1,65 @@
...
+	mcpsdk "github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+func (s *Server) handleDryRun(ctx context.Context, toolName mcp.ToolName, args map[string]any) (*mcpsdk.CallToolResult, error) {
+	integration, toolDef, err := s.findTool(toolName)
+	if err != nil {
+		return errorResult(err.Error()), nil
+	}
+
+	if err := validateArgs(toolDef, args); err != nil {
```
acmacalister, line 17: Dry-run validation runs against the raw `args` from the request, but the normal execute path first calls `resolveRefs(sess, args.Arguments)` and `sess.MergeDefaults(args.Arguments)` before reaching validation. That means a user who has set session defaults (e.g. `owner`/`repo`) or is passing a pinned reference like `$1` will get a "missing required parameter" error from dry-run even though the real execute would succeed.

If this is intentional (dry-run validates only explicit args), worth a comment. If not, `handleDryRun` needs access to the session to apply the same transformations the real path does — probably by passing the resolved args in from `handleExecute` rather than re-resolving inside `handleDryRun`.

server/dryrun_test.go
```diff
@@ -0,0 +1,234 @@
...
+	ctx := context.Background()
+
+	_, err := s.handleExecute(ctx, dryRunRequest("github_create_issue", map[string]any{
+		"title": "test",
+	}))
+	require.NoError(t, err)
+
+	sess := s.sessionStore.GetOrCreate("default")
+	assert.Equal(t, 0, sess.PinnedCount(), "dry-run should not pin results")
+}
```
acmacalister, line 234: The circuit-breaker-open path (`!cb.allow()` branch) has a bug (see comment on `dryrun.go:23`) but no test covering it — which is partly why it slipped through. Adding a `TestDryRun_CircuitBreakerOpen` that forces the breaker into the open state and asserts the correct error message (and that the breaker's failure count is *not* reset) would catch this and prevent regression.

Similarly, given the session-resolution gap noted above, a test that sets session defaults before calling dry-run would clarify the intended behavior.

acmacalister (commented):
Good shape overall — clean architecture, `DryRunIntegration` as an optional interface is the right extension point, and the test coverage is thorough for most paths. Two things need attention before merge: a logic inversion in the circuit-breaker path, and a question about whether dry-run should respect session context the way real execute does.

server/dryrun.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	cb := s.getBreaker(integration.Name())
+	if !cb.allow() {
+		return errorResult(fmt.Sprintf(
+			"dry-run: integration %q temporarily unavailable (circuit breaker open)",
+			integration.Name(),
+		)), nil
+	}
+	cb.recordSuccess()
```
acmacalister, line 29: `cb.recordSuccess()` fires here whenever `allow()` returns `true` — including when the breaker is in half-open state (cooldown elapsed, `failures >= threshold`). In that case we call `recordSuccess()` before we've verified the integration is healthy, so if `Healthy()` below returns `false`, the breaker is already reset to `failures = 0` and the next real execute will pass through as though the integration recovered.

The real `executeTool` path only calls `recordSuccess()` after a successful `Execute()`. We should mirror that — call it after the `Healthy()` check passes, and call `recordFailure()` when `Healthy()` returns `false` so a bad probe keeps the breaker open:

```go
cb := s.getBreaker(integration.Name())
if !cb.allow() {
    return errorResult(fmt.Sprintf(
        "dry-run: integration %q temporarily unavailable (circuit breaker open)",
        integration.Name(),
    )), nil
}

if !integration.Healthy(ctx) {
    cb.recordFailure() // failed probe — keep breaker open
    return errorResult(fmt.Sprintf(
        "dry-run: integration %q is unhealthy — call would likely fail",
        integration.Name(),
    )), nil
}
cb.recordSuccess() // health confirmed — close the breaker
```

`TestDryRun_CircuitBreakerOpen` covers the open-state rejection but not the half-open probe. A `TestDryRun_HalfOpenProbeFailure` that manipulates `lastFailure` to simulate a half-open state and then asserts the breaker is still open would lock this down.

server/server.go
```diff
@@ -671,6 +676,10 @@ func (s *Server) handleExecute(ctx context.Context, req *mcpsdk.CallToolRequest)
 	resolveRefs(sess, args.Arguments)
 	args.Arguments = sess.MergeDefaults(args.Arguments)
 
+	if args.DryRun {
```
acmacalister: If a caller passes both `dry_run: true` and `script`, the script runs normally — the `dry_run` flag is silently ignored because the `args.Script != ""` check above returns early before we reach this gate. The tool description says "Works with tool_name only", but a silent no-op is easy to miss. Worth returning an explicit error here:

```go
if args.DryRun && args.Script != "" {
    return errorResult("dry_run is not supported with script — use tool_name"), nil
}
```

acmacalister (commented):
Good shape overall — the tiered DryRunIntegration interface is clean and the fix commit resolved the two issues from the previous round. One remaining circuit-breaker edge case and a minor silent-ignore to look at inline.

server/dryrun_test.go
```diff
@@ -0,0 +1,357 @@
...
+	require.NoError(t, err)
+	require.False(t, result.IsError)
+
+	var resp map[string]any
+	require.NoError(t, json.Unmarshal([]byte(result.Content[0].(*mcpsdk.TextContent).Text), &resp))
+	assert.Equal(t, true, resp["dry_run"])
+	assert.Equal(t, "github_create_issue", resp["tool"])
+	assert.Equal(t, "github", resp["integration"])
+	assert.Equal(t, "ok", resp["status"])
+	args := resp["validated_args"].(map[string]any)
```
acmacalister: Direct type assertion on `resp["validated_args"]` will panic (not fail) if the field is absent or wrong type — making a future regression harder to diagnose. The same pattern appears again at line 353.

```go
validatedArgs, ok := resp["validated_args"].(map[string]any)
require.True(t, ok, "expected validated_args to be a map")
assert.Equal(t, "daltoniam", validatedArgs["owner"])
```

mcp.go
```diff
@@ -267,6 +267,15 @@ func (tn *ToolName) UnmarshalJSON(b []byte) error {
 // Alias lets adapter authors use mcp.Markdown without importing the markdown subpackage.
 type Markdown = markdown.Markdown
 
+// DryRunIntegration is an optional interface that integrations can implement
+// to provide native dry-run previews for mutation tools. When dry_run is set
+// on an execute call, the server checks this interface first. If the integration
+// returns (result, true), that result is used directly. Otherwise the server
+// falls back to a simulated dry-run (arg validation + preview).
+type DryRunIntegration interface {
+	DryRun(ctx context.Context, toolName ToolName, args map[string]any) (*ToolResult, bool)
```
acmacalister, line 276: `MarkdownIntegration.RenderMarkdown` has a method-level doc explaining its return contract; `DryRun` doesn't. Worth adding a note that `(nil, false)` is the expected "decline" signal (not an error), so adapter authors don't reach for `(*ToolResult{IsError: true}, true)` when they just want to fall back:

```go
// DryRun returns a native preview for the given tool call, or (nil, false)
// if this tool doesn't support native dry-run and the server should fall back
// to a simulated preview. Returning (nil, false) is not an error.
DryRun(ctx context.Context, toolName ToolName, args map[string]any) (*ToolResult, bool)
```

acmacalister (commented):
Excellent iteration — all five previous round comments are addressed and the new tests lock down the edge cases well. Two small nits inline, neither blocking.

server/server.go
```diff
@@ -643,11 +647,15 @@ func (s *Server) handleExecute(ctx context.Context, req *mcpsdk.CallToolRequest)
...
 		Arguments map[string]any `json:"arguments"`
 		Script    string         `json:"script"`
+		DryRun    bool           `json:"dry_run"`
 	}
 	if err := json.Unmarshal(req.Params.Arguments, &args); err != nil {
 		return errorResult("invalid arguments: " + err.Error()), nil
 	}
 
+	if args.DryRun && args.Script != "" {
+		return errorResult("dry_run is not supported with script — use tool_name"), nil
```
daltoniam: not being about to use script is a bit of bummer here, understandable why, but does suck for exercising maximum token use reduction

daltoniam (approved):

acmacalister (commented):
Solid iteration — previous round of comments are all addressed. Found two more small things inline.

server/dryrun.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	if !integration.Healthy(ctx) {
+		cb.recordFailure()
+		return errorResult(fmt.Sprintf(
+			"dry-run: integration %q is unhealthy — call would likely fail",
+			integration.Name(),
+		)), nil
+	}
+	cb.recordSuccess()
```
acmacalister: `recordSuccess()` here heals the circuit breaker based on `Healthy()` returning true — but `Healthy()` is typically a lightweight credential/ping check, not a full API round-trip. In `executeTool`, `recordSuccess()` only fires after a real `Execute()` call succeeds. The mismatch means a dry-run health ping can close a half-open breaker and let the next real call through against an endpoint that's still broken.

One option: skip `recordSuccess()` in the dry-run path and let the real execute do the healing:

```go
if !integration.Healthy(ctx) {
    cb.recordFailure()
    return errorResult(fmt.Sprintf(
        "dry-run: integration %q is unhealthy — call would likely fail",
        integration.Name(),
    )), nil
}
// Don't call cb.recordSuccess() here — only real Execute calls should heal the breaker.
```

If healing on a health check is intentional ("dry-run acts as a probe"), worth a comment saying so.

server/dryrun_test.go
```diff
@@ -0,0 +1,359 @@
...
+				{
+					Name:       "aws_lambda_invoke",
+					Parameters: map[string]string{"function_name": "Function"},
+					Required:   []string{"function_name"},
+				},
+			},
+		},
+		dryRunFn: func(_ context.Context, _ mcp.ToolName, args map[string]any) (*mcp.ToolResult, bool) {
+			return &mcp.ToolResult{
+				Data: `{"dry_run":true,"native":true,"would_invoke":"` + args["function_name"].(string) + `"}`,
```
acmacalister, line 113: Unsafe type assertion here — if `function_name` is absent from `args` or the wrong type, this panics rather than failing the test cleanly. The `validated_args` assertions were updated to use the two-value form, worth doing the same here:

```go
fn, ok := args["function_name"].(string)
require.True(t, ok, "expected function_name to be a string")
return &mcp.ToolResult{
    Data: `{"dry_run":true,"native":true,"would_invoke":"` + fn + `"}`,
}, true
```

acmacalister (commented):
Build, tests, and lint all pass — the dry-run design is solid. Two things from earlier review rounds look like they were lost in the squash; flagging them inline.

server/server.go
```diff
@@ -664,6 +669,10 @@ func (s *Server) handleExecute(ctx context.Context, req *mcpsdk.CallToolRequest)
 		args.Arguments = map[string]any{}
 	}
 
+	if args.DryRun {
+		return s.handleDryRun(ctx, args.ToolName, args.Arguments)
```
acmacalister, line 673: The `args.Script` check above this returns early before `dry_run` is ever evaluated — so `{ script: "api.call('github_delete_repo', ...)", dry_run: true }` executes the script for real. The tool description says "Works with tool_name only (not scripts)" but there's no enforcement.

This guard was in an intermediate commit but appears to have been dropped in the squash:

```go
if args.DryRun && args.Script != "" {
    return errorResult("dry_run is not supported with script — use tool_name"), nil
}
```

server/dryrun.go
acmacalister, line 23: This `cb.recordSuccess()` fires on the rejection path — `allow()` returned `false`, meaning the breaker is open. Resetting `failures` to zero here heals the breaker mid-rejection, so the next call passes through as if nothing happened.

This was flagged in the first review round but looks like it was lost in the squash. The fix is to remove this call and only keep the one at line 29:

```go
if !cb.allow() {
    return errorResult(fmt.Sprintf(
        "dry-run: integration %q temporarily unavailable (circuit breaker open)",
        integration.Name(),
    )), nil
}
cb.recordSuccess()
```

PR #142, opened 2026-04-23
fix: pganalyze_get_issues returns empty due to invalid state filter

aleksclark:
## Summary

- `pganalyze_get_issues` was passing `state: "open"` as a GraphQL variable, but pganalyze's API uses `"triggered"` for active issues — not `"open"`. The API silently returned empty results.
- Removed the undocumented server-side `$state` variable from the GraphQL query and switched to client-side filtering: exclude issues with `state: "resolved"` when `include_resolved` is false.
- Unified the severity + resolved filtering into a single pass over results.

## Test plan

- [x] Added `issues_test.go` with 8 tests covering: default excludes resolved, include_resolved returns all, severity filter, case-insensitive severity, combined severity+resolved, no state variable sent to API, org slug passthrough, empty response
- [x] All 37 pganalyze integration tests pass

🐘 Generated with Crush

### DIFF
--- a/integrations/pganalyze/issues.go
+++ b/integrations/pganalyze/issues.go
@@ -10,8 +10,8 @@ import (
 
 func getIssues(ctx context.Context, p *pganalyze, args map[string]any) (*mcp.ToolResult, error) {
 	query := `
-		query GetIssues($organizationSlug: ID, $serverId: ID, $databaseId: ID, $state: String) {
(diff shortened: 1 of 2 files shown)


acmacalister (approved):
Clean PR — the root-cause diagnosis is spot-on and the fix is surgical. Build passes, all 37 pganalyze tests pass (including the new 8), lint is clean. The `TestGetIssues_NoStateVariableSentToAPI` test is a nice touch that directly proves the regression is gone. LGTM.

PR #143, opened 2026-04-23
Fix release builds: semver-compatible calver tags and goreleaser config

daltoniam:
## Summary

- **Fixed calver tag format** — tags like `v2026.04.22.3` are invalid semver (leading zeros + 4 components). Switched to `vYYYY.MMDD.N` (e.g. `v2026.422.0`) which is always valid 3-part semver.
- **Deleted legacy `.goreleaser.yaml`** — GoReleaser resolves `.yaml` before `.yml`, so the minimal legacy config was silently overriding the full release config with Homebrew, Docker, nfpms, scoops, etc.
- **Fixed ldflags in `.goreleaser.yml`** — was injecting into nonexistent `main.version`/`main.commit`/`main.date`; now correctly targets `version.Version`/`.Commit`/`.Date`.

## Test plan

- [x] `go build` passes
- [x] `go vet` passes
- [x] Server and core tests pass
- [ ] Verify next merge to main produces a valid semver tag (e.g. `v2026.422.0`)
- [ ] Verify GoReleaser completes successfully with the tag

🐘 Generated with Crush


acmacalister (approved):
Clean PR — builds, tests pass, lint clean. The calver arithmetic is correct (no leading zeros possible since MONTH*100+DAY >= 101), glob pattern is safe, and the deleted .goreleaser.yaml correctly removes the config that was silently shadowing the full release config. LGTM.

PR #144, opened 2026-04-24
Add SigNoz observability platform integration

aleksclark:
## Summary

- Add SigNoz integration with 27 tools covering the full API surface: APM services, distributed traces, log search, metrics queries, dashboards CRUD, alert rules CRUD, saved explorer views, and notification channels
- Uses SigNoz v4 query_range builder API with `SIGNOZ-API-KEY` header auth, configurable `skip_verify` for self-signed TLS certs
- Includes field compaction specs for all 15 read tools, dispatch parity tests, HTTP helper tests, and handler tests (29 total)

## Test plan

- [x] `make ci` passes (build, test-race, vet, lint, gosec, govulncheck)
- [x] All 29 unit tests pass including dispatch parity, compaction spec parity, and HTTP handler tests
- [x] Live-tested all tools against a real SigNoz v0.120.0 instance

🐘 Generated with Crush


integrations/signoz/signoz.go
```diff
@@ -0,0 +1,193 @@
...
+		req.Header.Set("Content-Type", "application/json")
+	}
+
+	resp, err := s.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer func() { _ = resp.Body.Close() }()
+
+	data, err := io.ReadAll(resp.Body)
```
acmacalister: This is an unbounded read — every other adapter that makes external HTTP calls (cloudflare, notion, snowflake, readarr) caps with `io.LimitReader`. SigNoz responses can include full trace event dumps, and without a cap a misbehaving server could cause OOM. Simple fix:

```go

data, err := io.ReadAll(io.LimitReader(resp.Body, maxResponseSize))
```

integrations/signoz/signoz_test.go
```diff
@@ -0,0 +1,390 @@
...
+	toolNames := make(map[mcp.ToolName]bool)
+	for name := range dispatch {
+		toolNames[name] = true
+	}
+	for name := range rawFieldCompactionSpecs {
+		assert.True(t, toolNames[name], "compaction spec %s has no dispatch handler", name)
+	}
+}
+
+func TestFieldCompactionSpecs_AllReadToolsCovered(t *testing.T) {
```
acmacalister, line 369: The `readTools` set is hardcoded, so it can only go stale — if someone adds a new read tool later and forgets the compaction spec, this test won't catch it. `TestFieldCompactionSpecs_NoOrphanSpecs` does this correctly by iterating `dispatch` dynamically.

We could drive coverage from `i.Tools()` instead, e.g. treat any tool matching `signoz_list_*`, `signoz_get_*`, `signoz_search_*`, `signoz_query_*`, or `signoz_top_*` as a read tool that requires a spec.

integrations/signoz/services.go
```diff
@@ -0,0 +1,180 @@
...
+		"service": service,
+	})
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	return mcp.RawResult(data)
+}
+
+// buildBuilderQuery constructs the v4 query_range composite query payload.
+func buildBuilderQuery(_, panelType string, query map[string]any, start, end int64, step int) map[string]any {
```
acmacalister: The first parameter (`dataSource`) is blanked out and never used. Every call site passes `"logs"`, `"traces"`, or `"metrics"` but they go nowhere. Either drop the param and update the callers, or actually wire it into the composite query — a dead parameter is confusing to anyone reading the call sites.

integrations/signoz/services.go
```diff
@@ -0,0 +1,180 @@
...
+	if service != "" {
+		items = append(items, map[string]any{
+			"key":   map[string]any{"key": "serviceName", "dataType": "string", "type": "tag", "isColumn": true},
+			"op":    "=",
+			"value": service,
+		})
+	}
+
+	if filter != "" {
+		parts := strings.SplitN(filter, " ", 3)
```
acmacalister, line 170: If the filter string has fewer than 3 space-delimited tokens (e.g. `"severity_text"` or `"key value"`), the filter is silently dropped and the query runs unfiltered — the user gets back unexpected results with no indication of why. Worth either returning an error for malformed filters or documenting the expected format more explicitly in the tool description.

acmacalister (commented):
Solid integration — builds clean, all 31 tests pass with the race detector, lint is quiet. A few things worth looking at inline; the most material is the unbounded response read in `doRequest`.

integrations/signoz/services.go
```diff
@@ -0,0 +1,183 @@
...
+			"op":    op,
+			"value": value,
+		})
+	}
+
+	return map[string]any{"op": "AND", "items": items}, nil
+}
+
+func strMs(s string) int64 {
+	v, _ := strconv.ParseInt(s, 10, 64)
```
acmacalister: `strMs` silently returns 0 when parsing fails, so a value like `"now"` or `"2024-01-01T00:00:00Z"` for `start`/`end` becomes Unix epoch 0 with no error surfaced to the caller. The query runs and succeeds against garbage time bounds.

Since start/end are already extracted via `r.Str` (which validates they're present), we could validate they're actually integers there:

```go
func strMs(s string) (int64, error) {
    v, err := strconv.ParseInt(s, 10, 64)
    if err != nil {
        return 0, fmt.Errorf("invalid epoch milliseconds %q: %w", s, err)
    }
    return v, nil
}
```

Or at minimum document the silent-zero behavior so it's not surprising to the next person who touches this. Other adapters that accept timestamps (Datadog, PostHog) use string formats the API accepts directly, so this strconv path is unique to SigNoz.

integrations/signoz/dashboards.go
```diff
@@ -0,0 +1,110 @@
...
+	return mcp.RawResult(data)
+}
+
+func getDashboard(ctx context.Context, s *signoz, args map[string]any) (*mcp.ToolResult, error) {
+	r := mcp.NewArgs(args)
+	id := r.Str("id")
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	data, err := s.get(ctx, "/api/v1/dashboards/%s", id)
```
acmacalister: User-supplied `id` goes directly into the path via `fmt.Sprintf`. If someone passes `"abc/../../v1/version"`, the constructed URL becomes `/api/v1/version` — pointing at a different endpoint entirely.

Salesforce and Confluence both use `url.PathEscape` on user-supplied path segments; worth being consistent here:

```go
data, err := s.get(ctx, "/api/v1/dashboards/%s", url.PathEscape(id))
```

Same applies to `getDashboard`, `updateDashboard`, `deleteDashboard`, `getAlert`, `updateAlert`, `deleteAlert`, `getSavedView`, `updateSavedView`, `deleteSavedView`, and `getTrace` — anywhere a user-supplied ID goes into `fmt.Sprintf` to build a path.

acmacalister (commented):
Solid integration — clean architecture, good test coverage, and the previous round of feedback has all been addressed (LimitReader, dead param in buildBuilderQuery, parseFilterItems error return). Two things worth looking at inline.

integrations/signoz/signoz_test.go
```diff
@@ -0,0 +1,424 @@
...
+
+func TestSearchLogs_MalformedFilter(t *testing.T) {
+	s := &signoz{apiKey: "key", baseURL: "http://localhost", client: &http.Client{}}
+	result, err := s.Execute(context.Background(), "signoz_search_logs", map[string]any{
+		"start": "1700000000000", "end": "1700003600000", "filter": "bad",
+	})
+	require.NoError(t, err)
+	assert.True(t, result.IsError)
+	assert.Contains(t, result.Data, "invalid filter format")
+}
```
acmacalister, line 424: About 18 out of 27 handlers have no dedicated test — things like `searchTraces`, `getTrace`, `queryMetrics`, `topOperations`, `topLevelOperations`, `entryPointOperations`, `getServiceOverview`, `getDashboard`, `createDashboard`, `updateDashboard`, and all the alert/saved-view/channel variants. The dispatch parity tests confirm the routing wires up correctly, but they don't verify that each handler calls the right endpoint, sends the right method, or propagates args the way it's supposed to.

A short `httptest.NewServer` stub per handler is usually a few lines. The existing `TestListDashboards` / `TestListAlerts` / `TestDeleteDashboard` pattern is perfect to follow. The most important ones to add are probably the write paths (`createDashboard`, `updateDashboard`, `createAlert`, `updateAlert`) since those have the JSON-passthrough branching logic that's easy to get wrong silently.

integrations/signoz/queries.go
```diff
@@ -0,0 +1,183 @@
...
+func queryMetrics(ctx context.Context, s *signoz, args map[string]any) (*mcp.ToolResult, error) {
+	r := mcp.NewArgs(args)
+	start := r.Str("start")
+	end := r.Str("end")
+	metricName := r.Str("metric_name")
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+
+	aggregateOp, _ := mcp.ArgStr(args, "aggregate_op")
```
acmacalister, line 141: `aggregate_op` goes straight to the API as `aggregateOperator` without any validation. The tool description lists the 11 valid values, so an LLM is unlikely to pass garbage here, but if it does, SigNoz returns a 400 with an error message that has no `signoz:` context — hard to diagnose. Worth guarding with a simple allow-list:

```go
var validAggOps = map[string]bool{
    "avg": true, "sum": true, "min": true, "max": true,
    "count": true, "rate": true, "p50": true, "p75": true,
    "p90": true, "p95": true, "p99": true,
}

if !validAggOps[aggregateOp] {
    return mcp.ErrResult(fmt.Errorf("invalid aggregate_op %q: must be one of avg, sum, min, max, count, rate, p50, p75, p90, p95, p99", aggregateOp))
}
```

config/config_test.go
```diff
@@ -253,7 +253,7 @@ func TestEnabledIntegrations_Multiple(t *testing.T) {
 func TestDefaultConfig(t *testing.T) {
 	cfg := defaultConfig()
 	require.NotNil(t, cfg)
-	assert.Len(t, cfg.Integrations, 31)
+	assert.Len(t, cfg.Integrations, 32)
```
acmacalister, line 256: `signoz` is missing from the `expected` map in `TestDefaultConfig`. Every other integration has its credential keys verified there, so a future change to the signoz credentials struct (e.g., adding a `timeout` key or renaming `skip_verify`) won't be caught by the test. Worth adding:

```go
"signoz": {"api_key", "base_url", "skip_verify"},
```

acmacalister (commented):
Nice integration — good API surface coverage, clean hexagonal structure, and the fixes from the previous round (LimitReader, filter validation, strMs error handling, path escaping) all landed correctly. Three things worth looking at inline.

integrations/signoz/tools.go
```diff
@@ -0,0 +1,169 @@
...
+		Required:   []string{"service", "start", "end"},
+	},
+
+	// ── Query: Logs ──────────────────────────────────────────────────
+	{
+		Name: mcp.ToolName("signoz_search_logs"), Description: "Search and filter log entries. Supports attribute filters (severity, service, body text), ordering, and pagination. Start here for log exploration, debugging, and error investigation.",
+		Parameters: map[string]string{
+			"start":  "Start time in epoch milliseconds",
+			"end":    "End time in epoch milliseconds",
+			"filter": "Filter expression, e.g. \"severity_text = 'ERROR'\" or \"body CONTAINS 'timeout'\"",
```
acmacalister: `parseFilterItems` in `services.go` handles exactly one filter expression via `SplitN(filter, " ", 3)`, so only a single `key op value` triplet is accepted. The description says "Supports attribute filters" (plural), which could lead an LLM to attempt something like `"severity_text = 'ERROR' AND service = 'frontend'"` - that would fail with `invalid filter format`.

Worth clarifying the param description so the LLM knows this is a single-expression field. A small tweak like:

```go
"filter": "Single filter expression: key op value, e.g. severity_text = 'ERROR'. Only one expression per call.",
```

(`signoz_search_traces` already has a `service` param to layer in a second condition, so that path is covered.)

acmacalister (approved):
Really solid addition - 27 tools, comprehensive test coverage, retryable errors wired up, LimitReader, aggregate_op allow-list, all previous review feedback addressed. One small doc clarification inline.

integrations/signoz/services.go
```diff
@@ -0,0 +1,204 @@
...
+
+func listServices(ctx context.Context, s *signoz, args map[string]any) (*mcp.ToolResult, error) {
+	r := mcp.NewArgs(args)
+	start := r.Str("start")
+	end := r.Str("end")
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	data, err := s.post(ctx, "/api/v2/services", map[string]any{
+		"start": start,
```
acmacalister, line 23: The v2 service handlers (`listServices`, `topOperations`, `topLevelOperations`, `entryPointOperations`) forward `start` and `end` as raw strings to the API without going through `parseTimeRange`. Every other handler in the package (`searchLogs`, `searchTraces`, `queryMetrics`, `getServiceOverview`) validates timestamps first, so a value like `"not-a-number"` would get a clear `invalid epoch milliseconds` error from those tools but pass through silently here and return a confusing 400 from SigNoz.

Simple fix — same pattern already used just below in `getServiceOverview`:

```go
func listServices(ctx context.Context, s *signoz, args map[string]any) (*mcp.ToolResult, error) {
	r := mcp.NewArgs(args)
	start := r.Str("start")
	end := r.Str("end")
		return mcp.ErrResult(err)
	}
	startMs, endMs, err := parseTimeRange(start, end)
	if err != nil {
		return mcp.ErrResult(err)
	}
	data, err := s.post(ctx, "/api/v2/services", map[string]any{
		"start": startMs,
		"end":   endMs,
		"tags":  []any{},
	})
	...
}
```

`topOperations` and `entryPointOperations` follow the same pattern and would benefit from the same treatment.

acmacalister (commented):
Solid integration — clean architecture, thorough test coverage, and nice retryable error handling. One thing worth a look inline.

config/config.go
```diff
@@ -126,6 +126,10 @@ var envMapping = map[string]map[string]string{
 	"x": {
 		"bearer_token": "X_BEARER_TOKEN",
 	},
+	"signoz": {
+		"api_key":  "SIGNOZ_API_KEY",
+		"base_url": "SIGNOZ_BASE_URL",
+	},
 }
```
acmacalister, line 133: `skip_verify` is in the `defaultConfig()` credentials but not in `envMapping`, so `SIGNOZ_SKIP_VERIFY` doesn't get picked up automatically from the environment. The main use case for `skip_verify` is self-hosted SigNoz with a self-signed cert — exactly the kind of deployment where env-var configuration is the natural workflow. Worth adding alongside the other two:

```go
"signoz": {
    "api_key":     "SIGNOZ_API_KEY",
    "base_url":    "SIGNOZ_BASE_URL",
    "skip_verify": "SIGNOZ_SKIP_VERIFY",
},
```

And in `TestEnvMapping_ReturnsMapping` (currently only asserts `api_key`):
```go
assert.Equal(t, "SIGNOZ_SKIP_VERIFY", m["signoz"]["skip_verify"])
assert.Equal(t, "SIGNOZ_BASE_URL", m["signoz"]["base_url"])
```

integrations/signoz/services.go
```diff
@@ -0,0 +1,216 @@
...
+)
+
+func listServices(ctx context.Context, s *signoz, args map[string]any) (*mcp.ToolResult, error) {
+	r := mcp.NewArgs(args)
+	start := r.Str("start")
+	end := r.Str("end")
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	if _, _, err := parseTimeRange(start, end); err != nil {
```
acmacalister, line 19: Validate-and-discard is a bit surprising at first glance — the parsed `int64` values from `parseTimeRange` are thrown away and the original strings go into the request body. The v4 query handlers use the ints, so a reader might assume this is a bug and "fix" it. A short comment would prevent that:

```go
// v2 API expects epoch ms as strings, not integers.
if _, _, err := parseTimeRange(start, end); err != nil {
```

acmacalister (approved):
Really solid work — 27 tools, full test coverage, compaction specs, retryable errors, all the HTTP edge cases. Previous review feedback has all been addressed. One non-blocking thing inline (skip_verify env var) and a small comment suggestion for the v2 validate-and-discard pattern.

integrations/signoz/dashboards.go
```diff
@@ -0,0 +1,111 @@
...
+				tags = append(tags, t)
+			}
+		}
+	}
+
+	body := map[string]any{
+		"data": map[string]any{
+			"title":       title,
+			"description": description,
+			"tags":        tags,
```
acmacalister: When no tags are provided, `var tags []string` stays `nil`, which marshals to `"tags": null` in the request body. The SigNoz API probably accepts this, but it's safer to initialize to an empty slice so the field is always an array:

```go
tags := []string{}
if tagsStr != "" {
    // ...
}
```

acmacalister (approved):
Solid addition — 27 tools with full compaction coverage, good retryable error handling, and thorough tests. The prior round's fixes (io.LimitReader, timestamp validation, filter description) all landed correctly. One small thing inline.

integrations/signoz/signoz.go
```diff
@@ -0,0 +1,206 @@
...
+
+type signoz struct {
+	apiKey  string
+	baseURL string
+	client  *http.Client
+}
+
+func New() mcp.Integration {
+	return &signoz{
+		client: &http.Client{},
```
acmacalister, line 30: The `http.Client` has no `Timeout`, so a stalled SigNoz endpoint will hold a connection open until the parent context cancels. Every other adapter that owns its own client (cloudflare, readarr, notion, snowflake) sets `Timeout: 30 * time.Second`. The same applies to the `skip_verify` branch a few lines down.

```go
func New() mcp.Integration {
    return &signoz{
        client: &http.Client{Timeout: 30 * time.Second},
    }
}

// skip_verify branch in Configure():
s.client = &http.Client{
    Timeout: 30 * time.Second,
    Transport: &http.Transport{
        TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, // #nosec G402
    },
}
```

Worth adding a `TestNew_ClientHasTimeout` to match the pattern in `notion_test.go:26`.

acmacalister (commented):
Really solid integration — clean architecture, comprehensive test coverage, and most of the prior review feedback has been addressed. One new thing inline about the HTTP client timeout. The three earlier comments that are still open (skip_verify env mapping, nil tags, validate-and-discard comment) are worth closing out before merge.

PR #145, opened 2026-04-24
Add Nomad integration for cluster orchestration

aleksclark:
## Summary

- Add 28-tool Nomad integration covering jobs, allocations, nodes, deployments, evaluations, services, and cluster management
- Raw HTTP adapter against the Nomad v1 API (no SDK dependency) with `X-Nomad-Token` auth support
- Field compaction specs for all 18 read tools, keeping list responses compact for LLM consumption

## Tools

**Jobs** (6): list, get, versions, register, stop, force-evaluate
**Allocations** (6): list, get, job-allocs, stop, restart, logs
**Nodes** (5): list, get, node-allocs, drain, eligibility
**Deployments** (4): list, get, promote, fail
**Evaluations** (1): list
**Services** (1): list
**Cluster** (3): agent-self, cluster-status, gc

## Config

```json
{ "nomad": { "address": "http://host:4646", "token": "" } }
```

Env vars: `NOMAD_ADDR`, `NOMAD_TOKEN`

## Test plan

- [x] `make ci` passes (build, vet, test-race, lint, security)
- [x] Dispatch parity tests: all tools have handlers, no orphan handlers
- [x] Compaction spec parity: no orphan specs, no mutation tool specs
- [x] HTTP helper tests: success, 4xx, 204, 429 retryable, 5xx retryable
- [x] Handler tests: list_jobs, get_job, list_nodes, cluster_status, drain_node, node_eligibility
- [x] Configure tests: success, missing address, trailing slash trim, token optional
- [x] Healthy tests: healthy, unhealthy, nil client

🐨 Generated with Crush

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -33,7 +33,7 @@ func TestLoad_CreatesDefaultWhenMissing(t *testing.T) {
 	assert.NoError(t, err)
 
 	assert.Len(t, m.cfg.Integrations, 32)
-	for _, name := range []string{"github", "datadog", "linear", "sentry", "slack", "metabase", "aws", "posthog", "postgres", "clickhouse", "elasticsearch", "pganalyze", "rwx", "gmail", "notion", "ynab", "gcp", "suno", "amazon", "jira", "confluence", "readarr", "salesforce", "cloudflare", "digitalocean", "fly", "snowflake", "acp", "web", "botidentity", "x"} {
+	for _, name := range []string{"github", "datadog", "linear", "sentry", "slack", "metabase", "aws", "posthog", "postgres", "clickhouse", "elasticsearch", "pganalyze", "rwx", "gmail", "notion", "ynab", "gcp", "suno", "amazon", "jira", "confluence", "readarr", "salesforce", "cloudflare", "digitalocean", "fly", "snowflake", "acp", "web", "botidentity", "x", "nomad"} {
 		ic, ok := m.cfg.Integrations[name]
 		assert.True(t, ok, "missing default integration: %s", name)
 		assert.False(t, ic.Enabled)
@@ -280,6 +280,7 @@ func TestDefaultConfig(t *testing.T) {
 		"digitalocean":  {"api_token"},
 		"fly":           {"api_token", "base_url"},
 		"web":           {},
+		"nomad":         {"address", "token"},
 	}
 
 	for name, keys := range expected {
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -32,7 +32,7 @@ func TestLoad_CreatesDefaultWhenMissing(t *testing.T) {
 	_, err = os.Stat(path)
 	assert.NoError(t, err)
 
-	assert.Len(t, m.cfg.Integrations, 31)
+	assert.Len(t, m.cfg.Integrations, 32)
 	for _, name := range []string{"github", "datadog", "linear", "sentry", "slack", "metabase", "aws", "posthog", "postgres", "clickhouse", "elasticsearch", "pganalyze", "rwx", "gmail", "notion", "ynab", "gcp", "suno", "amazon", "jira", "confluence", "readarr", "salesforce", "cloudflare", "digitalocean", "fly", "snowflake", "acp", "web", "botidentity", "x"} {
 		ic, ok := m.cfg.Integrations[name]
 		assert.True(t, ok, "missing default integration: %s", name)
@@ -134,7 +134,7 @@ func TestSave(t *testing.T) {
 
 	var cfg mcp.Config
 	require.NoError(t, json.Unmarshal(data, &cfg))
-	assert.Len(t, cfg.Integrations, 31)
+	assert.Len(t, cfg.Integrations, 32)
 }
 
 func TestGet(t *testing.T) {
@@ -143,7 +143,7 @@ func TestGet(t *testing.T) {
 
 	cfg := m.Get()
 	assert.NotNil(t, cfg)
-	assert.Len(t, cfg.Integrations, 31)
+	assert.Len(t, cfg.Integrations, 32)
 }
 
 func TestUpdate(t *testing.T) {
@@ -253,7 +253,7 @@ func TestEnabledIntegrations_Multiple(t *testing.T) {
 func TestDefaultConfig(t *testing.T) {
 	cfg := defaultConfig()
 	require.NotNil(t, cfg)
-	assert.Len(t, cfg.Integrations, 31)
+	assert.Len(t, cfg.Integrations, 32)
 
 	expected := map[string][]string{
 		"github":        {"token", "client_id", "token_source"},
@@ -496,7 +496,9 @@ func TestEnvMapping_ReturnsMapping(t *testing.T) {
 	assert.Equal(t, "BOTIDENTITY_SLACK_CONFIG_TOKEN", m["botidentity"]["slack_config_token"])
 	assert.Equal(t, "BOTIDENTITY_SLACK_REFRESH_TOKEN", m["botidentity"]["slack_refresh_token"])
 	assert.Equal(t, "X_BEARER_TOKEN", m["x"]["bearer_token"])
-	assert.Len(t, m, 22)
+	assert.Equal(t, "NOMAD_ADDR", m["nomad"]["address"])
+	assert.Equal(t, "NOMAD_TOKEN", m["nomad"]["token"])
+	assert.Len(t, m, 23)
 }
 
 func TestToolGlobs_PersistThroughSaveLoad(t *testing.T) {
--- a/integrations/nomad/handlers.go
+++ b/integrations/nomad/handlers.go
@@ -1,0 +1,509 @@
+package nomad
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	mcp "github.com/daltoniam/switchboard"
+)
+
+func listJobs(ctx context.Context, n *nomad, args map[string]any) (*mcp.ToolResult, error) {
+	r := mcp.NewArgs(args)
+	ns := r.Str("namespace")
+	prefix := r.Str("prefix")
+	filter := r.Str("filter")
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	q := queryEncode(map[string]string{"namespace": ns, "prefix": prefix, "filter": filter})
+	data, err := n.get(ctx, "/v1/jobs%s", q)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	return mcp.RawResult(data)
+}
+
+func getJob(ctx context.Context, n *nomad, args map[string]any) (*mcp.ToolResult, error) {
+	r := mcp.NewArgs(args)
+	jobID := r.Str("job_id")
+	ns := r.Str("namespace")
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	q := queryEncode(map[string]string{"namespace": ns})
+	data, err := n.get(ctx, "/v1/job/%s%s", jobID, q)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	return mcp.RawResult(data)
+}
+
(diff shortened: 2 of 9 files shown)


integrations/nomad/nomad.go
```diff
@@ -0,0 +1,194 @@
...
+		req.Header.Set("Content-Type", "application/json")
+	}
+
+	resp, err := n.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer func() { _ = resp.Body.Close() }()
+
+	data, err := io.ReadAll(resp.Body)
```
acmacalister: `io.ReadAll(resp.Body)` has no size limit here. Every other HTTP adapter in the project (`cloudflare`, `notion`, `readarr`, `snowflake`) caps with `io.LimitReader`. For most endpoints this is fine, but `nomad_read_allocation_logs` proxies a Nomad log stream — a busy or misbehaving cluster could push unbounded data through this path.

```go

data, err := io.ReadAll(io.LimitReader(resp.Body, maxResponseSize))
```

acmacalister (commented):
Solid integration — clean architecture, great test coverage, and the compaction specs are well thought out. A couple of things worth looking at inline.

config/config_test.go
```diff
@@ -32,7 +32,7 @@ func TestLoad_CreatesDefaultWhenMissing(t *testing.T) {
 	_, err = os.Stat(path)
 	assert.NoError(t, err)
 
-	assert.Len(t, m.cfg.Integrations, 31)
+	assert.Len(t, m.cfg.Integrations, 32)
 	for _, name := range []string{"github", "datadog", "linear", "sentry", "slack", "metabase", "aws", "posthog", "postgres", "clickhouse", "elasticsearch", "pganalyze", "rwx", "gmail", "notion", "ynab", "gcp", "suno", "amazon", "jira", "confluence", "readarr", "salesforce", "cloudflare", "digitalocean", "fly", "snowflake", "acp", "web", "botidentity", "x"} {
```
acmacalister: The count was bumped to 32 but `"nomad"` wasn't added to this slice — so the test verifies 31 integration names by existence but silently skips the new one. Should be:

```go
for _, name := range []string{"github", "datadog", ... "x", "nomad"} {
```

Same gap exists in `TestDefaultConfig` around line 282 — `nomad` is missing from the `expected` map there too.

integrations/nomad/handlers.go
```diff
@@ -0,0 +1,509 @@
...
+	data, err := n.put(ctx, "/v1/system/gc", nil)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	return mcp.RawResult(data)
+}
+
+// parseDuration converts a human-readable duration (e.g., "1h", "30m") to nanoseconds
+// for the Nomad API. Returns 0 for empty or invalid input; -1 maps to -1ns (no deadline).
+func parseDuration(s string) int64 {
```
acmacalister, line 482: This custom duration parser only handles `s`, `m`, `h` — worth swapping for `time.ParseDuration`, which covers the same cases plus `ms`, `ns`, `µs`, and multi-unit strings like `"1h30m"`. The only special-case to preserve is `"-1"`:

```go
func parseDuration(s string) int64 {
    if s == "" { return 0 }
    if s == "-1" { return -1 }
    d, err := time.ParseDuration(s)
    if err != nil { return 0 }
    return d.Nanoseconds()
}
```

integrations/nomad/handlers.go
```diff
@@ -0,0 +1,492 @@
...
+	ignoreSystemJobs := r.Str("ignore_system_jobs")
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+
+	var body map[string]any
+	if enable {
+		drainSpec := map[string]any{}
+		if deadline != "" {
+			// Nomad expects a nanosecond duration — parse human-readable later if needed.
```
acmacalister: This comment is now stale — `parseDuration` already handles human-readable input and returns nanoseconds. The two lines here can be dropped:

```go
if deadline != "" {
    drainSpec["Deadline"] = parseDuration(deadline)
}
```

integrations/nomad/handlers.go
```diff
@@ -0,0 +1,492 @@
...
+			// For now, pass it through and let the API handle validation.
+			drainSpec["Deadline"] = parseDuration(deadline)
+		}
+		if strings.EqualFold(ignoreSystemJobs, "true") {
```
acmacalister: `ignore_system_jobs` is extracted as a string and then compared with `strings.EqualFold`, but `enable` (line 276) and `eligible` (line 320) both use `mcp.ArgBool`. Since the tool description already declares this as a `true/false` parameter, `r.Bool("ignore_system_jobs")` would be consistent and avoids the string comparison:

```go
ignoreSystemJobs := r.Bool("ignore_system_jobs")
    return mcp.ErrResult(err)
}
// ...
if ignoreSystemJobs {
    drainSpec["IgnoreSystemJobs"] = true
}
```

integrations/nomad/handlers.go
```diff
@@ -0,0 +1,492 @@
...
+
+func getJob(ctx context.Context, n *nomad, args map[string]any) (*mcp.ToolResult, error) {
+	r := mcp.NewArgs(args)
+	jobID := r.Str("job_id")
+	ns := r.Str("namespace")
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	q := queryEncode(map[string]string{"namespace": ns})
+	data, err := n.get(ctx, "/v1/job/%s%s", jobID, q)
```
acmacalister: Other adapters in the project (`salesforce`, `confluence`) wrap user-supplied path segments in `url.PathEscape`. Nomad IDs are typically UUIDs so this is low risk in practice, but it's worth being consistent — e.g. `n.get(ctx, "/v1/job/%s%s", url.PathEscape(jobID), q)`. Same pattern applies to `allocID`, `nodeID`, and `deploymentID` throughout this file.

acmacalister (commented):
Solid integration — clean architecture, good test coverage, and all the prior feedback has been addressed. Few small things inline.

integrations/nomad/handlers.go
```diff
@@ -0,0 +1,491 @@
...
+	enable, err := mcp.ArgBool(args, "enable")
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+
+	deadline := r.Str("deadline")
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	ignoreSystemJobs, _ := mcp.ArgBool(args, "ignore_system_jobs")
```
acmacalister: The error from `mcp.ArgBool` is silently discarded here. `ArgBool` only fails when the value is present but can't be converted to bool, so a caller passing e.g. an integer would silently skip the `IgnoreSystemJobs` field with no feedback. Consistent with how `enable` is handled just above, this should propagate the error:

```go
ignoreSystemJobs, err := mcp.ArgBool(args, "ignore_system_jobs")
if err != nil {
    return mcp.ErrResult(err)
}
```

Or use `r.Bool("ignore_system_jobs")` and let the Args reader accumulate it — then a single `r.Err()` check covers everything.

integrations/nomad/handlers.go
```diff
@@ -0,0 +1,491 @@
...
+	deploymentID := r.Str("deployment_id")
+	all := r.Str("all")
+	groups := r.Str("groups")
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+
+	body := map[string]any{
+		"DeploymentID": deploymentID,
+		"All":          all != "false",
```
acmacalister: The `all != "false"` expression works, but the default-true behavior here is load-bearing and easy to miss when reading. When `all` is omitted, `r.Str` returns `""`, and `"" != "false"` is `true` — which matches the documented default. But `r.Bool("all")` would return `false` when absent, silently flipping the default.

The simplest fix is to keep the string approach but make the intent explicit:

```go
// Default to true (promote all groups) when not specified
"All": all == "" || all == "true",
```

Or model it like a required-or-default bool by using a dedicated helper, similar to how `logType` defaults are set in `readAllocationLogs`.

acmacalister (commented):
Good shape overall — previous feedback has all been addressed. Two small things inline.

integrations/nomad/nomad.go
```diff
@@ -0,0 +1,195 @@
...
+
+type nomad struct {
+	address string
+	token   string
+	client  *http.Client
+}
+
+func New() mcp.Integration {
+	return &nomad{
+		client: &http.Client{},
```
acmacalister: Every other raw-HTTP adapter in the project sets a client timeout — `cloudflare`, `fly`, `jira`, `readarr`, `x`, `posthog`, and others all use `&http.Client{Timeout: 30 * time.Second}`. Without one, an unresponsive or partitioned Nomad cluster hangs this adapter's goroutine indefinitely. `nomad_read_allocation_logs` calling the client endpoint is especially at risk.

```go
client: &http.Client{Timeout: 30 * time.Second},
```

acmacalister (commented):
Nice work — 28 tools, clean architecture, all parity tests present, and all the previous review feedback addressed. One thing to fix inline: the `http.Client` needs a timeout to match the rest of the codebase.

integrations/nomad/handlers.go
```diff
@@ -0,0 +1,494 @@
...
+		body["TaskName"] = task
+	}
+	data, err := n.put(ctx, fmt.Sprintf("/v1/client/allocation/%s/restart", url.PathEscape(allocID)), body)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	return mcp.RawResult(data)
+}
+
+func readAllocationLogs(ctx context.Context, n *nomad, args map[string]any) (*mcp.ToolResult, error) {
```
acmacalister, line 189: `readAllocationLogs` is the only handler with a branching return path — when `plain=true` it returns `&mcp.ToolResult{Data: string(data)}` directly, bypassing `mcp.RawResult`. That branch is uncovered by the test suite. This is also the highest-risk tool (proxies log streams from the Nomad client endpoint), so a test that exercises both paths would be good to have:

```go
func TestReadAllocationLogs_PlainText(t *testing.T) {
	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		assert.Equal(t, "/v1/client/fs/logs/alloc-1", r.URL.Path)
		assert.Equal(t, "true", r.URL.Query().Get("plain"))
		w.Write([]byte(`2024/01/01 app started
`))
	}))
	defer ts.Close()

	n := &nomad{address: ts.URL, client: ts.Client()}
	result, err := n.Execute(context.Background(), "nomad_read_allocation_logs",
		map[string]any{"alloc_id": "alloc-1", "task": "web"})
	require.NoError(t, err)
	assert.False(t, result.IsError)
	assert.Contains(t, result.Data, "app started")
}
```

integrations/nomad/handlers.go
```diff
@@ -0,0 +1,494 @@
...
+	nodeID := r.Str("node_id")
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	enable, err := mcp.ArgBool(args, "enable")
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+
+	deadline := r.Str("deadline")
```
acmacalister, line 282: The `r` reader is used in two separate bursts here — `r.Str("node_id")` then `r.Err()`, then `r.Str("deadline")` then `r.Err()` again — with `mcp.ArgBool` calls in between. It's correct (NewArgs accumulates errors so the second check catches deadline extraction failures), but every other handler in this file extracts all args up front before the single `r.Err()` check. Worth pulling `deadline` up alongside `nodeID` to keep the pattern consistent:

```go
r := mcp.NewArgs(args)
nodeID := r.Str("node_id")
deadline := r.Str("deadline")
    return mcp.ErrResult(err)
}
enable, err := mcp.ArgBool(args, "enable")
...
```

acmacalister (commented):
Solid integration — all three rounds of prior feedback are addressed, the 10 MB read limit is in place, client timeout is set, the config test slice includes `nomad`, and the `all` default logic in `promoteDeployment` is now explicit. Two small things inline.

PR #146, opened 2026-04-26
Add switchboard self-management integration

aleksclark:
## Summary

- Adds a `switchboard` integration with 8 tools that let LLMs manage Switchboard itself — list/get/configure integrations, check health, browse/install/uninstall marketplace plugins, and get server info
- Operates directly on the `Services` struct (no HTTP) with marketplace attached post-construction via `SetMarketplace()`
- Includes field compaction specs for list/read tools, 30 tests with full dispatch parity coverage, and config defaults wiring

## Test plan

- [x] All 30 integration tests pass
- [x] Dispatch parity tests verify all tools have handlers and vice versa
- [x] Compact spec parity tests verify no orphan specs
- [x] Config test counts updated for new integration (33 → 34)

🐨 Generated with Crush

### DIFF
--- a/integrations/switchboard/switchboard.go
+++ b/integrations/switchboard/switchboard.go
@@ -1,0 +1,438 @@
+package switchboard
+
+import (
+	"context"
+	"sort"
+
+	mcp "github.com/daltoniam/switchboard"
+	"github.com/daltoniam/switchboard/marketplace"
+)
+
+var (
+	_ mcp.Integration                = (*switchboardInt)(nil)
+	_ mcp.FieldCompactionIntegration = (*switchboardInt)(nil)
+)
+
+type switchboardInt struct {
+	services    *mcp.Services
+	marketplace *marketplace.Manager
+}
+
+// New creates a switchboard self-management integration.
+// Call SetMarketplace after construction if marketplace is available.
+func New(services *mcp.Services) mcp.Integration {
+	return &switchboardInt{
+		services: services,
+	}
+}
+
+// SetMarketplace attaches the marketplace manager for plugin tools.
+// Must be called on the concrete type returned by New.
+func SetMarketplace(i mcp.Integration, mp *marketplace.Manager) {
+	if s, ok := i.(*switchboardInt); ok {
+		s.marketplace = mp
+	}
+}
+
+func (s *switchboardInt) Name() string { return "switchboard" }
+
+func (s *switchboardInt) Configure(_ context.Context, _ mcp.Credentials) error {
+	return nil
+}
+
+func (s *switchboardInt) Tools() []mcp.ToolDefinition { return tools }
+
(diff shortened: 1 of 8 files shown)


acmacalister (approved):

integrations/switchboard/switchboard.go
```diff
@@ -0,0 +1,438 @@
...
+		enabled := exists && ic.Enabled
+
+		if enabledOnly && !enabled {
+			continue
+		}
+
+		credKeys := s.services.Config.DefaultCredentialKeys(a.Name())
+		sort.Strings(credKeys)
+
+		results = append(results, integrationSummary{
```
acmacalister, line 99: The `Healthy` field is declared in `integrationSummary` and included in the compaction spec, but it's never populated — it'll always be `false` and omitted (due to `omitempty`). The tool description promises to show "which are healthy", so this is a silent lie to the LLM.

Either call `a.Healthy(ctx)` here (like `getIntegration` does), or drop the field from the struct and the compaction spec entirely. If health checks would be too expensive across all integrations, `getIntegration` is the right place to get that signal and the list tool should just omit it.

```go
results = append(results, integrationSummary{
    Name:      a.Name(),
    Enabled:   enabled,
    Healthy:   enabled && a.Healthy(ctx), // only check if enabled
    ToolCount: len(a.Tools()),
    ...
})
```

(Also note `listIntegrations` drops its `ctx` with `_ context.Context` — you'd need to thread it through to call `a.Healthy`.)

integrations/switchboard/switchboard.go
```diff
@@ -0,0 +1,438 @@
...
+			Credentials: mcp.Credentials{},
+		}
+	}
+
+	// Merge credentials if provided.
+	if credsRaw, ok := args["credentials"]; ok {
+		if credsMap, ok := credsRaw.(map[string]any); ok {
+			for k, v := range credsMap {
+				if vs, ok := v.(string); ok {
+					ic.Credentials[k] = vs
```
acmacalister, line 212: `GetIntegration` returns a pointer to the live `IntegrationConfig` stored in the config manager's map. Writing directly to `ic.Credentials[k]` mutates that shared struct without the config manager's write lock — concurrent calls to `configureIntegration` or any reader holding a reference would race on the map.

The safe approach is to copy the credentials into a new map before mutating:

```go
newCreds := make(mcp.Credentials, len(ic.Credentials))
for k, v := range ic.Credentials {
    newCreds[k] = v
}
// then merge into newCreds, and assign ic.Credentials = newCreds
```

The race detector doesn't catch this in the current tests because the mock returns a per-test pointer that nothing else touches concurrently, but the real `config.manager` would expose the shared pointer.

acmacalister (commented):
Nice addition — the self-management concept is solid, structure is clean, and test coverage is thorough. Two things worth looking at inline.

PR #148, opened 2026-04-26
Redesign admin UI: dashboard summary cards + integrations card grid

aleksclark:
## Summary

- **Dashboard**: Replace 4 generic stat cards (Integrations, Enabled, Healthy, Total Tools) with 3 color-coded summary cards: Connected (green), Disabled (muted), Errored (yellow). Add "Needs Attention" section listing errored integrations when any exist; remove full integration list (users navigate to /integrations instead).
- **Integrations page**: Replace flat list with 3-column responsive card grid organized into Needs Attention (errored, top), Connected, and Disabled sections. Each section sorted alphabetically. Errored section hidden when all integrations are healthy.
- **New component**: `StatCardColor` component with green/yellow/muted variants for dashboard summary cards.

## Test plan

- [x] `make ci` passes (build, vet, test-race, lint, security, govulncheck)
- [x] `TestDashboard_IntegrationCounts` — verifies connected/disabled/errored cards render and errored section appears
- [x] `TestDashboard_NoErroredHidesSection` — verifies "Needs Attention" hidden when no errored integrations
- [x] `TestIntegrationsList_CategorizedSections` — verifies all 3 sections render with card grid layout
- [x] `TestIntegrationsList_NoErroredHidesSection` — verifies errored section hidden when all healthy
- [x] `TestSortByName` — verifies alphabetical sort doesn't mutate original slice

🐒 Generated with Crush


web/web.go
```diff
@@ -177,24 +177,28 @@ func (w *WebServer) handleDashboard(rw http.ResponseWriter, r *http.Request) {
...
-		TotalIntegrations: len(summaries),
-		EnabledCount:      enabledCount,
-		HealthyCount:      healthyCount,
-		TotalTools:        totalTools,
-		Integrations:      summaries,
+		ConnectedCount:      connectedCount,
+		DisabledCount:       disabledCount,
+		ErroredCount:        erroredCount,
+		TotalTools:          totalTools,
+		Integrations:        summaries,
```
acmacalister: `Integrations: summaries` (and `TotalTools: totalTools` above it) are no longer referenced in `dashboard.templ` — the template was updated to use `ErroredIntegrations` and dropped the all-integrations list. These two fields are computed and heap-allocated on every page request but never consumed. Worth removing them from `DashboardData` and the handler to keep the struct lean.

web/web_test.go
```diff
@@ -633,6 +633,125 @@ func TestWasmModulesPage_ShowsName(t *testing.T) {
...
+	body := rr.Body.String()
+
+	// Should have the summary cards
+	assert.Contains(t, body, "Connected")
+	assert.Contains(t, body, "Disabled")
+	assert.Contains(t, body, "Errored")
+
+	// Should show errored integration in "Needs Attention" section
+	assert.Contains(t, body, "Needs Attention")
+	assert.Contains(t, body, "errored_int")
```
acmacalister, line 680: The assertions here confirm the three labels appear in the HTML, but not that the counts rendered are correct. Since we're adding `errored_int` (1 errored) and `disabled_int` (1 disabled) alongside the default `testint` (1 connected), we could assert the actual numeric values too:

```go
// Verify stat card values
assert.Contains(t, body, ">1<") // connected count
assert.Contains(t, body, ">1<") // or parse a bit more specifically
```

Even a coarse check like `assert.Contains(t, body, "1")` would catch a regression where all counts show 0. Not blocking, just an idea.

acmacalister (approved):
Clean UX improvement — the focused 'Needs Attention' section is a much better signal than the old flat list. Two small things inline.

web/templates/pages/dashboard.templ
```diff
@@ -11,13 +11,12 @@ import (
...
-	HealthyCount      int
-	TotalTools        int
-	Integrations      []IntegrationSummary
-	Metrics           *mcp.MetricsSnapshot
-	TopTools          []mcp.ToolRank
+	ConnectedCount      int
+	DisabledCount       int
+	ErroredCount        int
+	ErroredIntegrations []IntegrationSummary
+	Metrics             *mcp.MetricsSnapshot
```
acmacalister, line 18: `Integrations []IntegrationSummary` is no longer used anywhere in this template — the old integration list section was removed as part of this PR. The field is still being populated in `handleDashboard` (`Integrations: summaries`), which means we're allocating and passing a full slice on every dashboard load that nothing reads.

Worth removing the field from the struct and dropping the `Integrations: summaries` line from `web.go`.

web/templates/pages/dashboard.templ
```diff
@@ -11,13 +11,12 @@ import (
...
-	EnabledCount      int
-	HealthyCount      int
-	TotalTools        int
-	Integrations      []IntegrationSummary
-	Metrics           *mcp.MetricsSnapshot
-	TopTools          []mcp.ToolRank
+	ConnectedCount      int
+	DisabledCount       int
+	ErroredCount        int
+	ErroredIntegrations []IntegrationSummary
```
acmacalister, line 17: `TotalTools` is in the same boat — defined in the struct, calculated and set in `handleDashboard`, but not rendered anywhere in the updated template. Unless there's a plan to surface it somewhere, it can be removed.

web/templates/pages/integrations_list.templ
```diff
@@ -2,44 +2,83 @@ package pages
...
+func sortByName(items []IntegrationSummary) []IntegrationSummary {
+	sorted := make([]IntegrationSummary, len(items))
+	copy(sorted, items)
+	sort.Slice(sorted, func(i, j int) bool {
+		return sorted[i].Name < sorted[j].Name
+	})
+	return sorted
+}
+
+func intToStr(i int) string {
```
acmacalister, line 26: `intToStr` no longer has any callers — the new card template uses `fmt.Sprintf("%d tools", i.ToolCount)` directly. Worth deleting this function from the `.templ` file (and re-running `templ generate` to clean the generated file).

web/web.go
```diff
@@ -208,8 +209,25 @@ func (w *WebServer) handleDashboard(rw http.ResponseWriter, r *http.Request) {
 
 func (w *WebServer) handleIntegrationsList(rw http.ResponseWriter, r *http.Request) {
 	summaries := w.integrationSummaries(r.Context())
+
+	var errored, connected, disabled []pages.IntegrationSummary
+	for _, s := range summaries {
+		if !s.Enabled {
+			disabled = append(disabled, s)
```
acmacalister, line 216: The `if !s.Enabled / else if s.Healthy / else` categorization block now appears identically in both `handleDashboard` and `handleIntegrationsList`. Not a big deal at this scale, but if we ever need to adjust the logic (e.g. add a "degraded" state), we'd need to update it in two places. A small helper like `categorizeIntegrations(summaries []pages.IntegrationSummary) (connected, disabled, errored []pages.IntegrationSummary)` would give both handlers a single source of truth.

web/templates/pages/dashboard.templ
```diff
@@ -249,26 +247,25 @@ templ Dashboard(page layouts.PageData, data DashboardData) {
...
-					</div>
-				</a>
-			}
-		</section>
+		if len(data.ErroredIntegrations) > 0 {
+			<section class="metrics-section">
+				<h2 class="section-title">Needs Attention</h2>
+				for _, i := range data.ErroredIntegrations {
+					<a href={ templ.SafeURL("/integrations/" + i.Name) } class="integration-link">
+						<div>
```
acmacalister, line 255: The integrations list page wraps each section in `sortByName(...)` for consistent alphabetical ordering, but the "Needs Attention" items here iterate `data.ErroredIntegrations` in registry order. Minor, but if multiple integrations are errored the order will be non-deterministic. Could use `sortByName` here too, either in the template loop or by sorting the slice before passing it in `handleDashboard`.

acmacalister (approved):
Nice cleanup — the categorized card layout is a clear UX improvement and the test coverage is solid. Left a few small things inline, mainly two unused struct fields that sneaked through.

web/web.go
```diff
@@ -208,8 +209,25 @@ func (w *WebServer) handleDashboard(rw http.ResponseWriter, r *http.Request) {
 
 func (w *WebServer) handleIntegrationsList(rw http.ResponseWriter, r *http.Request) {
 	summaries := w.integrationSummaries(r.Context())
+
+	var errored, connected, disabled []pages.IntegrationSummary
+	for _, s := range summaries {
+		if !s.Enabled {
```
acmacalister, line 215: The categorization triage (`!s.Enabled → disabled / s.Healthy → connected / else → errored`) is duplicated verbatim from `handleDashboard` above. If a new state gets added later (e.g. "degraded" or "stale") we'd need to update both in sync. A small unexported helper keeps that logic in one place:

```go
func categorizeIntegrations(summaries []pages.IntegrationSummary) (connected, disabled, errored []pages.IntegrationSummary) {
    for _, s := range summaries {
        if !s.Enabled {
            disabled = append(disabled, s)
        } else if s.Healthy {
            connected = append(connected, s)
        } else {
            errored = append(errored, s)
        }
    }
    return
}
```

web/templates/pages/dashboard.templ
```diff
@@ -249,26 +247,25 @@ templ Dashboard(page layouts.PageData, data DashboardData) {
...
-						<span>{ fmt.Sprintf("%d tools", i.ToolCount) }</span>
-						<span>→</span>
-					</div>
-				</a>
-			}
-		</section>
+		if len(data.ErroredIntegrations) > 0 {
+			<section class="metrics-section">
+				<h2 class="section-title">Needs Attention</h2>
+				for _, i := range data.ErroredIntegrations {
```
acmacalister, line 253: The integrations list page wraps every grid with `sortByName(...)`, but this range iterates `data.ErroredIntegrations` in whatever order `integrationSummaries` returns them. With more than a couple errored integrations the ordering will be unpredictable. Worth either sorting in the handler before assigning to `ErroredIntegrations`, or calling `sortByName` here the same way the integrations list does:

```diff
-			for _, i := range data.ErroredIntegrations {
+			for _, i := range sortByName(data.ErroredIntegrations) {
```

web/templates/pages/integrations_list.templ
acmacalister, line 26: `intToStr` was carried over from the old list layout but is no longer called — the card footer now uses `fmt.Sprintf` directly. Since the `import "fmt"` is still needed for the card template, only the function itself can be dropped.

acmacalister (approved):
Clean refactor — the 3-category model is a clear UX improvement and the new tests verify all the key states. A few small things inline worth a look before merging.

PR #149, opened 2026-04-26
Fix switchboard integration review feedback from #146

aleksclark:
## Summary

Addresses review comments from #146 that were left after the merge:

- **Populate `Healthy` field in `listIntegrations`**: The field was declared in the struct and compaction spec but never set — always `false`/omitted. Now threads `ctx` through and calls `a.Healthy(ctx)` for enabled integrations, matching the behavior already present in `getIntegration` and `checkHealth`.
- **Copy `IntegrationConfig` before mutating in `configureIntegration`**: `GetIntegration` returns a pointer to the live config manager struct. Writing directly to `ic.Credentials` raced with concurrent readers. Now creates a fresh `IntegrationConfig` with copied credentials before merging.

## Test plan

- [x] Added `TestListIntegrations_HealthyNotSetWhenDisabled` — verifies `healthy` is omitted via `omitempty` for disabled integrations
- [x] Added `TestConfigureIntegration_DoesNotMutateOriginalCredentials` — verifies the original credentials map is not mutated
- [x] Existing `TestListIntegrations` now asserts `healthy: true` for enabled+healthy integrations

🐨 Generated with Crush

### DIFF
--- a/integrations/switchboard/switchboard.go
+++ b/integrations/switchboard/switchboard.go
@@ -73,7 +73,7 @@ var dispatch = map[mcp.ToolName]handlerFunc{
 }
 
 // listIntegrations returns all registered integrations with their status.
-func listIntegrations(_ context.Context, s *switchboardInt, args map[string]any) (*mcp.ToolResult, error) {
+func listIntegrations(ctx context.Context, s *switchboardInt, args map[string]any) (*mcp.ToolResult, error) {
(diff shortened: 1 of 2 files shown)


acmacalister (approved):
Clean PR — builds, all 38 packages pass under the race detector, lint clean. Both fixes are correct: the copy-on-write in configureIntegration eliminates the shared-pointer race, and the healthy guard matches the pattern already used in getIntegration and checkHealth. LGTM.

PR #150, opened 2026-04-30
feat(posthog): add posthog_query tool and expose insight result fields

daltoniam:
## Summary

- Add `posthog_query` tool for synchronous HogQL execution via `/api/projects/{id}/query/`. Lets agents run ad-hoc analytics inline (no insight artifact, no async-cache wait, no cleanup cron needed).
- Expose `result`, `hogql`, `is_cached`, `last_refresh`, and `query_status` on `posthog_list_insights` and `posthog_get_insight`. These fields were silently stripped by the compaction spec, so cached insights returned only metadata.

## Why

An agent needs to answer ad-hoc analytics questions. The agent's only path today is `posthog_create_insight` + `posthog_get_insight`, which has two problems:

1. The compact specs drop the computed `result` field, so even when an insight is cached the agent can't see the data.
2. Freshly POSTed insights have `result: null` because PostHog computes lazily — there's no synchronous query path through the existing tools.

`posthog_query` hits the same `/query/` endpoint the PostHog UI uses for HogQL Editor, returning `columns`, `results`, and execution metadata inline. This is the proper primitive for ad-hoc analytics; insights remain the right tool for *persisting* charts.

## Test plan

- [x] `gofmt -w` + `go vet ./integrations/posthog/...`
- [x] `go build ./...`
- [x] `go test ./...` (all packages green; existing posthog spec tests still pass — they don't pin specific field names)
- [ ] CI: golangci-lint (not installed locally)

### DIFF
--- a/integrations/posthog/compact_specs.go
+++ b/integrations/posthog/compact_specs.go
@@ -33,6 +33,12 @@ var rawFieldCompactionSpecs = map[mcp.ToolName][]string{
 	mcp.ToolName("posthog_list_insights"): {"results[].id", "results[].name", "results[].short_id", "results[].filters", "results[].query", "results[].result", "results[].hogql", "results[].is_cached", "results[].last_refresh", "results[].query_status", "results[].created_at", "results[].created_by.email", "results[].last_modified_at"},
 	mcp.ToolName("posthog_get_insight"):   {"id", "name", "short_id", "description", "filters", "query", "result", "hogql", "is_cached", "last_refresh", "query_status", "created_at", "created_by.email", "last_modified_at"},
 
+	// ── Query (HogQL) ───────────────────────────────────────────────
+	// /query/ responses include columns, results (2D row data), types, hogql,
+	// query (echo of the request), timings, and query_status. timings/types
+	// are rarely useful to the LLM and are dropped to save tokens.
+	mcp.ToolName("posthog_query"): {"columns", "results", "hogql", "query_status", "is_cached"},
+
 	// ── Persons ─────────────────────────────────────────────────────
 	mcp.ToolName("posthog_list_persons"): {"results[].id", "results[].distinct_ids", "results[].properties.email", "results[].properties.name", "results[].created_at"},
 	mcp.ToolName("posthog_get_person"):   {"id", "distinct_ids", "properties", "created_at"},
--- a/integrations/posthog/compact_specs.go
+++ b/integrations/posthog/compact_specs.go
@@ -27,8 +27,11 @@ var rawFieldCompactionSpecs = map[mcp.ToolName][]string{
 	mcp.ToolName("posthog_list_cohort_persons"): {"results[].id", "results[].distinct_ids", "results[].properties", "results[].created_at"},
 
 	// ── Insights ────────────────────────────────────────────────────
-	mcp.ToolName("posthog_list_insights"): {"results[].id", "results[].name", "results[].short_id", "results[].filters", "results[].created_at", "results[].created_by.email", "results[].last_modified_at"},
-	mcp.ToolName("posthog_get_insight"):   {"id", "name", "short_id", "description", "filters", "query", "result", "created_at", "created_by.email", "last_modified_at"},
+	// Note: result/hogql/is_cached/last_refresh/query_status expose the computed
+	// query output. PostHog computes results lazily, so freshly-created insights
+	// may have result: null until the first refresh (is_cached: true).
+	mcp.ToolName("posthog_list_insights"): {"results[].id", "results[].name", "results[].short_id", "results[].filters", "results[].query", "results[].result", "results[].hogql", "results[].is_cached", "results[].last_refresh", "results[].query_status", "results[].created_at", "results[].created_by.email", "results[].last_modified_at"},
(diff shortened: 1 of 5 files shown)


integrations/posthog/insights.go
```diff
@@ -135,3 +135,38 @@ func deleteInsight(ctx context.Context, p *posthog, args map[string]any) (*mcp.T
 	}
 	return mcp.RawResult(data)
 }
+
+func runQuery(ctx context.Context, p *posthog, args map[string]any) (*mcp.ToolResult, error) {
```
acmacalister, line 139: No handler test for `runQuery`. Every other handler in this file (feature flags, persons, annotations, etc.) has at least a happy-path `httptest.NewServer` test. Without one here, it's hard to catch regressions in the body structure or optional-field logic.

Worth adding a `TestRunQuery` that checks the POST goes to `/api/projects/{id}/query/`, the body contains `{"query": {"kind": "HogQLQuery", "query": "..."}}`, and that `client_query_id`/`refresh` are included when provided and absent when not. The empty-query validation path would also benefit from a test.

acmacalister (commented):
Clean implementation overall — the lazy-compute fix for `result` on existing insight tools is a nice catch, and the HogQL query wrapper is well-structured. Two things worth addressing inline.

integrations/posthog/compact_specs.go
```diff
@@ -27,8 +27,11 @@ var rawFieldCompactionSpecs = map[mcp.ToolName][]string{
...
-	mcp.ToolName("posthog_list_insights"): {"results[].id", "results[].name", "results[].short_id", "results[].filters", "results[].created_at", "results[].created_by.email", "results[].last_modified_at"},
-	mcp.ToolName("posthog_get_insight"):   {"id", "name", "short_id", "description", "filters", "query", "result", "created_at", "created_by.email", "last_modified_at"},
+	// Note: result/hogql/is_cached/last_refresh/query_status expose the computed
+	// query output. PostHog computes results lazily, so freshly-created insights
+	// may have result: null until the first refresh (is_cached: true).
+	mcp.ToolName("posthog_list_insights"): {"results[].id", "results[].name", "results[].short_id", "results[].filters", "results[].query", "results[].result", "results[].hogql", "results[].is_cached", "results[].last_refresh", "results[].query_status", "results[].created_at", "results[].created_by.email", "results[].last_modified_at"},
+	mcp.ToolName("posthog_get_insight"):   {"id", "name", "short_id", "description", "filters", "query", "result", "hogql", "is_cached", "last_refresh", "query_status", "created_at", "created_by.email", "last_modified_at"},
```
acmacalister, line 34: `posthog_query` is the one new tool without a compaction spec. The `/query/` response includes `results` (a 2D array of row data), `columns`, `types`, `hogql`, `query`, `timings`, and `query_status` — it can be quite wide. Without a spec the full response goes to the LLM unfiltered, which becomes a problem for queries that return many rows or columns.

Something like this would cover the useful fields while keeping the token budget sane:

```go
mcp.ToolName("posthog_query"): {"columns", "results", "hogql", "query_status", "is_cached"},
```

`timings` and `types` are rarely useful to an LLM and can stay out.

integrations/posthog/insights.go
daltoniam: Added in 541e84a: TestRunQuery (happy path + body shape), TestRunQuery_OptionalFields (client_query_id/refresh propagation), TestRunQuery_EmptyQuery (validation, server is asserted not-called), and TestRunQuery_UsesProjectArg (project_id arg fallback when no default configured).

integrations/posthog/compact_specs.go
daltoniam: Added in 541e84a with exactly that field set — kept columns/results/hogql/query_status/is_cached, dropped timings/types.

integrations/posthog/compact_specs.go
```diff
@@ -27,8 +27,17 @@ var rawFieldCompactionSpecs = map[mcp.ToolName][]string{
...
+
+	// ── Query (HogQL) ───────────────────────────────────────────────
+	// /query/ responses include columns, results (2D row data), types, hogql,
+	// query (echo of the request), timings, and query_status. timings/types
+	// are rarely useful to the LLM and are dropped to save tokens.
+	mcp.ToolName("posthog_query"): {"columns", "results", "hogql", "query_status", "is_cached"},
```
acmacalister, line 40: The generic parity tests (`TestFieldCompactionSpecs_NoOrphanSpecs`, `TestFieldCompactionSpecs_AllParse`) will catch a broken spec but don't pin the exact field list. A named shape test in `compact_specs_test.go` would make it obvious if someone accidentally drops `results` or `columns`:

```go
func TestFieldCompactionSpec_ReturnsFieldsForQueryTool(t *testing.T) {
	p := &posthog{}
	fields, ok := p.CompactSpec("posthog_query")
	require.True(t, ok)
	names := make([]string, len(fields))
	for i, f := range fields {
		names[i] = f.Name
	}
	assert.ElementsMatch(t, []string{"columns", "results", "hogql", "query_status", "is_cached"}, names)
}
```

Not blocking — just a nice regression guard given this is a POST-based query tool with a hand-curated field list.

acmacalister (approved):
Solid addition — `posthog_query` fills a real gap, the handler is clean, and the four new tests cover the important paths. Previous review feedback has all been addressed. One optional test suggestion inline.

PR #151, opened 2026-05-01
feat(wasm): add h2c support for HTTP/2 cleartext in plugins

aleksclark:
## Summary

- Adds an h2c (HTTP/2 cleartext) transport to the WASM host HTTP proxy, enabling plugins to talk to plaintext HTTP/2 endpoints like gRPC servers
- Plugins opt in per-request by setting `X-H2C` header — the host strips it before forwarding and routes through `http2.Transport` with `AllowHTTP` + plaintext dialer
- Adds `host_http_request_h2c()` convenience wrapper to the Rust guest SDK

## Test plan

- [x] `TestDoHostHTTP_H2C` — verifies HTTP/2.0 protocol negotiation over h2c
- [x] `TestDoHostHTTP_H2C_HeaderNotForwarded` — verifies X-H2C stripped from upstream request
- [x] `TestDoHostHTTP_HTTP1_Default` — verifies default path stays HTTP/1.1
- [x] `TestDoHostHTTP_H2C_POST` — verifies request body works over h2c

🐶 Generated with Crush

### DIFF
--- a/wasm/host.go
+++ b/wasm/host.go
@@ -3,14 +3,17 @@ package wasm
 import (
 	"bytes"
 	"context"
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"io"
 	"log/slog"
+	"net"
 	"net/http"
 	"time"
 
 	"github.com/tetratelabs/wazero/api"
+	"golang.org/x/net/http2"
 )
 
 type httpRequest struct {
@@ -28,7 +31,25 @@ type httpResponse struct {
 
 const maxHostResponseSize = 10 * 1024 * 1024
 
-var hostHTTPClient = &http.Client{Timeout: 30 * time.Second}
+// h2cHeaderKey is the header plugins set to request HTTP/2 cleartext (h2c)
+// transport. Any non-empty value enables h2c for that request.
+const h2cHeaderKey = "X-H2C"
+
+var (
+	hostHTTPClient = &http.Client{Timeout: 30 * time.Second}
+
+	// h2cClient uses an http2.Transport configured for cleartext (no TLS)
+	// connections. Plugins opt in by setting the X-H2C header.
+	h2cClient = &http.Client{
+		Timeout: 30 * time.Second,
+		Transport: &http2.Transport{
+			AllowHTTP: true,
+			DialTLSContext: func(ctx context.Context, network, addr string, _ *tls.Config) (net.Conn, error) {
+				return (&net.Dialer{}).DialContext(ctx, network, addr)
+			},
+		},
+	}
+)
 
 func hostHTTPRequest(ctx context.Context, mod api.Module, ptrSize uint64) uint64 {
 	ptr, size := unpackPtrSize(ptrSize)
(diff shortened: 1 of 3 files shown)


acmacalister (commented):
Clean PR overall — extraction of `doHostHTTP` is a nice structural improvement and the four test cases cover the important scenarios well. A couple of small things inline.

wasm/host.go
```diff
@@ -43,50 +64,72 @@ func hostHTTPRequest(ctx context.Context, mod api.Module, ptrSize uint64) uint64
...
-	resp, err := hostHTTPClient.Do(httpReq)
+	client := hostHTTPClient
+	if useH2C {
+		client = h2cClient
+	}
+
+	resp, err := client.Do(httpReq)
 	if err != nil {
-		return writeErrorResponse(ctx, mod, fmt.Sprintf("http error: %v", err))
+		return nil, err
```
acmacalister, line 114: `client.Do` is the only error in this function returned without wrapping — every other return wraps with `fmt.Errorf("...: %w", err)`. A bare network error like `dial tcp: connection refused` loses the call-site context. Worth wrapping for consistency:

```go
resp, err := client.Do(httpReq)
if err != nil {
    return nil, fmt.Errorf("do request: %w", err)
}
```

wasm/host_test.go
```diff
@@ -0,0 +1,127 @@
...
+		t.Errorf("server saw proto = %q, want HTTP/1.1", gotProto)
+	}
+}
+
+func TestDoHostHTTP_H2C_POST(t *testing.T) {
+	var gotProto, gotBody string
+	base := newH2CServer(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		gotProto = r.Proto
+		b := make([]byte, 1024)
+		n, _ := r.Body.Read(b)
```
acmacalister, line 104: `r.Body.Read(b)` reads at most 1024 bytes in a single syscall — for this 17-byte test body that's fine, but it'll silently truncate if the body grows and make the assertion fail in a confusing way. `io.ReadAll` is safer here:

```go
gotBodyBytes, _ := io.ReadAll(r.Body)
gotBody = string(gotBodyBytes)
```

wasm/host_test.go
```diff
@@ -0,0 +1,127 @@
...
+	go func() { _ = srv.Serve(ln) }()
+	t.Cleanup(func() { srv.Close() })
+
+	return fmt.Sprintf("http://%s", ln.Addr().String())
+}
+
+func TestDoHostHTTP_H2C(t *testing.T) {
+	var gotProto string
+	base := newH2CServer(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		gotProto = r.Proto
```
acmacalister, line 34: `gotProto` is written in the handler goroutine and read in the test goroutine after `doHostHTTP` returns. In practice this is safe because the HTTP response send/receive creates a happens-before edge, and the race detector doesn't flag it. But Go's memory model doesn't formally guarantee it without explicit sync. A `sync.Mutex` around the write/read, or encoding the observed value into the response body, would make this airtight for readers doing a strict model analysis.

PR #152, opened 2026-05-01
feat(rwx): upgrade integration for rwx CLI v3.13

aleksclark:
## Summary

- **Fix `rwx_get_run_results`**: parse actual CLI JSON output (`RunID`, `ResultStatus`, `Completed`, `Prompt`) — the old handler expected fields (`tasks`, `result`, `execution`, `duration_seconds`) that the CLI never returned
- **8 new tools**: `rwx_dispatch_run`, `rwx_docs_search`, `rwx_docs_pull`, `rwx_vaults_var_show/set/delete`, `rwx_vaults_secret_set/delete`
- **Enhanced existing tools**: branch/commit lookup for results, `--task` key resolution across all log and results tools, `--title`/`--init`/`--fail-fast` for launches
- **Min CLI version**: bumped to 3.13.0

## Test plan

- [x] `make ci` passes (build, vet, test-race, lint, security)
- [x] `parseResultsPrompt` tested with empty, failed-tasks-with-artifacts, and build-error prompts
- [x] Dispatch parity tests pass (`TestDispatchMap_AllToolsCovered`, `TestDispatchMap_NoOrphanHandlers`)
- [x] Compaction spec parity tests pass
- [ ] Manual smoke test with live RWX runs

🐾 Generated with Crush

### DIFF
--- a/integrations/rwx/docs.go
+++ b/integrations/rwx/docs.go
@@ -1,0 +1,89 @@
+package rwx
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	mcp "github.com/daltoniam/switchboard"
+)
+
+func docsSearch(_ context.Context, r *rwx, args map[string]any) (*mcp.ToolResult, error) {
+	ra := mcp.NewArgs(args)
+	query := ra.Str("query")
+	limit := ra.Int("limit")
+	if err := ra.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	if limit <= 0 {
+		limit = 5
+	}
+
+	cmdArgs := []string{"docs", "search", query, "--output", "json", "--limit", fmt.Sprintf("%d", limit)}
+	output, err := r.runRWXCommand(cmdArgs, 30000)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+
+	var parsed struct {
+		Query     string `json:"Query"`
+		TotalHits int    `json:"TotalHits"`
+		Results   []struct {
+			URL   string `json:"url"`
+			Path  string `json:"path"`
+			Title string `json:"title"`
+			Body  string `json:"body"`
+		} `json:"Results"`
+	}
+	if err := json.Unmarshal([]byte(output), &parsed); err != nil {
+		return mcp.ErrResult(fmt.Errorf("parse docs search: %w", err))
+	}
+
+	var results []map[string]any
+	for _, doc := range parsed.Results {
+		entry := map[string]any{
+			"title": doc.Title,
+			"url":   doc.URL,
+			"path":  doc.Path,
+		}
+		if len(doc.Body) > 500 {
+			entry["snippet"] = doc.Body[:500] + "..."
+		} else {
+			entry["snippet"] = doc.Body
+		}
+		results = append(results, entry)
+	}
+
+	return mcp.JSONResult(map[string]any{
+		"query":      parsed.Query,
+		"total_hits": parsed.TotalHits,
+		"count":      len(results),
+		"results":    results,
+	})
+}
+
+func docsPull(_ context.Context, r *rwx, args map[string]any) (*mcp.ToolResult, error) {
+	urlOrPath, err := mcp.ArgStr(args, "url_or_path")
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+
+	cmdArgs := []string{"docs", "pull", urlOrPath, "--output", "json"}
+	output, err := r.runRWXCommand(cmdArgs, 30000)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+
+	var parsed struct {
+		URL  string `json:"URL"`
+		Body string `json:"Body"`
+	}
+	if err := json.Unmarshal([]byte(output), &parsed); err != nil {
+		return mcp.ErrResult(fmt.Errorf("parse docs pull: %w", err))
+	}
+
+	return mcp.JSONResult(map[string]any{
+		"url":     parsed.URL,
+		"content": parsed.Body,
+	})
+}
--- a/integrations/rwx/logs.go
+++ b/integrations/rwx/logs.go
@@ -58,35 +58,44 @@ func (c *logCache) set(id, logs string) {
 
 // --- Log download ---
 
-func downloadLogs(ctx context.Context, r *rwx, id string) (string, error) {
-	if cached, ok := r.logCache.get(id); ok {
+func downloadLogs(ctx context.Context, r *rwx, id string, taskKey ...string) (string, error) {
+	cacheKey := id
+	if len(taskKey) > 0 && taskKey[0] != "" {
+		cacheKey = id + ":" + taskKey[0]
+	}
+	if cached, ok := r.logCache.get(cacheKey); ok {
 		return cached, nil
 	}
 
-	logs, err := r.downloadLogsFromRWX(id)
+	logs, err := r.downloadLogsFromRWX(id, taskKey...)
 	if err != nil {
 		return "", err
 	}
 
 	go func() {
-		status, isComplete, err := fetchRunStatus(ctx, r, id)
+		bgCtx := context.WithoutCancel(ctx)
+		status, isComplete, err := fetchRunStatus(bgCtx, r, id)
 		_ = status
 		if err == nil && isComplete {
-			r.logCache.set(id, logs)
+			r.logCache.set(cacheKey, logs)
 		}
 	}()
 
 	return logs, nil
 }
 
-func (r *rwx) downloadLogsFromRWX(id string) (string, error) {
+func (r *rwx) downloadLogsFromRWX(id string, taskKey ...string) (string, error) {
 	outputDir, err := os.MkdirTemp("", fmt.Sprintf("rwx-logs-%s-", id))
 	if err != nil {
 		return "", fmt.Errorf("create temp dir: %w", err)
 	}
 	defer func() { _ = os.RemoveAll(outputDir) }()
 
-	_, err = r.runRWXCommand([]string{"logs", id, "--output-dir", outputDir, "--auto-extract", "--output", "json"}, 0)
+	cmdArgs := []string{"logs", id, "--output-dir", outputDir, "--auto-extract", "--output", "json"}
+	if len(taskKey) > 0 && taskKey[0] != "" {
+		cmdArgs = append(cmdArgs, "--task", taskKey[0])
+	}
+	_, err = r.runRWXCommand(cmdArgs, 0)
 	if err != nil {
 		return "", err
 	}
@@ -140,12 +149,26 @@ func findLogFiles(dir string) ([]string, error) {
 // --- Log tool handlers ---
 
 func getTaskLogs(ctx context.Context, r *rwx, args map[string]any) (*mcp.ToolResult, error) {
-	taskIDRaw, err := mcp.ArgStr(args, "task_id")
-	if err != nil {
+	ra := mcp.NewArgs(args)
+	taskIDRaw := ra.Str("task_id")
+	runIDRaw := ra.Str("run_id")
+	taskKey := ra.Str("task_key")
+	if err := ra.Err(); err != nil {
 		return mcp.ErrResult(err)
 	}
-	id := extractRunID(taskIDRaw)
-	logs, err := downloadLogs(ctx, r, id)
+
+	var id string
+	var logTaskKey string
+	if taskKey != "" && runIDRaw != "" {
+		id = extractRunID(runIDRaw)
+		logTaskKey = taskKey
+	} else if taskIDRaw != "" {
+		id = extractRunID(taskIDRaw)
+	} else {
+		return mcp.ErrResult(fmt.Errorf("either task_id or run_id+task_key is required"))
+	}
+
+	logs, err := downloadLogs(ctx, r, id, logTaskKey)
 	if err != nil {
 		return mcp.ErrResult(err)
 	}
@@ -186,6 +209,7 @@ func headLogs(ctx context.Context, r *rwx, args map[string]any) (*mcp.ToolResult
 	idRaw := ra.Str("id")
 	numLines := ra.Int("lines")
 	offset := ra.Int("offset")
+	taskKey := ra.Str("task_key")
 	if err := ra.Err(); err != nil {
 		return mcp.ErrResult(err)
 	}
@@ -194,7 +218,7 @@ func headLogs(ctx context.Context, r *rwx, args map[string]any) (*mcp.ToolResult
 		numLines = maxLinesPerPage
 	}
 
-	logs, err := downloadLogs(ctx, r, id)
+	logs, err := downloadLogs(ctx, r, id, taskKey)
 	if err != nil {
 		return mcp.ErrResult(err)
 	}
@@ -232,6 +256,7 @@ func tailLogs(ctx context.Context, r *rwx, args map[string]any) (*mcp.ToolResult
 	idRaw := ra.Str("id")
 	numLines := ra.Int("lines")
 	offset := ra.Int("offset")
+	taskKey := ra.Str("task_key")
 	if err := ra.Err(); err != nil {
 		return mcp.ErrResult(err)
 	}
@@ -240,7 +265,7 @@ func tailLogs(ctx context.Context, r *rwx, args map[string]any) (*mcp.ToolResult
 		numLines = maxLinesPerPage
 	}
 
-	logs, err := downloadLogs(ctx, r, id)
+	logs, err := downloadLogs(ctx, r, id, taskKey)
 	if err != nil {
 		return mcp.ErrResult(err)
 	}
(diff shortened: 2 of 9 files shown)


integrations/rwx/proxy.go
```diff
@@ -23,10 +23,12 @@ var cliToMCPReplacements = []struct {
 	{regexp.MustCompile("`rwx logs[^`]*`"), "the log tools (rwx_get_task_logs, rwx_head_logs, rwx_tail_logs, rwx_grep_logs)"},
 	{regexp.MustCompile("`rwx results[^`]*`"), "the rwx_get_run_results tool"},
 	{regexp.MustCompile("`rwx artifacts[^`]*`"), "the rwx_get_artifacts tool"},
+	{regexp.MustCompile("`rwx dispatch[^`]*`"), "the rwx_dispatch_run tool"},
```
acmacalister, line 26: The two new `rwx dispatch` replacements are added here but `TestTransformCLIReferences` in `rwx_test.go` doesn't cover them. Worth adding a couple of cases to that table:

```go
{"`rwx dispatch abc`", "rwx_dispatch_run"},
{"Use rwx dispatch to trigger", "rwx_dispatch_run"},
```

integrations/rwx/runs.go
```diff
@@ -92,6 +101,84 @@ func launchCIRun(_ context.Context, r *rwx, args map[string]any) (*mcp.ToolResul
 	})
 }
 
+func dispatchRun(_ context.Context, r *rwx, args map[string]any) (*mcp.ToolResult, error) {
```
acmacalister, line 107: `dispatchRun` is entirely new but has no handler-level tests. The dispatch parity test confirms routing is wired up, but it doesn't verify the response shape, the `wait=true` path setting `IsError: true` on failure, or the URL fallback construction. A few fake-binary tests similar to `TestGetRecentRuns` would cover the core branches.

integrations/rwx/docs.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	mcp "github.com/daltoniam/switchboard"
+)
+
+func docsSearch(_ context.Context, r *rwx, args map[string]any) (*mcp.ToolResult, error) {
```
acmacalister, line 11: Both `docsSearch` and `docsPull` are new with no tests. The snippet-truncation at 500 chars in `docsSearch` and the JSON parse-then-passthrough in `docsPull` are straightforward to exercise with a fake binary (same pattern as `TestRunRWXCommand_StdoutOnly`). Would be good to have at least a shape test for each.

integrations/rwx/vaults.go
```diff
@@ -0,0 +1,159 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	mcp "github.com/daltoniam/switchboard"
+)
+
+func vaultsVarShow(_ context.Context, r *rwx, args map[string]any) (*mcp.ToolResult, error) {
```
acmacalister, line 12: Five new vault handlers, no tests. The empty-output branches (e.g. the synthesized `{status, name, vault}` response on lines 53–58) are particularly worth pinning — if the CLI ever returns output on success, that branch silently disappears. A fake-binary test per handler covering both the happy path and the empty-output fallback would be straightforward to add.

integrations/rwx/logs.go
```diff
@@ -72,21 +76,25 @@ func downloadLogs(ctx context.Context, r *rwx, id string) (string, error) {
 		status, isComplete, err := fetchRunStatus(ctx, r, id)
 		_ = status
 		if err == nil && isComplete {
-			r.logCache.set(id, logs)
+			r.logCache.set(cacheKey, logs)
```
acmacalister, line 80: The goroutine that writes to the cache uses `ctx` from the caller, which can be request-scoped. If the parent context is cancelled before the goroutine runs (e.g. client disconnects mid-request), `fetchRunStatus` will return an error and the cache write is silently skipped — which is functionally fine, but it means a completed run won't get cached on that request. Using `context.WithoutCancel(ctx)` (Go 1.21+) here would decouple the cache write from the request lifecycle and make the caching more reliable.

acmacalister (commented):
Build, tests, and lint all pass — solid addition. Few thoughts on test coverage for the new handlers and one minor note on the goroutine context inline.

integrations/rwx/runs.go
```diff
@@ -92,6 +101,84 @@ func launchCIRun(_ context.Context, r *rwx, args map[string]any) (*mcp.ToolResul
...
+
+	if wait {
+		status := normalizeStatus(parsed.Result)
+		resp := map[string]any{
+			"completed": true,
+			"run_id":    parsed.RunID,
+			"status":    status,
+			"url":       runURL,
+		}
+		if status == "failure" {
```
acmacalister, line 169: The success path here doesn't include `next_step`, but `launchCIRun`'s equivalent block does (`"Run completed successfully"`). An LLM that dispatches with `wait=true` gets no guidance on what to do next when the run succeeds. Worth adding the same key here for consistency:

```go
if status == "failure" {
    resp["next_step"] = "Use rwx_get_run_results to see task failures, or rwx_grep_logs to search for errors"
} else {
    resp["next_step"] = "Run completed successfully"
}
```

integrations/rwx/vaults.go
```diff
@@ -0,0 +1,159 @@
...
+
+	cmdArgs := []string{"vaults", "vars", "show", name, "--vault", vault, "--output", "json"}
+	output, err := r.runRWXCommand(cmdArgs, 30000)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+
+	var parsed any
+	if err := json.Unmarshal([]byte(output), &parsed); err != nil {
+		return &mcp.ToolResult{Data: output}, nil
```
acmacalister, line 34: When the CLI outputs nothing (or non-JSON), `output` is an empty string and this returns `&mcp.ToolResult{Data: ""}` without `IsError: true`. The set/delete handlers avoid this by checking `output == ""` first and synthesizing a response. For `show` though, empty output most likely means something went wrong — might be cleaner to surface it as an error:

```go
if err := json.Unmarshal([]byte(output), &parsed); err != nil {
    if output == "" {
        return mcp.ErrResult(fmt.Errorf("vault var show: no output from CLI"))
    }
    return &mcp.ToolResult{Data: output}, nil
}
```

acmacalister (commented):
Good work addressing the previous round of feedback — all 5 prior comments are resolved. Two small things inline.

integrations/rwx/runs.go
```diff
@@ -92,6 +101,86 @@ func launchCIRun(_ context.Context, r *rwx, args map[string]any) (*mcp.ToolResul
...
```
acmacalister, line 169: The `wait=true` success path (`status != "failure"`) has no test coverage — only `TestDispatchRun_WaitFailure` exists. Worth adding a `TestDispatchRun_WaitSuccess` to confirm `IsError` stays false and `next_step` is set correctly:

```go
func TestDispatchRun_WaitSuccess(t *testing.T) {
	bin := fakeBin(t, `echo '{"run_id":"dispatch-ok","result":"succeeded"}'`)
	r := &rwx{cliPath: bin, org: "org", baseURL: "https://cloud.rwx.com", logCache: newLogCache()}

	result, err := dispatchRun(context.Background(), r, map[string]any{
		"dispatch_key": "deploy-staging",
		"wait":         true,
	})
	require.NoError(t, err)
	assert.False(t, result.IsError)

	var parsed map[string]any
	require.NoError(t, json.Unmarshal([]byte(result.Data), &parsed))
	assert.Equal(t, "success", parsed["status"])
	assert.Equal(t, "Run completed successfully", parsed["next_step"])
}
```

integrations/rwx/runs.go
```diff
@@ -245,68 +334,106 @@ func getRecentRuns(ctx context.Context, r *rwx, args map[string]any) (*mcp.ToolR
...
 		return mcp.ErrResult(err)
 	}
-	id := extractRunID(runIDRaw)
-	output, err := r.runRWXCommand([]string{"results", id, "--output", "json"}, 0)
+
+	var cmdArgs []string
+	if runIDRaw != "" {
+		id := extractRunID(runIDRaw)
+		cmdArgs = []string{"results", id, "--output", "json"}
+	} else if branch != "" || commit != "" {
```
acmacalister, line 359: The branch/commit lookup path in `getRunResults` is new and has no test coverage. The `run_id` path is tested indirectly via `TestParseResultsPrompt_*`, but nobody exercises the `else if branch != "" || commit != ""` branch or the `"either run_id or branch/commit is required"` error. A fake-binary test covering at least the branch lookup happy path and the missing-args error would pin the interface down.

acmacalister (commented):
Good follow-through on addressing all the previous round's feedback — the `context.WithoutCancel`, the `vaultsVarShow` empty-output error, the `next_step` parity in `dispatchRun`, and the new test suite are all solid. Two small test-coverage gaps inline.

integrations/rwx/runs.go
```diff
@@ -38,6 +38,15 @@ func launchCIRun(_ context.Context, r *rwx, args map[string]any) (*mcp.ToolResul
 		cmdArgs = append(cmdArgs, "--target", t)
 	}
 
+	if title != "" {
+		cmdArgs = append(cmdArgs, "--title", title)
+	}
+
+	initParams, _ := mcp.ArgMap(args, "init")
```
acmacalister: `mcp.ArgMap` returns an error when the value can't be cast to `map[string]any` (e.g. an LLM passes `"init": "staging"` instead of a JSON object). Ignoring it here means the init params are silently dropped with no feedback — the run just starts without them. Every other adapter that calls `ArgMap` handles the error. Worth propagating it:

```go
initParams, err := mcp.ArgMap(args, "init")
if err != nil {
    return mcp.ErrResult(err)
}
```

integrations/rwx/runs.go
```diff
@@ -92,6 +101,86 @@ func launchCIRun(_ context.Context, r *rwx, args map[string]any) (*mcp.ToolResul
...
+		cmdArgs = append(cmdArgs, "--ref", ref)
+	}
+	if wait {
+		cmdArgs = append(cmdArgs, "--wait", "--fail-fast")
+	}
+	if title != "" {
+		cmdArgs = append(cmdArgs, "--title", title)
+	}
+
+	params, _ := mcp.ArgMap(args, "params")
```
acmacalister: Same issue as the `init` params above — `mcp.ArgMap` error is silently discarded. If `params` is the wrong type, dispatch params are dropped without any indication. Same fix applies:

```go
params, err := mcp.ArgMap(args, "params")
if err != nil {
    return mcp.ErrResult(err)
}
```

integrations/rwx/docs.go
```diff
@@ -0,0 +1,89 @@
...
+			Title string `json:"title"`
+			Body  string `json:"body"`
+		} `json:"Results"`
+	}
+	if err := json.Unmarshal([]byte(output), &parsed); err != nil {
+		return mcp.ErrResult(fmt.Errorf("parse docs search: %w", err))
+	}
+
+	var results []map[string]any
+	for _, r := range parsed.Results {
```
acmacalister: The loop variable `r` shadows the outer function parameter `r *rwx`. It works because the loop var is a different type (a struct value, not a pointer), but it's easy to misread — a future edit that tries to call `r.runRWXCommand` inside the loop would silently use the wrong `r`. `result` or `doc` would be a cleaner name here.

acmacalister (commented):
Nice work on this PR — the breadth of new functionality is solid, the `parseResultsPrompt` logic is well thought-out, and the test coverage is thorough. Two small issues to address inline.

integrations/rwx/vaults.go
```diff
@@ -0,0 +1,162 @@
...
+	value := ra.Str("value")
+	vault := ra.Str("vault")
+	if err := ra.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	if vault == "" {
+		vault = "default"
+	}
+
+	cmdArgs := []string{"vaults", "secrets", "set", fmt.Sprintf("%s=%s", name, value), "--vault", vault, "--output", "json"}
```
acmacalister: The secret value is passed as a CLI argument here — on Linux, process arguments are visible in `/proc/<pid>/cmdline` and world-readable via `ps aux`, so another user on the same host could read the secret before the process exits.

If `rwx vaults secrets set` supports reading from stdin, we could pipe the `NAME=VALUE` line via `cmd.Stdin` in `runRWXCommand` instead of putting it in `args`. If the CLI doesn't support stdin yet, it's worth noting in the tool description that this has process-listing exposure, or filing a CLI feature request for a `--value-stdin` flag.

The variable equivalent (`vaultsVarSet` line 50) has the same pattern, but secrets are specifically supposed to be unreadable — so this is more of a concern there.

integrations/rwx/vaults.go
```diff
@@ -0,0 +1,162 @@
...
+	value := ra.Str("value")
+	vault := ra.Str("vault")
+	if err := ra.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	if vault == "" {
+		vault = "default"
+	}
+
+	cmdArgs := []string{"vaults", "vars", "set", fmt.Sprintf("%s=%s", name, value), "--vault", vault, "--output", "json"}
```
acmacalister: If `value` contains an `=` sign (e.g. a base64-encoded string like `abc=` or `abc==`), `fmt.Sprintf("%s=%s", name, value)` produces `name=abc=` — the CLI then has to split on the first `=` to get the key/value correctly. Worth confirming the RWX CLI handles this, or switching to separate `--name`/`--value` flags if those exist.

acmacalister (commented):
Nice work on this upgrade — the test coverage for the new handlers is solid, all the previously raised issues look addressed, and it builds and lints clean. Two small things inline on vaults.go.

integrations/rwx/runs.go
```diff
@@ -6,28 +6,28 @@ import (
...
 	"io"
 	"net/http"
+	"regexp"
 	"strings"
 	"time"
 
 	mcp "github.com/daltoniam/switchboard"
 )
 
 func launchCIRun(_ context.Context, r *rwx, args map[string]any) (*mcp.ToolResult, error) {
```
acmacalister, line 16: `launchCIRun` picked up three new behaviors in this PR — `--fail-fast` on the wait path, `--title`, and `--init` map expansion — but there are still no handler-level tests for this function at all. The dispatch parity test confirms routing is wired, but it doesn't cover the new code paths.

A few fake-binary tests similar to `TestDispatchRun_Launched` would pin the important cases: that `wait=true` produces `--wait --fail-fast` in the args, that `title` shows up when set, and that `init` params get expanded as repeated `--init key=value` args. The last one is worth testing because map iteration order is non-deterministic — if the CLI is sensitive to ordering, a test that inspects the args slice (rather than running the binary) would catch that before it becomes a flaky prod issue.

integrations/rwx/logs.go
```diff
@@ -140,12 +149,26 @@ func findLogFiles(dir string) ([]string, error) {
 // --- Log tool handlers ---
 
 func getTaskLogs(ctx context.Context, r *rwx, args map[string]any) (*mcp.ToolResult, error) {
```
acmacalister, line 151: `getTaskLogs` grew a new three-branch dispatch: `task_key+run_id`, `task_id`, or error. The error path (`"either task_id or run_id+task_key is required"`) and the `task_key+run_id` resolution path are both new and neither is tested — only the dispatch parity test touches this handler at all.

The `run_id+task_key` path is the interesting one for LLMs to use (they get task keys from `rwx_get_run_results`, not task IDs), so having a test that exercises it via a fake binary would also serve as documentation for future readers. Something like:

```go
func TestGetTaskLogs_RunIDAndTaskKey(t *testing.T) {
	bin := fakeBin(t, `echo 'log output for task'`)
	r := &rwx{cliPath: bin, logCache: newLogCache()}
	result, err := getTaskLogs(context.Background(), r, map[string]any{
		"run_id":   "run-abc123",
		"task_key": "ci.checks.lint",
	})
	require.NoError(t, err)
	assert.False(t, result.IsError)
}
```

acmacalister (commented):
Solid PR — the `parseResultsPrompt` rewrite is well-tested, the security fixes in vaults (env file instead of CLI args) are a real improvement, and the `context.WithoutCancel` fix for the log cache goroutine is exactly right. Two coverage gaps in `launchCIRun` and `getTaskLogs` inline.

integrations/rwx/runs.go
```diff
@@ -316,6 +449,90 @@ func getRunResults(_ context.Context, r *rwx, args map[string]any) (*mcp.ToolRes
...
+}
+
+func parseResultsPrompt(prompt string) (tasks []failedTask, failedTests []string, otherProblems []string) {
+	if prompt == "" {
+		return nil, nil, nil
+	}
+
+	lines := strings.Split(prompt, "\n")
+	var section string
+	taskRe := regexp.MustCompile(`^- (.+?) \(task-id: ([a-f0-9]+)\)(.*)`)
```
acmacalister, line 490: `taskRe` is compiled fresh on every call to `parseResultsPrompt`, which runs on every `rwx_get_run_results` invocation. Since the pattern never changes, hoisting it to a package-level var avoids the allocation on the hot path — same pattern as `cliToMCPReplacements` in `proxy.go`:

```go
var taskRe = regexp.MustCompile(`^- (.+?) \(task-id: ([a-f0-9]+)\)(.*)`)
```

acmacalister (approved):
Solid work across the board — all six prior rounds of feedback are fully addressed, the new handlers are well-tested, and the security fix for vault values (env file instead of CLI args) is the right call. One small nit inline.

PR #153, opened 2026-05-01
feat(wasm): add optional raw bytes support for plugin HTTP bodies

aleksclark:
## Summary

- Adds `body_base64` field to the WASM plugin HTTP ABI for binary payloads (protobuf, images, gRPC, etc.)
- Plugins opt in per-request via `X-Raw-Body` header (same pattern as existing `X-H2C`); response body returned as base64 in `body_base64` instead of UTF-8 string in `body`
- Request bodies can also be sent as base64 via the `body_base64` field on `HttpRequest`
- Fully backward-compatible — existing plugins that only use the `body` string field are unaffected

## Test plan

- [x] `make ci` passes
- [x] Binary response round-trip (bytes → base64 → decode)
- [x] `X-Raw-Body` header not leaked to upstream
- [x] Normal requests still return string body (no base64)
- [x] Binary request body via `body_base64`
- [x] Invalid base64 error handling

🐾 Generated with Crush

### DIFF
--- a/wasm/guest-rust/sdk/src/lib.rs
+++ b/wasm/guest-rust/sdk/src/lib.rs
@@ -1,5 +1,6 @@
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
+use base64::Engine as _;
 
 // ── Types matching the host ABI ─────────────────────────────────────────────
 
@@ -88,6 +89,8 @@ pub struct HttpRequest {
     pub headers: HashMap<String, String>,
     #[serde(skip_serializing_if = "String::is_empty")]
     pub body: String,
+    #[serde(skip_serializing_if = "String::is_empty")]
+    pub body_base64: String,
 }
 
 #[derive(Deserialize)]
@@ -97,6 +100,8 @@ pub struct HttpResponse {
     pub headers: HashMap<String, String>,
     #[serde(default)]
     pub body: String,
+    #[serde(default)]
+    pub body_base64: String,
 }
 
 // ── Host imports ────────────────────────────────────────────────────────────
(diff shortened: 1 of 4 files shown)


acmacalister (commented):
Go side is clean — builds, tests pass, lint is green. One issue to fix in the Rust SDK: a name collision that prevents `cargo build` from succeeding.

wasm/guest-rust/sdk/src/lib.rs
```diff
@@ -136,11 +170,29 @@ pub fn host_http_request_h2c(req: &HttpRequest) -> Result<HttpResponse, String>
...
     patched.headers.insert("X-H2C".into(), "1".into());
     do_host_http_request(&patched)
 }
 
+/// Execute an HTTP request and receive the response body as raw bytes.
+///
+/// Sets the `X-Raw-Body` header so the host returns the body as base64
+/// in `body_base64` instead of a UTF-8 string in `body`. Use
+/// [`HttpResponse::body_bytes`] to decode the result.
+pub fn host_http_request_raw(req: &HttpRequest) -> Result<HttpResponse, String> {
```
acmacalister, line 184: The `extern "C"` block at line 111 already defines an internal symbol named `host_http_request_raw` (as an alias for the `host_http_request` ABI entry), and this new public function uses the same name — Rust sees them as a collision in the same value namespace.

This causes a hard compile error:
```
error[E0428]: the name `host_http_request_raw` is defined multiple times
```

Renaming the public function avoids the conflict and makes the intent clearer anyway:
```rust
pub fn host_http_request_raw_body(req: &HttpRequest) -> Result<HttpResponse, String> {
```
or if "raw body" is the preferred phrasing:
```rust
pub fn host_http_request_raw_response(req: &HttpRequest) -> Result<HttpResponse, String> {
```

PR #154, opened 2026-05-01
ci: add Rust SDK compile check + fix naming collision

aleksclark:
## Summary

- **Add `rust-sdk` CI job** that runs `cargo check --target wasm32-unknown-unknown` on the guest Rust SDK — catches compile errors before merge
- **Fix naming collision**: `host_http_request_raw` was defined twice — once as an FFI extern and once as a public function. Renamed the public function to `host_http_request_raw_body`

## Test plan

- [x] First commit adds CI job only — `rust-sdk` job should **fail** (proves the check works)
- [ ] Second commit fixes the bug — `rust-sdk` job should **pass**
- [x] `make ci` passes locally

🐾 Generated with Crush


wasm/guest-rust/sdk/src/lib.rs
```diff
@@ -306,20 +306,20 @@ fn unpack_ptr_size(v: u64) -> (u32, u32) {
...
 pub extern "C" fn guest_free(ptr: u32) {
     if ptr == 0 {
         return;
     }
-    unsafe { free(ptr as *mut u8) }
+    // We don't track allocation sizes, so we deallocate with size=1 alignment=1.
+    // This is safe because wasm linear memory isn't reclaimed per-allocation;
+    // the host only calls guest_free to signal the guest can reuse the memory.
+    let layout = std::alloc::Layout::from_size_align(1, 1).unwrap();
+    unsafe { std::alloc::dealloc(ptr as *mut u8, layout) }
```
acmacalister, line 324: The `dealloc` call here passes `Layout::from_size_align(1, 1)` but the corresponding `alloc` in `guest_malloc` uses `Layout::from_size_align(size, 1)` — those layouts don't match. Per Rust's `GlobalAlloc` safety contract, `dealloc` must be called with the **exact same layout** as the original `alloc`, otherwise behavior is undefined.

The comment's reasoning holds for dlmalloc (the default WASM allocator, which stores size metadata internally and ignores the layout passed to `dealloc`), but it's a hidden footgun: if anyone ever swaps in a different global allocator (e.g. `wee_alloc`, `lol_alloc`, or `talc`), this will silently corrupt or leak memory.

A simpler fix that stays correct is to just not call `dealloc` at all — since the comment already acknowledges WASM linear memory doesn't reclaim per-allocation, a no-op free is both correct and honest about the intent:

```rust
#[no_mangle]
pub extern "C" fn guest_free(_ptr: u32) {
    // WASM linear memory is not reclaimed per-allocation;
    // the host calls this only to signal it's done with the region.
}
```

Alternatively, track the allocation size at call time if you want a proper free.

.github/workflows/ci.yml
```diff
@@ -75,3 +75,16 @@ jobs:
 
       - name: Run govulncheck
         run: go tool govulncheck ./...
+
+  rust-sdk:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: dtolnay/rust-toolchain@stable
```
acmacalister, line 84: `dtolnay/rust-toolchain@stable` is a floating tag — both the action code and the Rust toolchain version can change beneath you. The `@stable` here is a branch/tag on the action repo, not a pinned commit SHA.

For supply-chain safety it's worth pinning to a commit SHA the same way you'd pin `actions/checkout`. You can find the current SHA at https://github.com/dtolnay/rust-toolchain — something like:

```yaml
- uses: dtolnay/rust-toolchain@b3b07ba8b418998c39fb20d7348b0e98e4f04b90  # stable as of 2024-xx-xx
  with:
    targets: wasm32-unknown-unknown
```

Minor note since the rest of the CI file also uses version tags (`@v4`, `@v5`) rather than SHAs — this is consistent with the existing project pattern, so totally fine to leave for a follow-up.

acmacalister (commented):
Nice fix for the FFI collision — the rename from `host_http_request_raw` to `host_http_request_raw_body` is clear and the CI job is a great addition. Two things worth a look inline.

PR #156, opened 2026-05-07
feat(pganalyze): migrate to official MCP server proxy

aleksclark:
## Summary

- Replaces the custom GraphQL API client (3 tools) with an HTTP-based MCP proxy to pganalyze's official MCP server (`https://app.pganalyze.com/mcp`)
- Expands from 3 tools to 20 tools: adds tables, EXPLAIN plans, backends, index advisor, postgres settings, query details/samples, and more
- Auth simplified to Bearer token (API key) — `organization_slug` credential removed

## New Tools (17 added)

| Category | Tools |
|----------|-------|
| Servers | `get_server_details`, `get_postgres_settings` |
| Databases | `get_databases` |
| Queries | `get_query_details`, `get_query_samples` |
| Tables | `get_tables`, `get_table`, `get_table_stats`, `get_index_selection`, `run_index_selection` |
| EXPLAIN Plans | `get_query_explains`, `get_query_explain`, `get_query_explain_from_trace` |
| Backends | `get_backend_counts`, `get_backends`, `get_backend_details` |
| Issues | `get_checkup_status` |

## Test plan

- [x] All 27 pganalyze unit tests pass (proxy lifecycle, tool execution, error handling, rate limiting)
- [x] Full `make ci` passes (build, vet, test-race, lint, security)
- [x] Search benchmark still indexes pganalyze tools correctly (static fallback)
- [x] Config tests updated for removed `organization_slug`

🐘 Generated with Crush


integrations/pganalyze/pganalyze.go
```diff
@@ -38,123 +35,57 @@ func (p *pganalyze) Configure(_ context.Context, creds mcp.Credentials) error {
...
 	return nil
 }
 
 func (p *pganalyze) Healthy(ctx context.Context) bool {
 	if p.apiKey == "" {
 		return false
 	}
-	_, err := p.gql(ctx, `{ __typename }`, nil)
-	return err == nil
+	if !p.started {
```
acmacalister, line 54: `started` and `proxy` are read here and in `Execute` while `startProxy` writes them with no synchronization — I confirmed this triggers the race detector. Two goroutines hitting `Healthy` and `Execute` concurrently (normal for an MCP server handling parallel tool calls) will race on both fields.

`sync.Once` is the idiomatic fix:

```go
type pganalyze struct {
    apiKey  string
    mcpURL  string
    client  *http.Client
    once    sync.Once
    proxy   *proxyClient  // written once via once.Do, safe to read after
}

func (p *pganalyze) ensureProxy(ctx context.Context) {
    p.once.Do(func() {
        proxy := newProxyClient(p.mcpURL, p.apiKey, p.client)
        if err := proxy.initialize(ctx); err != nil {
            return
        }
        if err := proxy.fetchTools(ctx); err != nil {
            return
        }
        p.proxy = proxy
    })
}
```

Then `Healthy` and `Execute` both call `p.ensureProxy(ctx)` instead of checking `p.started`. The `Once` guarantees `proxy` is fully written before any reader sees it.

integrations/pganalyze/proxy.go
```diff
@@ -0,0 +1,242 @@
...
+	httpReq.Header.Set("Accept", "application/json")
+	httpReq.Header.Set("Authorization", "Bearer "+p.apiKey)
+
+	resp, err := p.client.Do(httpReq)
+	if err != nil {
+		return nil, err
+	}
+	defer func() { _ = resp.Body.Close() }()
+
+	respBody, err := io.ReadAll(resp.Body)
```
acmacalister, line 124: Every other HTTP adapter in the codebase caps response reads with `io.LimitReader` (snowflake, nomad, jira, fly, etc. all use a `maxResponseSize` constant). This one doesn't, so a large or malicious response from the MCP endpoint — especially relevant since `base_url` is user-configurable — could cause unbounded memory growth.

```go

// in send():
respBody, err := io.ReadAll(io.LimitReader(resp.Body, maxResponseSize))
```

acmacalister (commented):
Nice refactor — the proxy architecture is much cleaner than the bespoke GraphQL client, and the test coverage is thorough. Two things worth fixing before merge: a confirmed data race on lazy proxy init, and an unbounded response read.

PR #159, opened 2026-05-09
Fix postgres web UI: JSON tag round-trip and slow setup page load

daltoniam:
## Summary

- Add snake_case JSON tags to `pages.PostgresConnection` so additional connections round-trip cleanly through the web UI. Addresses the latest review comment on #109 — without tags, snake_case payload fields like `connection_string` and `read_only` did not bind on reload, came back blank, and were effectively wiped on the next save (the JS gather skips empty inputs, so connections appeared to vanish after the second save).
- Stop calling `integration.Configure()` on every GET to `/integrations/postgres/setup`. With multi-postgres each call re-opens every configured database and pings them with a 5s timeout, which makes the setup page hang noticeably as more connections are added. The integration is already configured at startup; the page now only does a short-timeout `Healthy()` probe.
- Regenerate `postgres_setup_templ.go`.

## Changes

| Area | File | What |
|------|------|------|
| Web UI struct | `web/templates/pages/postgres_setup.templ` | Add `json:\"...\"` tags to all `PostgresConnection` fields |
| Generated | `web/templates/pages/postgres_setup_templ.go` | `templ generate` output |
| Web handler | `web/web.go` | Drop `Configure()` from `handlePostgresSetup`; bound `Healthy()` with a 1s timeout |
| Tests | `web/web_test.go` | `TestPostgresConnection_JSONRoundTrip`, `TestPostgresConnection_DecodesSnakeCase` |

## Test plan

- [x] `make ci` passes (build, vet, race tests, lint, gosec, govulncheck)
- [x] New unit tests cover the JSON round-trip and explicit snake-case decode case
- [ ] Manual: open the postgres setup page with 2+ additional connections, save, reload, and confirm fields are populated and the page loads instantly

🤖 Generated with Crush

### DIFF
--- a/web/templates/pages/postgres_setup.templ
+++ b/web/templates/pages/postgres_setup.templ
@@ -7,15 +7,15 @@ import (
 )
 
 type PostgresConnection struct {
-	Alias            string
-	ConnectionString string
-	Host             string
-	Port             string
-	User             string
-	Password         string
-	Database         string
-	SSLMode          string
-	ReadOnly         string
+	Alias            string `json:"alias"`
+	ConnectionString string `json:"connection_string"`
+	Host             string `json:"host"`
+	Port             string `json:"port"`
+	User             string `json:"user"`
+	Password         string `json:"password"`
+	Database         string `json:"database"`
+	SSLMode          string `json:"sslmode"`
+	ReadOnly         string `json:"read_only"`
 }
 
 type PostgresSetupData struct {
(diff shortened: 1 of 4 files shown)


acmacalister (approved):
Clean PR — builds, tests pass (race-clean), lint clean, go mod tidy. The JSON tag additions are the right fix and the two new tests cover both the round-trip contract and the exact snake_case failure mode. LGTM.

PR #160, opened 2026-05-09
Remove WASM Modules page, unify into Plugin Marketplace

daltoniam:
## Summary

- Remove the standalone WASM Modules page (`/wasm`) — the Plugin Marketplace now handles all plugin management (install, upload, load, update, uninstall)
- Credentials for WASM plugins are self-describing via the module's `metadata()` export and configurable on the Integrations page — no manual key naming needed
- Add "Load from Local Path" section to marketplace for dev workflow (load local `.wasm` files with live-reload, no restart)

## Changes

- Remove `/wasm` nav item, routes, 5 handlers, 2 templates, and ~300 lines of tests
- Add `POST /plugins/load-path` handler for local path loading
- Add "Load from Local Path" form to marketplace UI
- Net: **-1013 / +72 lines**

## Test plan

- [x] `go test ./...` passes
- [x] `go vet ./...` clean
- [x] New handler tests added (`TestPluginLoadPath`, `TestPluginLoadPath_EmptyPath`)
- [ ] Manual: verify marketplace page renders correctly
- [ ] Manual: verify "Load from Local Path" loads a plugin and it appears on Integrations page

🐨 Generated with Crush


web/web_marketplace.go
```diff
@@ -331,6 +331,23 @@ func (w *WebServer) liveUnloadPlugin(ctx context.Context, name string) {
...
+
+	path := strings.TrimSpace(r.FormValue("path"))
+	if path == "" {
+		http.Redirect(rw, r, "/plugins?error=Path+is+required", http.StatusSeeOther)
+		return
+	}
+
+	name := strings.TrimSpace(r.FormValue("name"))
+	w.liveLoadPlugin(r.Context(), path, name)
+	http.Redirect(rw, r, "/plugins?success=Plugin+loaded+from+local+path.", http.StatusSeeOther)
```
acmacalister, line 361: This always redirects with a success flash even when `liveLoadPlugin` fails. Unlike the install/upload/update handlers where `liveLoadPlugin` is best-effort (the binary was already validated by the marketplace), here *loading is the entire operation* — if the file doesn't exist or isn't valid WASM, the user still sees "Plugin loaded from local path."

`liveLoadPlugin` currently only logs the error. For this handler, we'd want to surface it:

```go
if err := w.liveLoadPlugin(r.Context(), path, name); err != nil {
    http.Redirect(rw, r, "/plugins?error=Load+failed:+"+urlEncode(err.Error()), http.StatusSeeOther)
    return
}
http.Redirect(rw, r, "/plugins?success=Plugin+loaded+from+local+path.", http.StatusSeeOther)
```

That would require `liveLoadPlugin` to return its error rather than swallowing it, but since the callers that need best-effort behavior can just ignore the return value, it's a backward-compatible change.

web/templates/layouts/base_templ.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by templ - DO NOT EDIT.
 
-// templ: version: v0.3.1001
+// templ: version: v0.3.977
 package layouts
```
acmacalister, line 4: All the generated `*_templ.go` files were regenerated with v0.3.977 here, but `go.mod` requires `github.com/a-h/templ v0.3.1001`. These should stay in sync — regenerate with the version pinned in `go.mod` (`go tool templ generate`) to avoid a version skew that could surface as subtle rendering differences.

acmacalister (commented):
Solid cleanup — removing the standalone WASM page and folding it into the marketplace is the right call. Two things worth looking at inline.

web/web_test.go
```diff
@@ -317,321 +317,33 @@ func TestMetricsAPI(t *testing.T) {
...
-	req := httptest.NewRequest("POST", "/wasm/delete", form)
+	form := strings.NewReader("path=/tmp/dev-plugin.wasm&name=dev-plugin")
+	req := httptest.NewRequest("POST", "/plugins/load-path", form)
 	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
 	rr := httptest.NewRecorder()
 	handler.ServeHTTP(rr, req)
 
 	assert.Equal(t, http.StatusSeeOther, rr.Code)
+	assert.Contains(t, rr.Header().Get("Location"), "/plugins")
 	assert.Contains(t, rr.Header().Get("Location"), "success")
```
acmacalister: `setupTestWeb` passes `nil` for `wasmLoader`, so `liveLoadPlugin` returns `nil` immediately via the nil guard — `TestPluginLoadPath` never exercises the new error-surfacing path added in the previous commit. The test asserts success, but it's really testing "loader not configured" rather than "plugin loaded successfully".

Worth adding a test that wires up a mock `wasmLoader` (or at least a mock that returns an error) to verify the error redirect fires:

```go
func TestPluginLoadPath_LoadError(t *testing.T) {
    ws, _, _ := setupTestWeb()
    ws.wasmLoader = &mockWasmLoader{err: errors.New("not a valid WASM file")}
    handler := ws.Handler()

    form := strings.NewReader("path=/tmp/bad.wasm")
    req := httptest.NewRequest("POST", "/plugins/load-path", form)
    req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
    rr := httptest.NewRecorder()
    handler.ServeHTTP(rr, req)

    assert.Equal(t, http.StatusSeeOther, rr.Code)
    assert.Contains(t, rr.Header().Get("Location"), "error")
}
```

acmacalister (commented):
Clean PR overall — the WASM Modules page removal is tidy and the liveLoadPlugin error-surfacing fix is the right approach. Two things inline worth a look.

web/web_marketplace.go
```diff
@@ -331,6 +333,26 @@ func (w *WebServer) liveUnloadPlugin(ctx context.Context, name string) {
 	}
 }
 
+func (w *WebServer) handlePluginLoadPath(rw http.ResponseWriter, r *http.Request) {
+	if err := r.ParseForm(); err != nil {
+		http.Redirect(rw, r, "/plugins?error=Invalid+form+data", http.StatusSeeOther)
+		return
+	}
+
+	path := strings.TrimSpace(r.FormValue("path"))
```
acmacalister, line 342: `path` is submitted directly from the browser form and passed to `os.ReadFile` with no validation. A user could submit `/etc/passwd` or `../../some-secret` — the WASM runtime would reject it as invalid WASM, but the error message ("not a valid WASM magic number") confirms the file exists and was readable. For a local-only dev tool this is lower severity, but worth at least validating the extension or confirming the path is absolute:

```go
if !strings.HasSuffix(path, ".wasm") {
    http.Redirect(rw, r, "/plugins?error=Path+must+end+in+.wasm", http.StatusSeeOther)
    return
}
```

That won't stop a determined user from renaming a file, but it prevents accidental (or trivially malicious) reads of arbitrary filesystem paths.

acmacalister (approved):
Build passes, full test suite clean with race detector, lint 0 issues. The private-repo auth feature is well-designed — functional options is the right pattern here, `GitHubTokenFunc` is tested thoroughly, and reading the token dynamically from the config store means it picks up a newly-configured credential without a restart. Two earlier comments on the load-path handler (error test coverage and path validation) are still worth addressing, but nothing new blocking merge.

web/web_marketplace.go
```diff
@@ -331,6 +333,30 @@ func (w *WebServer) liveUnloadPlugin(ctx context.Context, name string) {
...
+		http.Redirect(rw, r, "/plugins?error=Path+must+end+in+.wasm", http.StatusSeeOther)
+		return
+	}
+
+	name := strings.TrimSpace(r.FormValue("name"))
+	if err := w.liveLoadPlugin(r.Context(), path, name); err != nil {
+		http.Redirect(rw, r, "/plugins?error=Load+failed:+"+urlEncode(err.Error()), http.StatusSeeOther)
+		return
+	}
+	http.Redirect(rw, r, "/plugins?success=Plugin+loaded+from+local+path.", http.StatusSeeOther)
```
acmacalister, line 361: When `w.wasmLoader` is nil, `liveLoadPlugin` returns `nil` immediately and the user sees "Plugin loaded from local path." — even though nothing happened. The other load-path-adjacent handlers (install, upload, update) are intentionally best-effort since the binary was already validated and written to disk before `liveLoadPlugin` is called. This handler is different: loading *is* the entire operation, so a nil loader should be surfaced rather than silently succeeding.

```go
if w.wasmLoader == nil {
    http.Redirect(rw, r, "/plugins?error=WASM+loader+not+configured", http.StatusSeeOther)
    return
}
```

web/web_test.go
```diff
@@ -317,321 +318,64 @@ func TestMetricsAPI(t *testing.T) {
 	assert.Contains(t, body, "integrations")
 }
 
-func TestWasmModulesPage_Empty(t *testing.T) {
+func TestPluginLoadPath(t *testing.T) {
 	ws, _, _ := setupTestWeb()
```
acmacalister: `setupTestWeb()` passes `nil` for `wasmLoader`, so this test reaches the nil-guard in `liveLoadPlugin` and returns `nil` before any file I/O. The "success" assertion reflects "loader not configured" rather than "plugin loaded successfully". `TestPluginLoadPath_LoadError` covers the real path with an actual loader, which is great — worth adding a parallel success case there (e.g., a small valid WASM file or a mock `WasmLoader` interface) so the success redirect is tested with the loader actually engaged.

acmacalister (commented):
Nice cleanup — removing ~1k lines of the old WASM page while keeping the workflow intact is a solid simplification. The private-repo auth is well-structured and thoroughly tested. Two small things inline.

acmacalister (commented):
The previous round of feedback has all been addressed — error surfacing, nil loader guard, `.wasm` validation, and templ version sync are all clean. One gap remains inline.

web/web_test.go
```diff
@@ -317,62 +318,28 @@ func TestMetricsAPI(t *testing.T) {
...
 
 	assert.Equal(t, http.StatusSeeOther, rr.Code)
-	assert.Contains(t, rr.Header().Get("Location"), "/wasm")
-	assert.Contains(t, rr.Header().Get("Location"), "success")
-
-	require.Len(t, cfgService.cfg.WasmModules, 1)
-	assert.Equal(t, "/tmp/new.wasm", cfgService.cfg.WasmModules[0].Path)
-	assert.Equal(t, "secret", cfgService.cfg.WasmModules[0].Credentials["api_key"])
+	assert.Contains(t, rr.Header().Get("Location"), "error")
+	assert.Contains(t, rr.Header().Get("Location"), "Load+failed")
```
acmacalister, line 334: This test covers the error path well, but there's still no test that exercises the success redirect — the case where `wasmLoader` is set and `LoadPlugin` returns `nil`. Since `wasmLoader` is a concrete `*wasmmod.Loader`, the easiest path is to introduce a small `pluginLoader` interface in `web.go` (it only needs `LoadPlugin` and `UnloadPlugin`), then test with a mock that returns `nil`. That would let us assert the `?success=Plugin+loaded+from+local+path.` redirect fires correctly:

```go
type pluginLoader interface {
    LoadPlugin(ctx context.Context, path, name string) error
    UnloadPlugin(ctx context.Context, name string) error
}
```

Alternatively, `wasm/testdata/example.wasm` could be used directly in a test that wires up a real `Runtime`. Either way, the success branch of `handlePluginLoadPath` currently has no coverage.

PR #161, opened 2026-05-12
fix(rwx): inject RWX_ACCESS_TOKEN into CLI subprocess env

lukecurri:
The plugin holds the access token in r.accessToken from Configure() but never propagates it to the rwx CLI subprocesses. CLI-backed tools (logs, results, artifacts, mcp serve) fail with "no access token configured" even when the UI shows CONNECTED.

The rwx CLI v3.13+ authenticates via RWX_ACCESS_TOKEN, --access-token, or a credentials file from rwx login. Users who paste the token into switchboard's UI without running rwx login have no auth path.

Fix: new r.cmdEnv() appends RWX_ACCESS_TOKEN=<token> to subprocess env. Applied in runRWXCommand and proxyClient.start.


daltoniam (approved):
Clean PR — builds, tests pass, lint clean. Nice fix for the subprocess auth gap; cmdEnv() is well-documented and the test coverage is solid. LGTM.

PR #162, opened 2026-05-14
Add Microsoft Teams integration

daltoniam:
## Summary

- Adds a full Microsoft Teams adapter (`integrations/teams/`) implementing the `mcp.Integration` interface against the Microsoft Graph API.
- 26 tools spanning auth/tenant management, chats, joined teams + channels, users, and presence — multi-tenant token store with proactive + 401-triggered refresh.
- Uses Microsoft device-code OAuth against the Azure CLI public client (no client secret required for personal/dev use); persists tokens to `~/.teams-mcp-tokens.json`.

## Test plan

- [x] `go test ./integrations/teams/...` passes (29.9% coverage, 30+ unit tests)
- [x] `go tool golangci-lint run ./integrations/teams/...` clean
- [x] `TestDispatchMap_AllToolsCovered` + `TestDispatchMap_NoOrphanHandlers` parity tests
- [x] `TestFieldCompactionSpecs_*` parity tests
- [x] `TestNewArgs_ErrCheckParity` covers all new handlers automatically
- [x] `make ci` passes for all packages except `browser.TestPage_Evaluate` (pre-existing failure on `main`, unrelated to this change)
- [ ] Manual: `teams_login` → device-code flow → `teams_list_chats` end-to-end against a real tenant

💘 Generated with Crush

### DIFF
--- a/integrations/teams/auth.go
+++ b/integrations/teams/auth.go
@@ -1,0 +1,257 @@
+package teams
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"math"
+	"strings"
+	"time"
+
+	mcp "github.com/daltoniam/switchboard"
+)
+
+// teamsLogin starts (or returns the in-progress) device-code OAuth flow.
+func teamsLogin(ctx context.Context, t *teamsIntegration, args map[string]any) (*mcp.ToolResult, error) {
+	r := mcp.NewArgs(args)
+	tenantHint := r.Str("tenant")
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+
+	// If a flow is already in progress and not yet done, return the same prompt
+	// so the human can keep using the existing user_code.
+	activeFlow.mu.Lock()
+	existing := activeFlow.flow
+	activeFlow.mu.Unlock()
+	if existing != nil {
+		existing.mu.Lock()
+		stillRunning := !existing.done && time.Since(existing.startedAt) < time.Duration(existing.deviceResp.ExpiresIn)*time.Second
+		var dcr *deviceCodeResponse
+		if stillRunning {
+			dcr = existing.deviceResp
+		}
+		existing.mu.Unlock()
+		if dcr != nil {
+			return loginResponse(dcr, "resumed"), nil
+		}
+	}
+
+	dcr, err := t.startOAuth(ctx, tenantHint)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	return loginResponse(dcr, "started"), nil
+}
+
+func loginResponse(dcr *deviceCodeResponse, status string) *mcp.ToolResult {
+	payload := map[string]any{
+		"status":           status,
+		"user_code":        dcr.UserCode,
+		"verification_uri": dcr.VerificationURI,
+		"expires_in":       dcr.ExpiresIn,
+		"interval_seconds": dcr.Interval,
+		"message":          dcr.Message,
+		"next_step":        "Visit verification_uri in a browser, enter user_code, then call teams_login_poll to detect completion.",
+	}
+	data, _ := json.Marshal(payload)
+	return &mcp.ToolResult{Data: string(data)}
+}
+
+// teamsLoginPoll reports the state of the active device-code flow.
+func teamsLoginPoll(_ context.Context, _ *teamsIntegration, _ map[string]any) (*mcp.ToolResult, error) {
+	res := pollOAuth()
+	return mcp.JSONResult(res)
+}
+
+// teamsTokenStatus reports per-tenant token health.
+func teamsTokenStatus(_ context.Context, t *teamsIntegration, _ map[string]any) (*mcp.ToolResult, error) {
+	type entry struct {
+		TenantID    string  `json:"tenant_id"`
+		TenantName  string  `json:"tenant_name,omitempty"`
+		UserUPN     string  `json:"user_upn,omitempty"`
+		UserDisplay string  `json:"user_display,omitempty"`
+		Status      string  `json:"status"`
+		HasRefresh  bool    `json:"has_refresh"`
+		AgeMinutes  float64 `json:"age_minutes"`
+		ExpiresIn   string  `json:"expires_in,omitempty"`
+		Source      string  `json:"source,omitempty"`
+		IsDefault   bool    `json:"is_default"`
+	}
+	defaultID := t.store.defaultID()
+	var entries []entry
+	for _, tn := range t.store.all() {
+		age := 0.0
+		if !tn.UpdatedAt.IsZero() {
+			age = math.Round(time.Since(tn.UpdatedAt).Minutes()*10) / 10
+		}
+		status := "healthy"
+		expiresIn := ""
+		if !tn.ExpiresAt.IsZero() {
+			remaining := time.Until(tn.ExpiresAt)
+			if remaining <= 0 {
+				status = "expired"
+				expiresIn = "0s"
+			} else {
+				if remaining < 5*time.Minute {
+					status = "expiring_soon"
+				}
+				expiresIn = remaining.Round(time.Second).String()
+			}
+		}
+		if tn.AccessToken == "" {
+			status = "missing"
+		}
+		entries = append(entries, entry{
+			TenantID:    tn.TenantID,
+			TenantName:  tn.TenantName,
+			UserUPN:     tn.UserUPN,
+			UserDisplay: tn.UserDisplay,
+			Status:      status,
+			HasRefresh:  tn.RefreshToken != "",
+			AgeMinutes:  age,
+			ExpiresIn:   expiresIn,
+			Source:      tn.Source,
+			IsDefault:   tn.TenantID == defaultID,
+		})
+	}
+	return mcp.JSONResult(map[string]any{
+		"tenant_count":      len(entries),
+		"default_tenant_id": defaultID,
+		"tenants":           entries,
+		"auto_refresh": map[string]any{
+			"enabled":          true,
+			"interval_minutes": 15,
+			"skew_seconds":     int(refreshSkew.Seconds()),
+		},
+	})
+}
+
+// teamsRefreshTokens forces a refresh for the (default or specified) tenant.
+func teamsRefreshTokens(ctx context.Context, t *teamsIntegration, args map[string]any) (*mcp.ToolResult, error) {
+	r := mcp.NewArgs(args)
+	tid := r.Str("tenant_id")
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	var tn *tenant
+	if tid != "" {
+		tn = t.store.get(tid)
+		if tn == nil {
+			return mcp.ErrResult(fmt.Errorf("unknown tenant: %s", tid))
+		}
+	} else {
+		tn = t.activeTenant()
+		if tn == nil {
+			return mcp.ErrResult(fmt.Errorf("no tenant configured — run teams_login first"))
+		}
+	}
+	if tn.RefreshToken == "" {
+		return &mcp.ToolResult{
+			Data:    fmt.Sprintf("tenant %s has no refresh_token; re-run teams_login to acquire one", tn.TenantID),
+			IsError: true,
+		}, nil
+	}
+	if err := t.refreshTenant(ctx, tn); err != nil {
+		return mcp.ErrResult(err)
+	}
+	refreshed := t.store.get(tn.TenantID)
+	return mcp.JSONResult(map[string]any{
+		"status":     "refreshed",
+		"tenant_id":  refreshed.TenantID,
+		"expires_at": refreshed.ExpiresAt.UTC().Format(time.RFC3339),
+	})
+}
+
+// teamsListTenants returns all configured tenants for the user.
+func teamsListTenants(_ context.Context, t *teamsIntegration, _ map[string]any) (*mcp.ToolResult, error) {
+	type info struct {
+		TenantID    string `json:"tenant_id"`
+		TenantName  string `json:"tenant_name,omitempty"`
+		UserUPN     string `json:"user_upn,omitempty"`
+		UserDisplay string `json:"user_display,omitempty"`
+		IsDefault   bool   `json:"is_default"`
+	}
+	defaultID := t.store.defaultID()
+	all := t.store.all()
+	out := make([]info, 0, len(all))
+	for _, tn := range all {
+		out = append(out, info{
+			TenantID:    tn.TenantID,
+			TenantName:  tn.TenantName,
+			UserUPN:     tn.UserUPN,
+			UserDisplay: tn.UserDisplay,
+			IsDefault:   tn.TenantID == defaultID,
+		})
+	}
+	return mcp.JSONResult(map[string]any{
+		"count":             len(out),
+		"default_tenant_id": defaultID,
+		"tenants":           out,
+	})
+}
+
+// teamsRemoveTenant forgets cached tokens for a tenant.
+func teamsRemoveTenant(_ context.Context, t *teamsIntegration, args map[string]any) (*mcp.ToolResult, error) {
+	r := mcp.NewArgs(args)
+	tid := r.Str("tenant_id")
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	if tid == "" {
+		return mcp.ErrResult(fmt.Errorf("tenant_id is required"))
+	}
+	if t.store.get(tid) == nil {
+		return mcp.ErrResult(fmt.Errorf("unknown tenant: %s", tid))
+	}
+	t.store.remove(tid)
+	_ = t.store.saveToFile()
+	return mcp.JSONResult(map[string]any{
+		"status":            "removed",
+		"tenant_id":         tid,
+		"default_tenant_id": t.store.defaultID(),
+	})
+}
+
+// teamsSetDefault pins the default tenant for tools called without tenant_id.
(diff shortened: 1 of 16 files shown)


integrations/teams/oauth.go
```diff
@@ -0,0 +1,409 @@
...
+// startOAuth kicks off a device-code flow. The returned response includes the
+// user_code + verification_uri the human needs.
+func (t *teamsIntegration) startOAuth(ctx context.Context, tenantHint string) (*deviceCodeResponse, error) {
+	if tenantHint == "" {
+		tenantHint = "common"
+	}
+	t.mu.RLock()
+	loginBase := t.loginBaseURL
+	clientID := t.clientID
+	clientSecret := "" // We'll allow override via creds later; not stored on the struct yet.
```
acmacalister: `client_secret` is wired in config, env mapping, and README but always hardcoded to `""` here — the note "not stored on the struct yet" signals intent, but right now users who set `TEAMS_CLIENT_SECRET` get silent no-op behavior. Either read it from the struct (which also needs a `clientSecret` field on `teamsIntegration` and a read in `Configure`) or drop it from the config/README until it's implemented. The current state of advertising a credential that does nothing is confusing.

```go
// In Configure, after reading t.clientID:
t.clientSecret = strings.TrimSpace(creds["client_secret"])

// In startOAuth:
clientSecret := t.clientSecret
```

integrations/teams/oauth.go
```diff
@@ -0,0 +1,409 @@
...
+			continue
+		}
+		req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+		req.Header.Set("Accept", "application/json")
+
+		resp, err := f.integration.httpClient.Do(req)
+		if err != nil {
+			continue
+		}
+		body, _ := io.ReadAll(resp.Body)
```
acmacalister: `io.ReadAll` here has no size cap, unlike `graphRequestInner` which wraps the body in `io.LimitReader(resp.Body, maxResponseSize)`. A misconfigured or malicious token endpoint could stream an arbitrarily large response and OOM the process. Worth applying the same limit here:

```go
body, _ := io.ReadAll(io.LimitReader(resp.Body, maxResponseSize))
```

acmacalister (commented):
Solid implementation — the multi-tenant token store, proactive refresh, and 401-triggered retry are all handled cleanly. Tests are thorough and the dispatch/compaction parity tests are all present. Two things inline.

integrations/teams/oauth.go
daltoniam: Done in ac2e59f — added `clientSecret` to the struct, read it in `Configure`, and pass it through both the device-code and refresh-token flows. Added to `OptionalKeys` (kept out of `PlainTextKeys` so it stays redacted).

integrations/teams/oauth.go
daltoniam: Done in ac2e59f — all three `io.ReadAll` calls in oauth.go now wrap the body in `io.LimitReader(resp.Body, maxResponseSize)` to match `graphRequestInner`.

integrations/teams/teams.go
```diff
@@ -0,0 +1,439 @@
...
+	// Resolve user identity for each loaded tenant. Failures are logged but
+	// non-fatal — tokens may legitimately be expired and refreshable.
+	t.resolveTenantIdentities(ctx)
+
+	// Background refresh: skips for config-injected tokens since they're
+	// managed externally.
+	if t.stopBg != nil {
+		close(t.stopBg)
+	}
+	t.stopBg = make(chan struct{})
```
acmacalister: `t.stopBg` is written here outside `t.mu` (the lock was dropped at the `t.mu.Unlock()` above), and read by the `backgroundRefresh` goroutine in its `select` without any lock. If `Configure` is called a second time while the goroutine is still running, the Go race detector will flag a data race on the channel value itself.

The fix is to move the `stopBg` management inside the mutex, or to capture the channel into a local variable before launching the goroutine and pass it as a parameter:

```go
// inside t.mu.Lock() scope:
if t.stopBg != nil {
    close(t.stopBg)
}
t.stopBg = make(chan struct{})
stopCh := t.stopBg
t.mu.Unlock()

go t.backgroundRefresh(bgCtx, stopCh)
```

Then `backgroundRefresh` accepts `stop <-chan struct{}` instead of reading `t.stopBg` directly.

integrations/teams/oauth.go
```diff
@@ -0,0 +1,413 @@
...
+		form := url.Values{
+			"client_id":   {f.clientID},
+			"device_code": {f.deviceResp.DeviceCode},
+			"grant_type":  {"urn:ietf:params:oauth:grant-type:device_code"},
+		}
+		if f.clientSecret != "" {
+			form.Set("client_secret", f.clientSecret)
+		}
+
+		req, err := http.NewRequest(http.MethodPost, endpoint, strings.NewReader(form.Encode()))
```
acmacalister: `http.NewRequest` here (without context) is the only HTTP call in the package not using `http.NewRequestWithContext`. The poll goroutine will continue running until the device code expires (~15 min) even if the server shuts down, because it has no way to be cancelled.

This is fixable by threading a context through from `startOAuth` into the `oauthFlow` struct:

```go
type oauthFlow struct {
    ctx context.Context  // add this
    ...
}

// startOAuth:
flow := &oauthFlow{
    ctx: ctx,
    ...
}

// poll():
req, err := http.NewRequestWithContext(f.ctx, http.MethodPost, endpoint, strings.NewReader(form.Encode()))
```

integrations/teams/oauth.go
```diff
@@ -0,0 +1,413 @@
...
+		tenantHint = "common"
+	}
+	t.mu.RLock()
+	loginBase := t.loginBaseURL
+	clientID := t.clientID
+	clientSecret := t.clientSecret
+	scopes := t.scopes
+	t.mu.RUnlock()
+
+	endpoint := fmt.Sprintf("%s/%s/oauth2/v2.0/devicecode", loginBase, tenantHint)
```
acmacalister: `tenantHint` is user-supplied (the `tenant` arg from `teamsLogin`) and is interpolated directly into the URL path without `url.PathEscape`. Everywhere else in this package — team IDs, channel IDs, user IDs — are properly escaped. Worth being consistent:

```go
endpoint := fmt.Sprintf("%s/%s/oauth2/v2.0/devicecode", loginBase, url.PathEscape(tenantHint))
```

acmacalister (commented):
Solid integration — clean architecture, good test coverage, proactive token refresh, proper resource limits everywhere. Two concurrency nits and one consistency note inline.

integrations/teams/teams.go
daltoniam: Done in 388b125 — moved `stopBg` allocation inside `t.mu`, captured into a local `stopCh`, and pass it as a parameter to `backgroundRefresh`. The goroutine no longer reads `t.stopBg` directly, so a re-`Configure` can't race.

integrations/teams/oauth.go
daltoniam: Done in 388b125 — added `ctx context.Context` field on `oauthFlow` (set via `context.WithoutCancel(ctx)` from `startOAuth` so request-scoped cancel doesn't kill the long poll, but the field opens the door to a shutdown signal later). Switched `poll()` to `NewRequestWithContext` and replaced `time.Sleep` with a `select { case <-ctx.Done(): ...; case <-time.After(interval): }` so cancellation actually exits the loop.

integrations/teams/oauth.go
daltoniam: Done in 388b125 — wrapped `tenantHint` in `url.PathEscape` in all three OAuth endpoint constructions (devicecode, token poll, refresh).

integrations/teams/oauth.go
```diff
@@ -0,0 +1,425 @@
...
+func base64URLDecode(s string) ([]byte, error) {
+	// JWT spec uses base64url without padding.
+	return base64.RawURLEncoding.DecodeString(strings.TrimRight(s, "="))
+}
+
+// refreshTenant exchanges a refresh token for a fresh access token. Called by
+// graphRequestInner on 401 and proactively before expiry.
+func (t *teamsIntegration) refreshTenant(ctx context.Context, tn *tenant) error {
+	if tn == nil || tn.RefreshToken == "" {
+		return fmt.Errorf("no refresh_token available for tenant %s", tn.TenantID)
```
acmacalister, line 353: The nil guard on line 349 is correct, but the error message on line 350 dereferences `tn.TenantID` unconditionally — if `tn` really is `nil`, this panics instead of returning an error. The defensive check ends up introducing the very crash it was meant to prevent.

```go
if tn == nil {
    return fmt.Errorf("refreshTenant called with nil tenant")
}
if tn.RefreshToken == "" {
    return fmt.Errorf("no refresh_token available for tenant %s", tn.TenantID)
}
```

integrations/teams/teams.go
```diff
@@ -0,0 +1,442 @@
...
+	// falls back to "_config" when not provided.
+	if at := strings.TrimSpace(creds["access_token"]); at != "" {
+		tid := t.defaultTenant
+		if tid == "" {
+			tid = "_config"
+		}
+		t.store.upsert(&tenant{
+			TenantID:     tid,
+			AccessToken:  at,
+			RefreshToken: strings.TrimSpace(creds["refresh_token"]),
```
acmacalister, line 144: `expires_at` is declared in `OptionalKeys` and included in the default config (so it shows up in the web UI and can be set via `TEAMS_EXPIRES_AT`), but `Configure` never reads it. A config-injected token will always have `ExpiresAt.IsZero() == true`, meaning `shouldProactivelyRefresh` returns false and the token only gets refreshed after a real 401. Worth parsing it alongside `access_token`:

```go
if raw := strings.TrimSpace(creds["expires_at"]); raw != "" {
    if exp, err := time.Parse(time.RFC3339, raw); err == nil {
        tn.ExpiresAt = exp
    }
}
```
Slot this into the `upsert` block on line 141 before calling `store.upsert`.

integrations/teams/users.go
```diff
@@ -0,0 +1,123 @@
...
+	q := url.Values{}
+	q.Set("$top", fmt.Sprintf("%d", top))
+	if filter != "" {
+		q.Set("$filter", filter)
+	}
+	if selectFields != "" {
+		q.Set("$select", selectFields)
+	}
+	if search != "" {
+		q.Set("$search", fmt.Sprintf("\"displayName:%s\" OR \"userPrincipalName:%s\"", search, search))
```
acmacalister: The `search` value goes straight into the OData `$search` string with no quoting. A value containing a double-quote — e.g. `foo" OR "mail:secret` — would produce a malformed expression and an unexpected Graph 400. The same pattern repeats in `searchUsers` at line 98. A minimal fix is to strip or escape inner double-quotes before interpolating:

```go
search = strings.ReplaceAll(search, `"`, "")
q.Set("$search", fmt.Sprintf(`"displayName:%s" OR "userPrincipalName:%s"`, search, search))
```

acmacalister (commented):
Solid adapter — the multi-tenant token store, proactive refresh, and device-code flow are well implemented. Three things worth a look inline.

integrations/teams/oauth.go
daltoniam: Done in 0e69030 — split the guard so a `nil` tenant returns "refreshTenant called with nil tenant" without ever touching `tn.TenantID`. Added `TestRefreshTenant_NilTenant`.

integrations/teams/teams.go
daltoniam: Done in 0e69030 — `Configure` now parses `expires_at` as RFC3339 and stores it on the tenant, so config-injected tokens get proactive refresh before expiry. Added `TestConfigure_ParsesExpiresAt`.

integrations/teams/users.go
daltoniam: Done in 0e69030 — added `sanitizeSearchTerm` that strips bare double-quotes and backslashes; both `listUsers` and `searchUsers` feed user input through it before interpolating into the OData `$search` expression. Added `TestSanitizeSearchTerm` covering the `foo" OR "mail:secret` case.

integrations/teams/teams_test.go
```diff
@@ -0,0 +1,413 @@
...
+	p := i.Placeholders()
+	assert.NotEmpty(t, p["tenant_id"])
+	assert.Equal(t, defaultGraphBaseURL, p["graph_base_url"])
+}
+
+// --- Configure ---
+
+func TestConfigure_DefaultsApplied(t *testing.T) {
+	ti := newTestIntegration(t, nil)
+	err := ti.Configure(context.Background(), mcp.Credentials{})
```
acmacalister, line 121: These `Configure` tests call `resolveTenantIdentities` which fires a real `GET https://graph.microsoft.com/v1.0/me` for each stored tenant. The tests pass only because the 401 is non-fatal, but they're making live network calls and will be slow or broken in air-gapped CI. Since `newTestIntegration` already knows how to point at an `httptest.Server`, these tests could wire a stub graph server the same way `TestGraphGet_AddsBearerToken` does — set it on `ti.graphBaseURL` before calling `Configure`, and the `resolveTenantIdentities` probe will hit the stub instead of the real internet.

acmacalister, line 121: The `Configure` calls here also each spawn a `backgroundRefresh` goroutine that lives until `stopBg` is closed. Since there's no `t.Cleanup(func() { close(ti.stopBg) })`, these goroutines outlive the test and accumulate across the test suite run. The background ticker fires every 15 minutes so they stay quiet during tests, but they'd show up if the project ever adds goleak. A simple cleanup registration would fix it:

```go
t.Cleanup(func() { close(ti.stopBg) })
```

acmacalister (approved):
Solid integration — the multi-tenant token store, proactive refresh, and device-code flow are all well-structured. Two small test hygiene items inline; nothing blocking.

integrations/teams/teams_test.go
daltoniam: Done in 4d10779 — `newTestIntegration` now always wires `graphBaseURL` to a local `httptest.Server` (a 401 stub when no caller-supplied server is passed), so `Configure`-spawned `resolveTenantIdentities` calls never reach the real graph.microsoft.com.

daltoniam: Done in 4d10779 — `newTestIntegration` registers a `t.Cleanup` that closes `ti.stopBg` (under `ti.mu`), so the `backgroundRefresh` goroutine `Configure` spawns is reaped at test exit instead of accumulating.

integrations/teams/auth.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+// teamsLoginPoll reports the state of the active device-code flow.
+func teamsLoginPoll(_ context.Context, _ *teamsIntegration, _ map[string]any) (*mcp.ToolResult, error) {
+	res := pollOAuth()
+	return mcp.JSONResult(res)
+}
+
+// teamsTokenStatus reports per-tenant token health.
+func teamsTokenStatus(_ context.Context, t *teamsIntegration, _ map[string]any) (*mcp.ToolResult, error) {
```
acmacalister, line 68: `teamsTokenStatus` has non-trivial logic — it computes `age_minutes`, determines `status` ("expired" / "expiring_soon" / "healthy"), and formats `expires_in` — but has no unit test. A subtle bug like checking `remaining <= 0` vs `remaining < 0`, or the `5*time.Minute` threshold, would be invisible. Same gap covers `teamsRefreshTokens`, `teamsGetMe`, `formatContentType`, and all the chat/channel/user handlers. The HTTP layer is solid, but argument-parsing and query-construction paths are untested.

A lightweight table-driven test covering the three status branches would cost about 20 lines:

```go
func TestTeamsTokenStatus(t *testing.T) {
    cases := []struct{
        name    string
        exp     time.Time
        want    string
    }{
        {"healthy",  time.Now().Add(10*time.Minute), "healthy"},
        {"expiring", time.Now().Add(2*time.Minute),  "expiring_soon"},
        {"expired",  time.Now().Add(-1*time.Second), "expired"},
    }
    for _, tc := range cases {
        t.Run(tc.name, func(t *testing.T) {
            ti := newTestIntegration(t, nil)
            ti.store.upsert(&tenant{TenantID: "t1", AccessToken: "a", ExpiresAt: tc.exp})
            res, err := ti.Execute(context.Background(), "teams_token_status", nil)
            require.NoError(t, err)
            assert.Contains(t, res.Data, tc.want)
        })
    }
}
```

integrations/teams/teams.go
```diff
@@ -0,0 +1,450 @@
...
+	}
+	defer func() { _ = resp.Body.Close() }()
+
+	data, err := io.ReadAll(io.LimitReader(resp.Body, maxResponseSize))
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode == http.StatusUnauthorized && canRetry && tn.RefreshToken != "" {
+		if rerr := t.refreshTenant(ctx, tn); rerr == nil {
```
acmacalister, line 354: The `tn` variable here is a snapshot read at line 304, before the HTTP round-trip. If `backgroundRefresh` fires concurrently during that round-trip and Microsoft rotates the refresh token, `tn.RefreshToken` is now stale — the refresh will fail with "invalid refresh token" even though the current token is sitting in the store. The window is narrow, but the fix is cheap:

```go
if resp.StatusCode == http.StatusUnauthorized && canRetry {
    if fresh := t.store.get(tenantID); fresh != nil && fresh.RefreshToken != "" {
        if rerr := t.refreshTenant(ctx, fresh); rerr == nil {
            return t.graphRequestInner(ctx, tenantID, method, path, body, extraHeaders, false)
        }
    }
}
```

acmacalister (approved):
Solid integration — auth infrastructure (proactive refresh, 401 retry, device-code poll with slow_down handling, atomic token file writes) is well thought out and the HTTP layer tests are thorough. All previous review comments addressed. Two small things inline.

PR #163, opened 2026-05-14
Self-heal stale Slack xoxc tokens at startup

daltoniam:
## Summary

- `xoxc-*` Slack browser session tokens rotate constantly. Two startup-time failure modes were leaving workspaces dead until a manual re-extract from the web UI: (a) `resolveWorkspaceIdentities` logged `auth test failed for workspace …: invalid_auth` and moved on without attempting recovery, and (b) `Configure` treated *any* config-supplied token as authoritative, skipping both `loadFromFile()` and the background refresh — which is correct for OAuth/bot tokens but wrong for `xoxc-*` browser snapshots whose fresher copy is sitting in `~/.slack-mcp-tokens.json`.
- This PR makes startup self-healing: a one-shot cookie/browser refresh + retry on initial auth.test failure, and a new "browser-snapshot config token" path (detected via `token_source=browser` or `xoxc-` prefix) that loads the persisted file, lets its fresher token win, and enables the normal background refresh loop. Genuine externally-managed `xoxp-`/`xoxb-` tokens keep the existing "config is authoritative, no local refresh" behavior so they're never clobbered.

## Test plan

- [x] `go test -race ./integrations/slack/` — 9 new tests added (5 self-heal paths + 4 Configure paths); all pass
- [x] `go tool golangci-lint run ./integrations/slack/` — 0 issues
- [x] `go vet ./...` — clean
- [x] `make build` — succeeds
- [x] Verified end-to-end on a real workspace whose persisted `xoxc-` token had gone stale: restarting switchboard now logs `attempting startup refresh` → `tokens refreshed from chrome` → `auth recovered for workspace T… after startup refresh` instead of the previous dead-until-manual-re-extract failure

💘 Generated with Crush

Assisted-by: Claude Sonnet 4.5 via Crush <crush@charm.land>

### DIFF
--- a/integrations/slack/slack.go
+++ b/integrations/slack/slack.go
@@ -30,6 +30,9 @@ type slackIntegration struct {
 	clients map[string]*slack.Client // keyed by team_id
 	store   *tokenStore
 	stopBg  chan struct{}
+
+	// refreshWorkspace is overridable in tests; defaults to (*slackIntegration).tryRefreshWorkspace.
+	refreshWorkspace func(teamID string) bool
 }
 
 func New() mcp.Integration {
@@ -45,25 +48,40 @@ func (s *slackIntegration) Configure(ctx context.Context, creds mcp.Credentials)
 	s.mu.Unlock()
 
 	configToken := creds["token"]
+	// xoxc-* tokens are browser session tokens that rotate constantly. Even
+	// when bootstrapped from config, they must participate in the local
+	// refresh/file-persistence loop or they go stale within hours.
+	//   - token_source: "browser"            → explicitly a browser snapshot
+	//   - token has xoxc- prefix             → structurally a session token
+	// Genuine externally-managed tokens (xoxb-, xoxp-, OAuth) keep the old
+	// "config is authoritative" behavior so we never clobber them.
+	isBrowserConfig := configToken != "" &&
+		(creds[mcp.CredKeyTokenSource] == "browser" || strings.HasPrefix(configToken, "xoxc-"))
+
 	if configToken != "" {
 		teamID := creds["team_id"]
 		if teamID == "" {
 			teamID = "_config"
 		}
+		source := "config"
+		if isBrowserConfig {
+			source = "browser"
+		}
 		s.store.setWorkspace(&workspace{
 			TeamID: teamID,
 			Token:  configToken,
 			Cookie: creds["cookie"],
-			Source: "config",
+			Source: source,
 		})
 		if creds["team_id"] != "" {
 			s.store.setDefault(creds["team_id"])
 		}
 	}
 
-	// When a token was provided via config (e.g. OAuth in hosted environments),
-	// skip local file and Chrome extraction — the config token is authoritative.
-	if configToken == "" {
+	// When the config token is a rotating browser snapshot, also load the
+	// persisted file. The file may carry a fresher copy (background refresh
+	// writes there) and we want that to win.
+	if configToken == "" || isBrowserConfig {
 		s.store.loadFromFile()
 
 		if len(s.store.allWorkspaces()) == 0 {
@@ -89,9 +107,10 @@ func (s *slackIntegration) Configure(ctx context.Context, creds mcp.Credentials)
 		s.store.setDefault(tid)
 	}
 
-	// Only persist and run background refresh for locally-sourced tokens.
-	// Config-provided tokens are managed externally and don't need local refresh.
-	if configToken == "" {
+	// Persist + run background refresh for any non-OAuth-style token (file-
+	// loaded or browser-sourced from config). Only genuine externally-managed
+	// tokens skip this.
+	if configToken == "" || isBrowserConfig {
 		_ = s.store.saveToFile()
 
 		if s.stopBg != nil {
@@ -131,8 +150,11 @@ func (s *slackIntegration) resolveWorkspaceIdentities(ctx context.Context) {
 		}
 		resp, err := client.AuthTestContext(ctx)
 		if err != nil {
-			log.Printf("slack: auth test failed for workspace %s: %v", ws.TeamID, err)
-			continue
+			recovered, retryResp := s.tryRecoverAuth(ctx, ws, err)
+			if !recovered {
+				continue
+			}
+			resp = retryResp
 		}
 		if ws.TeamID == resp.TeamID {
 			if ws.TeamName == "" {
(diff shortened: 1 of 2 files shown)


acmacalister (commented):
Solid work — the self-healing logic is clean, the refreshFn indirection for test injection is elegant, and the 9 new tests cover the important policy branches thoroughly. Two small things inline.

integrations/slack/slack.go
```diff
@@ -166,6 +188,50 @@ func (s *slackIntegration) resolveWorkspaceIdentities(ctx context.Context) {
...
+// log diagnostics. xoxc-/xoxp-/xoxb-/xoxd- prefixes are intentionally kept;
+// the secret body is replaced with ellipsis.
+func tokenPrefix(tok string) string {
+	if tok == "" {
+		return "<empty>"
+	}
+	if len(tok) <= 6 {
+		return tok
+	}
+	return tok[:6] + "…"
```
acmacalister: `"xoxc-"` is 5 bytes, so `tok[:6]` captures the prefix *plus the first character of the secret body*. The comment above says we intentionally keep only the prefix — this should be `tok[:5]` to match that intent.

```go
return tok[:5] + "…"
```

Low severity (one leaked byte, only in logs), but worth fixing to match the stated design.

integrations/slack/slack.go
```diff
@@ -166,6 +188,50 @@ func (s *slackIntegration) resolveWorkspaceIdentities(ctx context.Context) {
...
+		return false, nil
+	}
+	log.Printf("slack: auth test failed for workspace %s: %v — attempting startup refresh", ws.TeamID, origErr)
+	if !s.refreshFn()(ws.TeamID) {
+		log.Printf("slack: startup refresh failed for workspace %s — manual re-extract may be required (open Slack in Chrome and re-extract via web UI)", ws.TeamID)
+		return false, nil
+	}
+	c := s.getClientForTeam(ws.TeamID)
+	if c == nil {
+		log.Printf("slack: startup refresh produced no client for workspace %s", ws.TeamID)
```
acmacalister, line 214: This guard looks unreachable: `tryRefreshWorkspace` always calls `buildClientForWorkspace(ws)` before returning `true`, so a successful refresh will always leave a client in place. Worth adding a brief comment explaining the defensive intent, or (if it really is unreachable) dropping it and letting the nil dereference surface as a panic — which would be a louder signal than a logged warning that silently leaves the workspace dead.

integrations/slack/slack_test.go
```diff
@@ -401,9 +405,264 @@ func TestConfigure_SwitchToConfigTokenStopsBackgroundRefresh(t *testing.T) {
...
+	s.Configure(t.Context(), mcp.Credentials{
+		"token":                "xoxc-rotating-session-token",
+		"team_id":              "T_BROWSER",
+		mcp.CredKeyTokenSource: "browser",
+	})
+
+	ws := s.store.getWorkspace("T_BROWSER")
+	require.NotNil(t, ws)
+	assert.Equal(t, "browser", ws.Source, "browser-source xoxc-* config token should not be marked source=config")
+	assert.NotNil(t, s.stopBg, "background refresh should run for browser-snapshot tokens")
```
acmacalister: This test starts a background goroutine (via `Configure`) but never stops it. The goroutine blocks on a 4-hour ticker and `s.stopBg`, so it will keep running until the test binary exits. Add a cleanup so the goroutine is stopped when the test finishes:

```go
assert.NotNil(t, s.stopBg, "background refresh should run for browser-snapshot tokens")
t.Cleanup(func() { close(s.stopBg) })
```

Same pattern applies to the identical assertion in `TestConfigure_XoxcConfigTokenWithoutSourceFlagIsStillBrowser` (line 448).

integrations/slack/slack.go
```diff
@@ -175,6 +247,29 @@ func (s *slackIntegration) getClientForTeam(teamID string) *slack.Client {
...
+// managed externally; OAuth (xoxp-*) tokens do not rotate and a refresh path
+// for them does not exist.
+func (s *slackIntegration) canSelfRefresh(ws *workspace) bool {
+	if ws == nil || ws.Source == "config" {
+		return false
+	}
+	if strings.HasPrefix(ws.Token, "xoxp-") {
+		return false
+	}
+	return true
```
acmacalister: The guard blocks `nil` workspace, `source=config`, and `xoxp-*` tokens, but it allows through `xoxb-*` (bot) and `xoxd-*` (cookie-only) tokens if they happen to have a non-config source. In practice this can't happen through the UI (which only sets `token_source=browser` alongside `xoxc-*` tokens), but a hand-edited config with `xoxb-* + token_source=browser` would end up here and trigger a browser-extract attempt on a bot token.

Since the entire premise of this feature is that *only* `xoxc-*` tokens rotate via the browser path, tightening to an explicit prefix check makes the invariant self-documenting and removes the gap:

```go
func (s *slackIntegration) canSelfRefresh(ws *workspace) bool {
	if ws == nil || ws.Source == "config" {
		return false
	}
	// Only xoxc-* browser session tokens rotate and have a local refresh path.
	return strings.HasPrefix(ws.Token, "xoxc-")
}
```

acmacalister (commented):
Nice approach — the `refreshFn` injection point keeps the logic testable without any interface churn, and the `tokenPrefix` leak-test is a good safety net. Two small things inline, both easy fixes.

integrations/slack/slack.go
```diff
@@ -175,6 +247,29 @@ func (s *slackIntegration) getClientForTeam(teamID string) *slack.Client {
...
+// managed externally; OAuth (xoxp-*) tokens do not rotate and a refresh path
+// for them does not exist.
+func (s *slackIntegration) canSelfRefresh(ws *workspace) bool {
+	if ws == nil || ws.Source == "config" {
+		return false
+	}
+	// Only xoxc-* browser session tokens rotate and have a local refresh
+	// path. xoxb-/xoxp-/xapp-/etc. are externally managed and must never
+	// be replaced by a browser extract, even when stored locally.
+	return strings.HasPrefix(ws.Token, "xoxc-")
```
acmacalister, line 261: Now that `canSelfRefresh` tightens the policy to `xoxc-` only, `tryRefresh()` (the background loop) is inconsistent — it still uses the old `if strings.HasPrefix(ws.Token, "xoxp-") { continue }` guard, so it will attempt a browser/cookie extraction on `xoxb-`, `xoxd-`, and `xapp-` tokens that this function would correctly reject. Today the blast radius is limited because `backgroundRefresh` only starts when `isBrowserConfig` is true, but if the store ever ends up with mixed token types the loop will call `tryRefreshWorkspace` on tokens that shouldn't be refreshed.

Worth unifying the filter:

```go
func (s *slackIntegration) tryRefresh() bool {
	allOk := true
	for _, ws := range s.store.allWorkspaces() {
		if !s.canSelfRefresh(ws) {
			continue
		}
		if !s.tryRefreshWorkspace(ws.TeamID) {
			allOk = false
		}
	}
	return allOk
}
```

This also keeps the skip policy in one place so future token types only need to be handled in `canSelfRefresh`.

acmacalister (commented):
Great work on this — the two-axis detection, the `refreshWorkspace` seam, and the `tokenPrefix` safety property tests are all solid. One inline note on a filter inconsistency between the new `canSelfRefresh` and the existing background `tryRefresh` loop.

acmacalister (approved):
All previous review feedback has been addressed — token prefix is correctly 5 bytes, canSelfRefresh is tightened to xoxc- only, tryRefresh delegates to canSelfRefresh, and the test goroutine cleanups are in place. Build passes, race detector is clean, lint is 0 issues. LGTM.

PR #164, opened 2026-05-15
feat(compact): yaml specs for custom views and compaction fields

arjansingh:
## Problem

It's actually really annoying to tweak the compaction settings when they are go structs, especially if someone isn't super technical.

## Move Compaction Specs to YAML

By moving them to YAML files, the compaction settings for each integration are easier to read, making them easier to refine.

They also can now be made run-time files (not done in this PR yet), so that different users can adjust to include or remove fields  they don't need.

## Custom Views

One nice side effect of this is that we can also can now dynamically create custom views (e.g., summary views for a document) for each tool. Eventually this can be exposed for configuration in the web interface as well.


server/project_server.go
```diff
@@ -332,7 +332,7 @@ func (pr *ProjectRouter) makeExecuteHandler(def *project.Definition, scopeRule *
 		}
 
 		if !result.IsError {
-			result.Data = processResult(buildResultProcessor(integration), tool, result.Data, pr.services.Metrics)
+			result.Data = processResult(buildResultProcessor(integration), tool, nil, result.Data, pr.services.Metrics)
```
acmacalister: Passing `nil` here disables view selection for the project router, which creates a subtle inconsistency: view-aware tools (like `notion_get_page_content`) will still go through `processViewsResult` and get a `_more` envelope advertising alternate views — but those views are unreachable because the project router's arg validation doesn't allow `view`/`format`, and view selection is silently skipped.

The `_more` hint teaches the LLM about views it can't actually use here, which wastes a tool call and produces a confusing error. `args.Arguments` is available at this call site — passing it would enable view selection consistently (the same way `handleExecute` does):

```go
result.Data = processResult(buildResultProcessor(integration), tool, args.Arguments, result.Data, pr.services.Metrics)
```

If disabling views in the project context is intentional, a comment explaining why would prevent this from looking like an oversight when the next person touches this line.

acmacalister (approved):
Solid PR — clean migration of 30+ adapters to YAML specs, all tests pass with race detector, lint is clear. The lenient-at-runtime / strict-in-CI split is a nice pattern, and the load-time renderer resolution gives good early failure on broken configs. One thing to look at inline.

server/server.go
```diff
@@ -908,9 +928,213 @@ func processResult(rp resultProcessor, toolName mcp.ToolName, data string, metri
...
+	}
+
+	out = appendMoreHint(out, viewSet, selection)
+
+	if view.MaxBytes > 0 && len(out) > view.MaxBytes {
+		return tooLargeEnvelope(toolName, len(out), view.MaxBytes)
+	}
+
+	if metrics != nil {
+		metrics.RecordCompaction(toolName, originalLen, len(out))
```
acmacalister, line 1014: The flat path (`processResult`) only records `RecordCompaction` when `len(result) < originalLen`, but here it's called unconditionally. When `appendMoreHint` adds the `_more` envelope, the output can be larger than the original — this would log a "compaction" event with negative savings, which will look odd in monitoring dashboards.

```go
if metrics != nil && len(out) < originalLen {
	metrics.RecordCompaction(toolName, originalLen, len(out))
}
```

server/server.go
```diff
@@ -908,9 +928,213 @@ func processResult(rp resultProcessor, toolName mcp.ToolName, data string, metri
...
+
+	return selection, nil
+}
+
+func listViewNames(vs compact.ViewSet) string {
+	names := make([]string, 0, len(vs.Views))
+	for v := range vs.Views {
+		names = append(names, string(v))
+	}
+	return strings.Join(names, ", ")
```
acmacalister, line 1056: Both `listViewNames` and `listFormats` iterate over maps, so the order of the comma-separated values in error messages is non-deterministic across runs. No existing test currently asserts the exact ordering, so nothing is broken today — but the next test that does assert message content will flake. `sort.Strings` before `strings.Join` closes the gap.

```go
func listViewNames(vs compact.ViewSet) string {
	names := make([]string, 0, len(vs.Views))
	for v := range vs.Views {
		names = append(names, string(v))
	}
	sort.Strings(names)
	return strings.Join(names, ", ")
}
```

Same fix for `listFormats`.

acmacalister (approved):
Solid PR — builds clean, all tests pass with -race, lint is quiet. The load-time strict/lenient split is a great design, and the back-compat mirror into Specs means this lands without touching every downstream caller. Two small things inline worth a look before this ships.

compact/overlay.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+// EnvOverrideDir is the environment variable that points to a directory
+// containing per-adapter YAML override files.
+const EnvOverrideDir = "SWITCHBOARD_COMPACT_DIR"
+
+// LoadWithOverlay loads embedded defaults, then merges per-tool overrides from
+// $SWITCHBOARD_COMPACT_DIR/<adapter>.yaml if the file exists.
+// Merge is per-tool key: overlay entries replace embedded ones; missing entries fall through.
+func LoadWithOverlay(adapter string, embedded []byte, opts Options) (Result, error) {
```
acmacalister, line 18: Overlay load errors and warnings (bad YAML, read failures, unknown tool names) get absorbed into `base.Warnings` via `absorbOverlayErr`, but `Result.Warnings` is never read by any adapter. Operators setting `SWITCHBOARD_COMPACT_DIR` with a broken override file would have silent failures — their customizations just don't apply, with no log message to explain why.

The simplest fix is to log warnings inside `MustLoadWithOverlay` before returning:

```go
func MustLoadWithOverlay(adapter string, embedded []byte, opts Options) Result {
	res, err := LoadWithOverlay(adapter, embedded, opts)
	if err != nil {
		panic(err)
	}
	for _, w := range res.Warnings {
		slog.Warn("compact: load warning", "adapter", adapter, "err", w)
	}
	return res
}
```

Alternatively, `LoadWithOverlay` could log them directly before returning. Either way, a silent broken override is the kind of thing that takes hours to diagnose in production.

integrations/notion/notion.go
```diff
@@ -12,14 +13,62 @@ import (
...
+})
+
+var fieldCompactionSpecs = compactResult.Specs
+var maxBytesByTool = compactResult.MaxBytes
+var viewSets = compactResult.Views
+
 var (
 	_ mcp.Integration                = (*notion)(nil)
 	_ mcp.FieldCompactionIntegration = (*notion)(nil)
 	_ mcp.MarkdownIntegration        = (*notion)(nil)
```
acmacalister, line 69: `markdownRenderers` is now empty, so `RenderMarkdown` always returns `(nil, false)`. Since the views pipeline owns all rendering, the `mcp.MarkdownIntegration` interface assertion isn't pulling its weight anymore — it just causes `buildResultProcessor` to wire up a markdown function that always returns false for every non-view-tool call.

Worth dropping the assertion and the `RenderMarkdown` method (or at least the compile-time assertion) to reflect the new reality. The SKILL.md guidance says to remove a tool from `markdownRenderers` when it moves to views; this is the next step of that same cleanup.

acmacalister (approved):
Solid PR. The parse-don't-validate pattern, the load-time renderer resolution, the ViewArgs typed boundary — these are all good design choices that will pay off when the next adapter adds views. Build, tests, and lint all pass. Two small things inline.

integrations/notion/markdown.go
```diff
@@ -47,12 +47,13 @@ type renderedComment struct {
...
-	"notion_get_page_content":  renderPageContentMD,
-	"notion_retrieve_comments": renderCommentsMD,
-}
+// markdownRenderers maps tool names to their markdown rendering functions
+// for the legacy MarkdownIntegration path. Tools that opted into the views
+// pipeline (get_page_content, retrieve_comments) bypass this map at runtime —
+// processResult dispatches via the views renderer registry first. The map
+// is left in place so RenderMarkdown stays callable for tools that have
+// not yet migrated to views.
+var markdownRenderers = map[mcp.ToolName]func([]byte) (markdown.Markdown, bool){}
```
acmacalister, line 56: `markdownRenderers` is now empty, which means `RenderMarkdown` always returns `false` for every tool — `get_page_content` and `retrieve_comments` both moved to the views pipeline. The comment says "tools that have not yet migrated to views" but at this point there are none left in notion.

The concern is for the next developer who adds a new markdown-rendering tool to notion: they'll see `markdownRenderers` and add their entry there, the map will silently do nothing (views pipeline owns rendering now), and they'll have a confusing debugging session. The `_ mcp.MarkdownIntegration = (*notion)(nil)` assertion in `notion.go` stays green because the interface is satisfied, but `buildResultProcessor` will set `rp.markdown` for notion on every call, and it'll always miss.

If the empty `MarkdownIntegration` path is being kept intentionally as a migration on-ramp for future tools, a stronger comment makes this explicit:

```go
// markdownRenderers is intentionally empty — all notion tools that return
// document content use the views pipeline (compactRenderers in notion.go).
// New tools with rich markdown output should register a custom renderer in
// compactRenderers, NOT here.
var markdownRenderers = map[mcp.ToolName]func([]byte) (markdown.Markdown, bool){}
```

Otherwise, the cleaner move is to drop `markdownRenderers`, `RenderMarkdown`, and the `_ mcp.MarkdownIntegration` assertion entirely — `buildResultProcessor` won't wire the no-op path, and there's one fewer empty map to explain.

compact/overlay.go
```diff
@@ -0,0 +1,88 @@
...
+		return Result{}, err
+	}
+
+	dir := os.Getenv(EnvOverrideDir)
+	if dir == "" {
+		return base, nil
+	}
+
+	path := filepath.Join(dir, adapter+".yaml")
+	data, err := os.ReadFile(path) // #nosec G703 -- dir is operator-controlled (SWITCHBOARD_COMPACT_DIR), adapter is a hardcoded integration name
```
acmacalister, line 30: The `#nosec` annotation cites G703 ("Errors unhandled"), but the error on this line is handled — `err` is checked immediately below. If the intent was to suppress G304 ("File path provided as taint input"), the annotation should be `#nosec G304`. The path is safe here (`adapter` is a hardcoded integration name, `dir` is operator-set via `SWITCHBOARD_COMPACT_DIR`), so the suppression is justified — just for the wrong rule.

```go
data, err := os.ReadFile(path) // #nosec G304 -- adapter is a hardcoded name; dir is operator-controlled via SWITCHBOARD_COMPACT_DIR
```

acmacalister (approved):
Builds, tests pass, lint clean. The three things from the previous review session are all fixed in the latest commits. Two small things inline — both are cleanup-level, neither blocks.

PR #165, opened 2026-05-15
Fix Jira/Confluence list and nested-list formatting

PerishableDave:
## Summary

- **Confluence lists broken**: The HTML renderer (`markdown/html.go`) passed `<li>` children directly to `renderNode`, so `<p>` elements inside list items emitted `\n\n`, producing `- text\n\n\n` per item. Confluence storage format wraps all list-item content in `<p>` tags, so this affected essentially every Confluence list. Fixed by buffering inline children and `TrimRight`-trimming trailing whitespace before writing each item with a single `\n`.
- **Jira ADF nested lists**: `renderADFNodes` always appended a trailing `\n` after bullet/ordered lists regardless of nesting depth, creating an extra blank line after nested lists. Fixed to only emit the trailing blank-line separator when at top level (`listPrefix == ""`).

## Test plan

- [ ] `make ci` passes (all tests green, no lint/vet/security findings)
- [ ] New `TestFromHTML` cases cover `<p>`-wrapped `<li>` items for both `<ul>` and `<ol>`, and nested lists with `<p>`-wrapped parent items
- [ ] New `TestADFToMarkdown` case covers nested bullet lists without extra trailing blank line

### DIFF
--- a/integrations/jira/markdown.go
+++ b/integrations/jira/markdown.go
@@ -202,7 +202,9 @@ func renderADFNodes(sb *strings.Builder, nodes []any, listPrefix string) {
 				sb.WriteString(listPrefix + "- ")
 				renderADFListItem(sb, li, listPrefix)
 			}
-			sb.WriteString("\n")
+			if listPrefix == "" {
+				sb.WriteString("\n")
+			}
 
 		case "orderedList":
 			for idx, item := range content {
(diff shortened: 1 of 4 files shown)


acmacalister (approved):
Clean PR. The two-pass buffering + TrimRight fix for Confluence p-wrapped list items is exactly right, and the ADF listPrefix guard is minimal and correct. Builds clean, all tests pass, lint is green.

PR #166, opened 2026-05-15
Add Stripe integration

daltoniam:
## Summary

- Adds a native Go integration adapter for the Stripe API with ~75 tools across balance, customers, charges, payment intents, refunds, disputes, payouts, subscriptions, products, prices, invoices, payment methods, setup intents, coupons, promotion codes, and events.
- Handles Stripe's distinct conventions: `application/x-www-form-urlencoded` request bodies with bracket-notation flattening, optional `Stripe-Account` Connect header, and `Retry-After`-aware `RetryableError` on 429/5xx.
- Configurable via `STRIPE_API_KEY` (required) and `STRIPE_ACCOUNT` (optional Connect header).

## Test plan

- [x] `make ci` passes (build, vet, test-race, lint)
- [x] Unit tests cover form encoder, HTTP helpers (success/error/retryable/204), dispatch parity, compaction-spec parity, and handler smoke tests via `httptest`
- [x] Live verification against the real Stripe test API: 18/18 tool calls succeeded end-to-end including create/retrieve/update/search/delete on customers, products, prices, payment intents, and pagination on every list endpoint
- [x] `Healthy()` ping verified against `/balance`

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -140,8 +140,9 @@ var envMapping = map[string]map[string]string{
 		"token":    "AGENTS_TOKEN",
 	},
 	"stripe": {
-		"api_key": "STRIPE_API_KEY",
-		"account": "STRIPE_ACCOUNT",
+		"api_key":  "STRIPE_API_KEY",
+		"account":  "STRIPE_ACCOUNT",
+		"base_url": "STRIPE_BASE_URL",
 	},
 }
 
--- a/config/config.go
+++ b/config/config.go
@@ -139,6 +139,10 @@ var envMapping = map[string]map[string]string{
 		"a2a_url":  "AGENTS_A2A_URL",
 		"token":    "AGENTS_TOKEN",
 	},
+	"stripe": {
+		"api_key": "STRIPE_API_KEY",
+		"account": "STRIPE_ACCOUNT",
+	},
 }
 
 // EnvMapping returns the env var mapping table. Useful for documentation and debugging.
@@ -246,6 +250,10 @@ func defaultConfig() *mcp.Config {
 			"ynab": {
 				Enabled:     false,
 				Credentials: mcp.Credentials{"api_key": ""},
+			},
+			"stripe": {
+				Enabled:     false,
+				Credentials: mcp.Credentials{"api_key": "", "account": "", "base_url": ""},
 			},
 			"jira": {
 				Enabled:     false,
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -33,7 +33,7 @@ func TestLoad_CreatesDefaultWhenMissing(t *testing.T) {
 	assert.NoError(t, err)
 
 	assert.Len(t, m.cfg.Integrations, 36)
-	for _, name := range []string{"github", "datadog", "linear", "sentry", "slack", "metabase", "aws", "posthog", "postgres", "clickhouse", "elasticsearch", "pganalyze", "rwx", "gmail", "notion", "ollama", "ynab", "gcp", "suno", "amazon", "jira", "confluence", "salesforce", "cloudflare", "digitalocean", "fly", "snowflake", "acp", "web", "botidentity", "x", "signoz", "nomad", "switchboard"} {
+	for _, name := range []string{"github", "datadog", "linear", "sentry", "slack", "metabase", "aws", "posthog", "postgres", "clickhouse", "elasticsearch", "pganalyze", "rwx", "gmail", "notion", "ollama", "ynab", "stripe", "gcp", "suno", "amazon", "jira", "confluence", "salesforce", "cloudflare", "digitalocean", "fly", "snowflake", "acp", "web", "botidentity", "x", "signoz", "nomad", "agents", "switchboard"} {
 		ic, ok := m.cfg.Integrations[name]
 		assert.True(t, ok, "missing default integration: %s", name)
 		assert.False(t, ic.Enabled)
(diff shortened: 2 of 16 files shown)


config/config.go
```diff
@@ -247,6 +251,10 @@ func defaultConfig() *mcp.Config {
 				Enabled:     false,
 				Credentials: mcp.Credentials{"api_key": ""},
 			},
+			"stripe": {
+				Enabled:     false,
+				Credentials: mcp.Credentials{"api_key": "", "account": "", "base_url": ""},
```
acmacalister, line 257: `base_url` appears in the stripe `defaultConfig()` entry here but has no matching env var in `envMapping["stripe"]` above. Every other integration with a `base_url` credential maps it to an env var (`PGANALYZE_APP_URL`, `SIGNOZ_BASE_URL`, etc.) — so `STRIPE_BASE_URL` will silently do nothing if a user tries it.

If `base_url` is production-useful (e.g. for Stripe-compatible APIs or stripe-mock), add it to envMapping:

```go
"stripe": {
    "api_key": "STRIPE_API_KEY",
    "account": "STRIPE_ACCOUNT",
    "base_url": "STRIPE_BASE_URL",
},
```

If it's purely an internal test escape hatch, drop it from `defaultConfig` so it doesn't show up in users' config files.

config/config_test.go
```diff
@@ -32,7 +32,7 @@ func TestLoad_CreatesDefaultWhenMissing(t *testing.T) {
 	_, err = os.Stat(path)
 	assert.NoError(t, err)
 
-	assert.Len(t, m.cfg.Integrations, 35)
+	assert.Len(t, m.cfg.Integrations, 36)
```
acmacalister, line 35: Two coverage gaps left after the count bumps:

1. `"stripe"` is absent from the name-check slice on the next line — so if the stripe entry ever disappears from `defaultConfig`, this test won't catch it.

2. `TestEnvMapping_ReturnsMapping` (further down) bumps the count from 25 → 26 but adds no assertions for the new keys. A typo in the env var names would go undetected.

Would be good to add both in the same pass:

```go
// In TestLoad_CreatesDefaultWhenMissing name slice:
"...", "switchboard", "stripe",

// In TestEnvMapping_ReturnsMapping:
assert.Equal(t, "STRIPE_API_KEY", m["stripe"]["api_key"])
assert.Equal(t, "STRIPE_ACCOUNT", m["stripe"]["account"])
```

acmacalister (approved):
Solid addition — the form encoder, retry logic, compaction specs, and handler tests are all well-crafted. Two small things inline, both in config_test.go / config.go.

config/config.go
daltoniam: Good catch. Added `STRIPE_BASE_URL` to `envMapping`, documented it in the README, and asserted it in `TestEnvMapping_ReturnsMapping`. Use case is stripe-mock / test escape hatch. Fixed in c1976f9.

config/config_test.go
daltoniam: Both fixed in c1976f9. Added `stripe` to the name-check slice (plus the pre-existing missing `agents`, which had the same gap) and added explicit `STRIPE_API_KEY`/`STRIPE_ACCOUNT`/`STRIPE_BASE_URL` assertions in `TestEnvMapping_ReturnsMapping`.

integrations/stripe/stripe.go
```diff
@@ -0,0 +1,229 @@
...
+
+	fullURL := s.baseURL + path
+	var bodyReader io.Reader
+	switch method {
+	case http.MethodGet, http.MethodDelete:
+		if form != "" {
+			if strings.Contains(fullURL, "?") {
+				fullURL += "&" + form
+			} else {
+				fullURL += "?" + form
```
acmacalister, line 89: The DELETE branch appends params to the query string, but `cancelSubscription` passes `invoice_now`, `prorate`, and `cancellation_details` through this path expecting them to reach Stripe as form-encoded body params. The Stripe API documents those as body parameters for `DELETE /v1/subscriptions/{id}` — sending them as query string values isn't the documented path and Stripe may silently ignore them.

One fix is to treat DELETE the same as POST when there's a body:

```go
default: // POST and DELETE-with-body
	bodyReader = strings.NewReader(form)
```

or add a `delWithBody` helper alongside `del` for the cancel case. Either way, `TestCancelSubscription_UsesDELETE` should gain a case that passes `invoice_now: true` and asserts the param appears in the request body, not the URL.

acmacalister (commented):
Solid integration — 75 tools, full dispatch parity, RetryableError plumbing, LimitReader, and a genuinely thorough form-encoder test suite. One thing to look at inline: the DELETE/body routing for the subscription cancel optional params.

integrations/stripe/stripe.go
daltoniam: Fixed in 8192180. Changed `doRequest` so only GET puts params on the query string; POST and DELETE now both send form-encoded body when params are non-empty. Existing DELETE-without-body callers (`deleteCustomer`, `deleteCoupon`, `deleteInvoice`, `deleteProduct`) are unaffected because `encodeForm(nil)` returns `""` and no body is sent. Updated `TestCancelSubscription_UsesDELETE` to pass `invoice_now: true` and assert it appears in the form-encoded request body (not the URL query string), plus assert the Content-Type header is set.

integrations/stripe/tools.go
```diff
@@ -0,0 +1,770 @@
...
+		Required:    []string{"id"},
+	},
+
+	// ── Products ─────────────────────────────────────────────────────
+	{
+		Name:        mcp.ToolName("stripe_list_products"),
+		Description: "List products (goods or services sold on Stripe). Use for product catalog browsing and finding SKUs.",
+		Parameters: mergeParams(listParams, map[string]string{
+			"active":    "Filter by active (true/false)",
+			"ids":       "Comma-separated product IDs",
```
acmacalister, line 405: The description says "Comma-separated product IDs" but `listParamsFrom` passes the value straight to `encodeForm`, which encodes an array as `ids[0]=prod_1&ids[1]=prod_2`. If an LLM follows the description and passes the string `"prod_1,prod_2"`, `encodeForm` encodes it as `ids=prod_1%2Cprod_2` — a single URL-encoded value that Stripe's API won't parse as an array filter.

Compare with the `types` param in `stripe_list_events` which correctly says `"Array of event types"`. This one should match:

```go
"ids": "Array of product IDs to filter by",
```

`TestFieldCompactionSpecs_ListShapeIncludesPaginationFields` won't catch this, so a handler test that passes `ids` as `[]string{"prod_1", "prod_2"}` and asserts `ids[0]` appears in the query string would lock it down.

acmacalister (approved):
Really solid integration — 75 tools, thorough compaction specs, form encoder handles all the Stripe bracket-notation edge cases correctly, and the DELETE-with-body fix is clean. One small description mismatch inline that could trip up LLM callers.

PR #167, opened 2026-05-15
Add "Context window saved by Switchboard" dashboard hero

daltoniam:
CleanShot 2026-05-15 at 15 34 23@2x

## Summary

- New dashboard hero card frames Switchboard's value-prop as **bytes/tokens of LLM context never sent**, broken out into four buckets (tool catalog, response compaction, markdown rendering, script intermediates)
- Lifetime counters persist to disk (dirty-flag flusher + shutdown flush) so savings survive restarts
- Opt-in dollar-cost estimate with configurable input $/MTok rate, toggled from Settings
- Compaction crediting now includes columnar-reshape byte savings (not just `CompactSpec` matches), so tabular responses like `snowflake_execute_query` register their real reduction

## How it's measured

| Bucket | Source |
|---|---|
| Tool catalog | `tools/list` JSON size at startup minus the columnarized search-result size, recorded per search |
| Response compaction | `len(original) - len(post-pipeline JSON)` after `CompactAny` + `ColumnarizeAny` |
| Markdown rendering | `len(raw) - len(rendered markdown)` for tools implementing `MarkdownIntegration` |
| Script intermediates | sum of `api.call()` payload sizes minus the script's final return value (clamped ≥ 0) |

Tokens = bytes / 4 (configurable via `CharsPerToken`). Dollars = tokens × `DollarsPerMTokInput` / 1M.

## Test plan

- [x] `make build` clean
- [x] `make vet` / `make lint` clean (0 issues)
- [x] `go test -race ./...` passes for all touched packages (server, web, script, metrics, config)
- [x] Verified live in production: hero renders with 62.1K tokens / 242.5KB saved on first search
- [x] Compaction bucket now credits columnarization (verified by removing the `compacted &&` gate)
- [ ] Verify persistence survives a restart (open dashboard, restart server, reopen — counters intact)
- [ ] Toggle dollar estimate in Settings, confirm value appears/disappears on dashboard
- [ ] Trigger a script with hidden intermediates to confirm Script Intermediates bucket fires


server/server.go
```diff
@@ -741,6 +790,14 @@ func (s *Server) handleScriptExecute(ctx context.Context, source string) (*mcpsd
 	if err != nil {
 		return errorResult(err.Error()), nil
 	}
+	// Record script byte flow regardless of error state: even an errored
+	// script may have already issued api.call() invocations whose bytes
+	// stayed server-side. The "final" portion is the actual ToolResult.Data
+	// the LLM ends up seeing (post-columnarization, computed below for the
+	// happy path; pre-columnarization is close enough for error/empty cases).
+	if s.services.Metrics != nil {
+		s.services.Metrics.RecordScriptSavings(result.IntermediateBytes, result.FinalBytes)
```
acmacalister: `result.FinalBytes` here is the pre-columnarization size set by the script engine. The comment above says "post-columnarization, computed below for the happy path" but that computation never actually happens — `columnarizeResult` runs a few lines later without updating `FinalBytes`. For scripts returning tabular data, columnarization can cut output by 30-50%, so the reported script savings will consistently understate reality.

The fix is to move metric recording after columnarization for the non-error path:

```go
if result.IsError {
    if s.services.Metrics != nil {
        s.services.Metrics.RecordScriptSavings(result.IntermediateBytes, result.FinalBytes)
    }
    return &mcpsdk.CallToolResult{...}, nil
}
result.Data = columnarizeResult(result.Data)
if s.services.Metrics != nil {
    s.services.Metrics.RecordScriptSavings(result.IntermediateBytes, int64(len(result.Data)))
}
```

web/templates/pages/dashboard.templ
```diff
@@ -109,6 +109,37 @@ func formatBytes(b int64) string {
...
+		return fmt.Sprintf("%.1f%%", pct)
+	}
+	return fmt.Sprintf("%.0f%%", pct)
+}
+
+func hasSavings(snap *mcp.MetricsSnapshot) bool {
+	if snap == nil {
+		return false
+	}
+	return snap.TotalBytesSaved > 0 || snap.CatalogAvoidedCount > 0 || snap.CompactionSamples > 0 || snap.MarkdownSamples > 0 || snap.ScriptSavingsSamples > 0
```
acmacalister: The `|| snap.ScriptSavingsSamples > 0` clause can surface the hero card with "0 tokens saved" when a script runs but makes no `api.call()` invocations (e.g., a pure-computation script). In that case `ScriptIntermediateHidden` is 0 and all the other buckets are also zero, so the headline reads "0 tokens". Showing "0 tokens saved by Switchboard" as the hero is worse UX than hiding it.

Worth considering whether to gate the sample-count checks on the corresponding byte total being positive, or just drop `ScriptSavingsSamples > 0` from the condition (the `TotalBytesSaved > 0` guard already covers the real savings case).

acmacalister (approved):
Clean implementation overall — the four-bucket framing is intuitive, persistence is wired correctly with the dirty-flag flusher and shutdown defer, and the test coverage is solid. Two things inline worth a second look.

metrics.go
```diff
@@ -315,3 +506,180 @@ type ToolRank struct {
...
+	}
+	tmp := m.filePath + ".tmp"
+	if err := os.WriteFile(tmp, data, 0600); err != nil {
+		return fmt.Errorf("write metrics tmp: %w", err)
+	}
+	if err := os.Rename(tmp, m.filePath); err != nil {
+		_ = os.Remove(tmp)
+		return fmt.Errorf("rename metrics: %w", err)
+	}
+	m.dirty.Store(false)
```
acmacalister, line 650: The dirty flag is cleared AFTER reading all the atomics, which opens a small but real race window. If a `Record*` call increments a counter AND sets `dirty=true` between the last `.Load()` on line ~615 and this `Store(false)`, the `Store(false)` overwrites that record's `true`. The result is `dirty=false` with in-memory totals one increment ahead of what's on disk — if the server restarts before the next `Record*` fires to re-set dirty, that increment is gone.

The fix is to swap dirty to `false` BEFORE reading the atomics, so any racing records will re-set it and be picked up on the next flush:

```go
func (m *Metrics) Flush() error {
    if m.filePath == "" {
        return nil
    }
    if !m.dirty.Swap(false) { // clear before snapshot
        return nil
    }
    p := persistedMetrics{ /* read atomics here */ }
    ...
    if err := os.Rename(tmp, m.filePath); err != nil {
        _ = os.Remove(tmp)
        m.dirty.Store(true) // write failed, ensure retry
        return fmt.Errorf("rename metrics: %w", err)
    }
    // dirty already cleared by Swap above — don't Store(false) again
    return nil
}
```

Low-probability in practice since records fire constantly, but for a persistence path it's worth getting right.

acmacalister (commented):
Really solid feature — four-bucket accounting is cleanly structured, test coverage is excellent (especially the persistence round-trip and hero gate tests), and the atomic write pattern in Flush is the right approach. One concurrency nit inline.

PR #170, opened 2026-05-18
Add Vercel integration

daltoniam:
## Summary

- Add a raw HTTP Vercel adapter for teams, projects, deployments, logs/events, environment variables, domains, and aliases.
- Wire Vercel into server registration, default config/env var overlays, response compaction, tests, and README setup docs.

## Test plan

- [x] `go test ./integrations/vercel ./config ./cmd/server`
- [x] `go test ./compact ./server ./integrations/vercel ./config ./cmd/server`
- [x] `go tool golangci-lint run ./integrations/vercel ./config ./cmd/server`
- [ ] `make ci` currently blocked by pre-existing `browser.TestPage_Evaluate` mismatch: expected `float64(2)`, actual `int(2)`.

💘 Generated with Crush


arjansingh (approved):

integrations/vercel/vercel.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+	if resp.StatusCode == http.StatusTooManyRequests || resp.StatusCode >= 500 {
+		re := &mcp.RetryableError{StatusCode: resp.StatusCode, Err: fmt.Errorf("vercel API error (%d): %s", resp.StatusCode, vercelErrorMessage(data))}
+		re.RetryAfter = mcp.ParseRetryAfter(resp.Header.Get("Retry-After"))
+		return nil, re
+	}
+	if resp.StatusCode >= 400 {
+		return nil, fmt.Errorf("vercel API error (%d): %s", resp.StatusCode, vercelErrorMessage(data))
+	}
+	if resp.StatusCode == http.StatusNoContent || resp.StatusCode == http.StatusAccepted || len(data) == 0 {
```
acmacalister: This bundles `http.StatusAccepted` (202) with `http.StatusNoContent` (204) when deciding to discard the response body. Every other adapter in the codebase (`gmail`, `notion`, `posthog`, `signoz`) only guards on 204:

```go
if resp.StatusCode == 204 || len(data) == 0 {
```

202 semantically means "accepted for async processing" and can carry a useful body — e.g. a job ID or a polling URL. If a Vercel endpoint ever returns 202 with payload, this silently replaces it with `{"status":"success"}`. Suggest dropping the 202 arm and relying on `len(data) == 0` for the bodyless case:

```go
if resp.StatusCode == http.StatusNoContent || len(data) == 0 {
    return json.RawMessage(`{"status":"success"}`), nil
}
```

integrations/vercel/teams.go
```diff
@@ -0,0 +1,110 @@
...
+		return mcp.ErrResult(err)
+	}
+	params["teamId"] = teamID
+	params["slug"] = teamSlug
+	params = v.scopedQuery(params)
+	if params["teamId"] == "" && params["slug"] == "" {
+		return mcp.ErrResult(fmt.Errorf("team_id or team_slug is required"))
+	}
+	if params["teamId"] != "" {
+		data, err := v.get(ctx, "/v3/teams/%s/members%s", url.PathEscape(params["teamId"]), queryEncode(params))
```
acmacalister: `params` already has `teamId` set (line 64), and `queryEncode(params)` will include it in the query string too. So the request URL becomes `/v3/teams/team_123/members?teamId=team_123&limit=20&...` — the team ID ends up in both the path segment and the query. The Vercel API ignores the duplicate, but it's cleaner to strip path params from the query map before encoding:

```go
teamIDPath := params["teamId"]
if teamIDPath == "" {
    teamIDPath = params["slug"]
}
delete(params, "teamId")
delete(params, "slug")
data, err := v.get(ctx, "/v3/teams/%s/members%s", url.PathEscape(teamIDPath), queryEncode(params))
```

Same applies to the `slug` branch on line 77.

acmacalister (commented):
Build passes, tests green with the race detector, lint clean. Two things worth a look inline — one pattern deviation and one minor redundancy.

integrations/vercel/vercel_test.go
```diff
@@ -0,0 +1,451 @@
...
+		var body map[string]string
+		require.NoError(t, json.NewDecoder(r.Body).Decode(&body))
+		assert.Equal(t, "app.example.com", body["alias"])
+		_, _ = w.Write([]byte(`{"alias":"app.example.com"}`))
+	})
+	result, err := v.Execute(context.Background(), "vercel_assign_deployment_alias", map[string]any{"deployment_id": "dpl_1", "alias": "app.example.com"})
+	require.NoError(t, err)
+	assert.False(t, result.IsError)
+	assert.Contains(t, result.Data, "app.example.com")
+}
```
acmacalister, line 435: 14 of the 26 handlers still have no execution test: `listTeams`, `getTeam`, `listUserEvents`, `createProject`, `updateProject`, `deleteProject`, `deleteDeployment`, `listRuntimeLogs`, `updateProjectEnvVar`, `deleteProjectEnvVar`, `removeProjectDomain`, `getDomainConfig`, `listDeploymentAliases`, and `deleteDeploymentAlias`. The dispatch parity tests confirm they're wired, but they can't catch URL path typos (`/v9/projects` instead of `/v10/projects`) or wrong query param names that only show up at runtime. They all follow the same shape as the tests above, so adding an `httptest.NewServer` for each is straightforward.

acmacalister (approved):
Clean implementation overall — the scoping abstraction, retryable error handling, and 10MB response cap are all done well. The two fixes from the previous round land correctly and each has a targeted regression test to prove it. One suggestion inline about handler coverage.

integrations/vercel/compact.yaml
```diff
@@ -0,0 +1,215 @@
...
+            - '[].payload.text'
+            - '[].payload.id'
+            - '[].payload.deploymentId'
+            - '[].payload.source'
+            - '[].payload.level'
+            - '[].payload.statusCode'
+    vercel_list_runtime_logs:
+        max_bytes: 131072
+        spec:
+            - logs[].level
```
acmacalister, line 172: The spec here uses `logs[].field` paths (expecting `{"logs": [...], "pagination": {...}}`), but `TestListRuntimeLogs` returns a bare array `[{"id":"log_1","message":"ready"}]`. Compare with `vercel_list_deployment_events` just above, which correctly uses `'[].type'` for a bare-array response.

If the Vercel runtime logs API returns an object wrapper, the test mock needs to match:
```go
_, _ = w.Write([]byte(`{"logs":[{"rowId":"log_1","message":"ready","level":"info"}],"pagination":{"count":1}}`))
```
If it returns a bare array, the spec should align with the events spec instead:
```yaml
vercel_list_runtime_logs:
    spec:
        - '[].level'
        - '[].message'
        - '[].rowId'
        ...
```
Either way one of them has the wrong shape — and the `id` field in the mock vs `rowId` in the spec adds to the inconsistency.

acmacalister (approved):
Solid integration — build, tests, and lint all clean. All three previous review comments are nicely resolved in the latest commit. One thing to double-check inline.

acmacalister (approved):
All four previous review concerns are fully resolved — the 202/204 response fix, the listTeamMembers path dedup, the runtime logs spec/test alignment, and the missing execution tests are all addressed. Build passes, full test suite passes (race detector clean), lint clean, go.mod tidy. The scopedQuery helper is a nice touch for keeping team scoping DRY across 26 handlers. LGTM.

PR #168, opened 2026-05-17
chore(deps): bump modelcontextprotocol/go-sdk to v1.5.0

daltoniam:
## Summary

- Bumps `github.com/modelcontextprotocol/go-sdk` from **v1.4.1 → v1.5.0** for the stabilized client-side OAuth APIs, protocol version `2025-11-25`, Origin/Content-Type header verification, tool-input-validation errors returned as tool results, and data-race fixes.
- Transitively bumps `golang.org/x/oauth2` 0.34.0 → 0.35.0.
- None of v1.5.0's breaking changes affect us — confirmed via grep. We don't touch `auth.AuthorizationCodeHandlerConfig`, `PreregisteredClientConfig`, `AuthorizationCodeFetcher`, the `oauthex` package, or the now-removed `mcp_go_client_oauth` build tag. We only use the `mcp` client/server surface, unchanged in v1.5.0.

## Test plan

- [x] `go build ./...`
- [x] `go test ./...` — all suites green except a pre-existing `browser.TestPage_Evaluate` failure on `main` (chromedp JS-number type assertion, unrelated to this bump).
- [ ] CI green

💘 Generated with Crush

PR #171, opened 2026-05-18
Add Google Workspace integration suite

daltoniam:
## Summary

- Adds 11 Google Workspace integrations (`gcal`, `gchat`, `gdocs`, `gdrive`, `gforms`, `gmeet`, `gpeople`, `gsheets`, `gslides`, `gtasks`) — taking the integration count from 35 to 46.
- Extracts a shared `googleoauth/` package (PKCE flow, token refresh, in-progress flow registry, polling) so each adapter is ~250 lines lighter; `gmail` is migrated onto it with no behavior change.
- Each adapter ships with dispatch map, `compact.yaml`, markdown renderers for document-shaped responses, OAuth wrapper with the right scope set, and table-driven tests (≥75% coverage; most >85%).
- Web UI: 11 new `/integrations/<name>/setup` pages plus full OAuth start/callback/poll/save plumbing in `web/web.go`.

### Automated

- [x] `make ci` passes (build, vet, race tests, lint, security). Only failure is the pre-existing `TestPage_Evaluate` float64-vs-int issue in `browser/` which also fails on clean `main`.
- [x] Dispatch map parity tests (`TestDispatchMap_AllToolsCovered` / `TestDispatchMap_NoOrphanHandlers`) in every new adapter.
- [x] Compaction parity tests in every new adapter.
- [x] Cross-adapter strict-mode compaction check (`compact/all_adapters_test.go`).
- [x] `TestNewArgs_ErrCheckParity` — auto-covers all new adapters.
- [x] Markdown parity (`TestRenderMarkdown_ToolsCovered`) for all adapters that return document content.

### Manual

1. **OAuth setup** (per integration)
   - `make build && ./switchboard` then open the web UI.
   - Visit `/integrations/<name>/setup` (e.g. `/integrations/gcal/setup`).
   - Paste a Google Cloud OAuth Client ID + Secret with the redirect URI shown on the page, save, then **Connect**.
   - Confirm: redirect to Google consent → granted scopes match what the page lists → status badge turns green → ``Healthy`` probe succeeds.
2. **End-to-end via MCP** (recommended path — exercises search, compaction, scripts together)
   - Point any MCP client at the running server.
   - For each integration, run a smoke sequence, e.g.:
     - `gcal`: `gcal_list_calendars` → `gcal_list_events` on the primary calendar → `gcal_create_event` → `gcal_delete_event`.
     - `gdrive`: `gdrive_list_files` → `gdrive_get_file_metadata` → `gdrive_export_file` on a Doc.
     - `gdocs`: `gdocs_get_document` on the Doc above; verify markdown rendering of headings, lists, tables.
     - `gsheets`: `gsheets_get_spreadsheet` → `gsheets_values_get` → `gsheets_values_update`.
     - `gmail`: regression — `gmail_list_messages` + `gmail_get_message` (refactor onto `googleoauth` must not change behavior).
     - `gmeet`: `gmeet_create_space` → `gmeet_get_space` (with meeting code) → `gmeet_list_conference_records`.
     - `gchat`: `gchat_list_spaces` → `gchat_create_message`.
     - `gforms`/`gslides`/`gtasks`/`gpeople`: equivalent list+get smoke.
   - The `mcp-benchmark` skill can drive this systematically (run "benchmark gcal", etc.).
3. **Token refresh** — wait until the access token expires (or shorten it), reissue any call, confirm one silent refresh occurs (each adapter has a `TestTokenRefreshOnUnauthorized` test that asserts this, but worth eyeballing in logs).
4. **Search discovery** — in an MCP client, run a short query like ``{"query":"create calendar event"}`` and confirm `gcal_create_event` is in the top hits; repeat for a few other natural-language phrasings ("send chat message", "list drive files", "new meet space").

### Notes / caveats

- This branch was cut from `208e695`; `main` has since advanced (e.g. SDK bump in #168). A merge or rebase before merging is fine — there are no expected conflicts in adapter code.
- The pre-existing `browser/TestPage_Evaluate` failure is not addressed here.

🤖 Generated with Claude Code (Sonnet 4.5)


integrations/gdrive/gdrive.go
```diff
@@ -0,0 +1,362 @@
...
+			g.mu.Lock()
+			if g.accessToken == currentToken {
+				g.accessToken = newToken
+				g.persistToken(newToken)
+			}
+			g.mu.Unlock()
+			return g.doRawInner(ctx, method, fullURL, false)
+		}
+	}
+	if resp.StatusCode >= 400 {
```
acmacalister, line 259: `doRawInner` collapses all 4xx/5xx into a plain `fmt.Errorf`, but `doRequestInner` (line 201) correctly wraps 429 and 5xx in `mcp.RetryableError` so the server's retry middleware can act on them. Since `gdrive_download_file` and `gdrive_export_file` both go through `doRaw → doRawInner`, rate-limit and server errors on those paths will surface as plain errors and never get retried.

```go
// replace the single check with the same pattern doRequestInner uses:
if resp.StatusCode == 429 || resp.StatusCode >= 500 {
    re := &mcp.RetryableError{
        StatusCode: resp.StatusCode,
        Err:        fmt.Errorf("gdrive API error (%d): %s", resp.StatusCode, string(data)),
    }
    re.RetryAfter = mcp.ParseRetryAfter(resp.Header.Get("Retry-After"))
    return nil, "", re
}
if resp.StatusCode >= 400 {
    return nil, "", fmt.Errorf("gdrive API error (%d): %s", resp.StatusCode, string(data))
}
```

integrations/gcal/events.go
```diff
@@ -0,0 +1,306 @@
...
+	if rules := r.StrSlice("recurrence"); len(rules) > 0 {
+		body["recurrence"] = rules
+	}
+	if v := r.Str("reminders_use_default"); v != "" {
+		body["reminders"] = map[string]any{"useDefault": v == "true"}
+	}
+	if v := r.Str("create_meet"); v == "true" {
+		body["conferenceData"] = map[string]any{
+			"createRequest": map[string]any{
+				"requestId":             "switchboard-" + r.Str("summary"),
```
acmacalister: Google Calendar treats `requestId` as an idempotency key for conference data creation — the API will silently reuse the Meet link from the first request that carried this ID rather than provision a new room. Two events with the same title (or both with no title at all, giving `"switchboard-"`) will collide: the second event ends up sharing the first event's conference.

`github.com/google/uuid` is already in `go.mod`, so a one-liner avoids the collision:

```go
import "github.com/google/uuid"

// ...
"requestId": "switchboard-" + uuid.New().String(),
```

googleoauth/oauth.go
```diff
@@ -0,0 +1,329 @@
...
+	data := url.Values{
+		"code":          {code},
+		"redirect_uri":  {f.redirectURI},
+		"client_id":     {f.clientID},
+		"client_secret": {f.clientSecret},
+		"code_verifier": {f.codeVerifier},
+		"grant_type":    {"authorization_code"},
+	}
+
+	req, err := http.NewRequest("POST", tokenURL, strings.NewReader(data.Encode()))
```
acmacalister: The token exchange fires via `http.DefaultClient` which has no timeout. If Google's token endpoint is slow, `HandleCallback` will hang here indefinitely while holding `f.mu`, which means any concurrent `Poll` call for the same integration will stall for the duration.

A 30-second deadline on the exchange itself covers the realistic worst case:

```go
tctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
defer cancel()
req, err := http.NewRequestWithContext(tctx, "POST", tokenURL, strings.NewReader(data.Encode()))
```

googleoauth/oauth.go
```diff
@@ -0,0 +1,329 @@
...
+// token via Google's token endpoint.
+func RefreshAccessToken(clientID, clientSecret, refreshToken string) (string, error) {
+	data := url.Values{
+		"client_id":     {clientID},
+		"client_secret": {clientSecret},
+		"refresh_token": {refreshToken},
+		"grant_type":    {"refresh_token"},
+	}
+
+	req, err := http.NewRequest("POST", tokenURL, strings.NewReader(data.Encode()))
```
acmacalister: Same no-timeout issue as `HandleCallback` above, but this one is on the hot path — every adapter calls `RefreshAccessToken` inline on a 401 response while the caller's `ctx` is available. Adding a `context.Context` first parameter and threading it through lets the adapter's existing request deadline apply:

```go
func RefreshAccessToken(ctx context.Context, clientID, clientSecret, refreshToken string) (string, error) {
    // ...
    req, err := http.NewRequestWithContext(ctx, "POST", tokenURL, strings.NewReader(data.Encode()))
```

All eleven call sites already have `ctx` in scope (`doRequestInner`, `doRawInner`), so the update is mechanical.

acmacalister (approved):
Really thorough PR — 11 adapters, a clean shared OAuth package, parity tests everywhere, and all CI checks green. A few non-blocking things inline worth a look before this sees heavy production use.

acmacalister (commented):
Solid PR — build passes, `go test -race ./...` is green across all 61 packages, and golangci-lint is clean. The `googleoauth/` extraction is the right design call and the 11 adapters are consistent and well-tested.

Agreeing with all three of @acmacalister's inline findings — the `doRawInner` missing `RetryableError` is the most important to fix before merge since it silently swallows rate-limit signals on download/export paths. The two timeout issues in `googleoauth` are straightforward to close as well.

One additional "consider" I noted: the new adapters don't have per-adapter `compact_specs_test.go` files verifying that `viewSets` is populated for the tools that declare views — consistent with `gmail` today, but AGENTS.md lists this as a MUST. Worth a quick follow-up ticket rather than blocking this PR.

acmacalister (approved):
The three issues from the previous round are all addressed in ed13cd4 — doRawInner now wraps 429/5xx in RetryableError, HandleCallback has a 30s timeout, and RefreshAccessToken threads context through. Build, tests (race), and lint all pass clean. The googleoauth extraction is a nice refactor and the 11 new adapters follow every established convention. LGTM.

integrations/gchat/gchat.go
```diff
@@ -0,0 +1,267 @@
...
+		return false
+	}
+	req.Header.Set("Authorization", "Bearer "+g.accessToken)
+	resp, err := g.client.Do(req)
+	if err != nil {
+		return false
+	}
+	defer func() { _ = resp.Body.Close() }()
+	_, _ = io.Copy(io.Discard, io.LimitReader(resp.Body, 4096))
+	return resp.StatusCode != 401 && resp.StatusCode != 403
```
acmacalister: `gcal` and `gdrive` route their health probes through `g.get()`, which transparently attempts a token refresh on 401 before declaring failure. These eight adapters skip that path — once the access token expires (~60 min for Google tokens), they'll flip the health badge red even though a valid refresh token is present and every real API call would succeed after a silent refresh.

For adapters that have a cheap list endpoint (`gchat`, `gmeet`, `gtasks`, `gpeople`), the fix matches `gcal`:

```go
func (g *gchat) Healthy(ctx context.Context) bool {
    _, err := g.get(ctx, "/spaces?pageSize=1")
    return err == nil
}
```

For adapters that use a sentinel probe ID because there's no list endpoint (`gdocs`, `gforms`, `gsheets`, `gslides`) — where a 404 means "auth is valid, probe doc doesn't exist" — you'd want:

```go
func (g *gdocs) Healthy(ctx context.Context) bool {
    _, err := g.get(ctx, "/documents/HEALTHCHECK_PROBE")
    return err == nil || strings.Contains(err.Error(), "(404)")
}
```

Same pattern applies to `gforms`, `gsheets`, and `gslides`.

acmacalister (approved):
Solid PR — the googleoauth refactor is clean and all three previous review items have been addressed. One pattern worth fixing before merge: the Healthy implementation in 8 of the 10 new adapters won't attempt a token refresh on 401, so the health badge will go red every hour even when the integration is working fine. Left a note inline.

integrations/gcal/gcal_test.go
```diff
@@ -0,0 +1,589 @@
...
+	assert.False(t, result.IsError)
+	assert.Equal(t, `{"id":"ev1"}`, result.Data)
+}
+
+func TestErrResult(t *testing.T) {
+	result, err := mcp.ErrResult(fmt.Errorf("test error"))
+	require.NoError(t, err)
+	assert.True(t, result.IsError)
+	assert.Equal(t, "test error", result.Data)
+}
```
acmacalister, line 604: Both `gcal` and `gdrive` are missing all `TestHealthy_*` tests — `gcal_test.go` has none and `gdrive_test.go` has none either. Every other new adapter has at least `TestHealthy_TrueOn200/404` and `TestHealthy_FalseOn401`.

The implementation is correct (`Healthy` routes through `g.get()` which handles refresh), but without tests a future refactor could silently break it. The pattern from `gchat_test.go` is a direct template:

```go
func TestHealthy_TrueOn200(t *testing.T) {
	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
		_, _ = w.Write([]byte(`{"kind":"calendar#calendarList"}`))
	}))
	defer ts.Close()
	g := &gcal{accessToken: "tok", client: ts.Client(), baseURL: ts.URL}
	assert.True(t, g.Healthy(context.Background()))
}

func TestHealthy_FalseOn401(t *testing.T) {
	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
		w.WriteHeader(401)
	}))
	defer ts.Close()
	g := &gcal{accessToken: "bad", client: ts.Client(), baseURL: ts.URL}
	assert.False(t, g.Healthy(context.Background()))
}
```

And a `TestHealthy_TrueAfterRefresh` mirroring the one in `gchat_test.go` would cover the refresh path.

integrations/gforms/gforms_test.go
```diff
@@ -0,0 +1,413 @@
...
+
+func TestHealthy_FalseOn401(t *testing.T) {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(401)
+	}))
+	defer ts.Close()
+
+	g := &gforms{accessToken: "bad", client: ts.Client(), baseURL: ts.URL}
+	assert.False(t, g.Healthy(context.Background()))
+}
```
acmacalister, line 379: Six adapters (`gforms`, `gmeet`, `gpeople`, `gsheets`, `gslides`, `gtasks`) have `TestHealthy_TrueOn200/404` and `TestHealthy_FalseOn401` but not a `TestHealthy_TrueAfterRefresh`. The previous review called out that health probes weren't routing through the refresh path, and that was fixed by using `g.get()` — but only `gchat` and `gdocs` got the corresponding regression test at the same time.

A `TestHealthy_TrueAfterRefresh` for each adapter (same structure as the one in `gchat_test.go`) would lock in the fix and prevent a silent regression if the `Healthy` implementation is ever changed.

acmacalister (approved):
Solid PR — build clean, all tests pass, lint zero. The fixes from the previous round (uuid requestId, HandleCallback timeout, RefreshAccessToken ctx, health probes through g.get()) are all in place. Two small test coverage gaps inline.

integrations/gcal/gcal_test.go
```diff
@@ -0,0 +1,653 @@
...
+
+func TestTools_NoDuplicateNames(t *testing.T) {
+	i := New()
+	seen := make(map[mcp.ToolName]bool)
+	for _, tool := range i.Tools() {
+		assert.False(t, seen[tool.Name], "duplicate tool name: %s", tool.Name)
+		seen[tool.Name] = true
+	}
+}
```
acmacalister, line 74: Eight of the ten new adapters (gchat, gdocs, gforms, gmeet, gpeople, gsheets, gslides, gtasks) have a `TestTools_EntryPointHasStartHere` test that asserts the primary tool description contains "Start here". `gcal` and `gdrive` are missing it — the text is in the descriptions today, but without the test a future edit that trims the phrase won't be caught by CI.

```go
func TestTools_EntryPointHasStartHere(t *testing.T) {
	i := New()
	for _, tool := range i.Tools() {
		if tool.Name == mcp.ToolName("gcal_list_events") {
			assert.Contains(t, tool.Description, "Start here",
				"entry-point tool gcal_list_events must include 'Start here' for wayfinding")
			return
		}
	}
	t.Fatal("gcal_list_events tool not found")
}
```

integrations/gdrive/gdrive_test.go
```diff
@@ -0,0 +1,740 @@
...
+}
+
+func TestTools_NoDuplicateNames(t *testing.T) {
+	i := New()
+	seen := make(map[mcp.ToolName]bool)
+	for _, tool := range i.Tools() {
+		assert.False(t, seen[tool.Name], "duplicate tool name: %s", tool.Name)
+		seen[tool.Name] = true
+	}
+}
```
acmacalister, line 76: Same gap as `gcal_test.go` — `gdrive_list_files` has "Start here" in its description but there is no `TestTools_EntryPointHasStartHere` to lock that in.

```go
func TestTools_EntryPointHasStartHere(t *testing.T) {
	i := New()
	for _, tool := range i.Tools() {
		if tool.Name == mcp.ToolName("gdrive_list_files") {
			assert.Contains(t, tool.Description, "Start here",
				"entry-point tool gdrive_list_files must include 'Start here' for wayfinding")
			return
		}
	}
	t.Fatal("gdrive_list_files tool not found")
}
```

acmacalister (approved):
Really solid PR — the shared `googleoauth` package is a clean extraction, all 11 adapters build cleanly, tests pass with the race detector, and lint is zero. Both issues from the previous review round are addressed. Just two small test gaps inline.

acmacalister (approved):
Thorough PR — builds, all tests pass with the race detector, lint clean, go mod tidy clean. All three previous review comments (doRawInner RetryableError, TestHealthy_TrueAfterRefresh, TestTools_EntryPointHasStartHere) are addressed. The googleoauth extraction is clean, every adapter has the required parity tests, and the gdocs compact.yaml's formats: [json, markdown] is the right approach for that structure. LGTM.

PR #172, opened 2026-05-18
Marketplace: rewrite GitHub HTML URLs to Contents API

daltoniam:
## Summary

- Manifest sources pasted as the familiar `https://github.com/{owner}/{repo}/blob/{ref}/manifest.json` HTML link now succeed instead of returning HTTP 404 / HTML — the URL is rewritten to `https://api.github.com/repos/{owner}/{repo}/contents/{path}?ref={ref}` and accompanied by `Accept: application/vnd.github.raw`.
- Combined with the existing `GitHubTokenFunc` wiring in `cmd/server/main.go` (which attaches the GitHub integration's PAT as a bearer token for GitHub hosts), **private-repo manifests and WASM downloads now work transparently** without any new UI or per-source token config.
- `raw.githubusercontent.com`, `api.github.com`, public repos, and non-GitHub URLs are unaffected.

## Why

Internal Curri plugin manifest at `teamcurri/switchboard_plugins` (private) couldn't be fetched: pasting the HTML `/blob/main/manifest.json` link returned a 404 even with the integration token configured, because GitHub serves HTML at that path. The fix normalizes that exact URL form into a Contents API request that the existing bearer-token logic already authenticates correctly.

## Implementation

- New `marketplace/github_url.go`: pure `normalizeGitHubURL(rawURL) → (effectiveURL, useRawAccept)`.
- `FetchManifest` and `downloadWasm` call the normalizer and conditionally set the `Accept: application/vnd.github.raw` header. Everything else (auth header, retries, caching) is unchanged.

## Test plan

- [x] `TestNormalizeGitHubURL` — 9 cases: blob/raw HTML, master/main branches, nested paths, raw.githubusercontent passthrough, API form passthrough, non-GitHub passthrough, `/releases/download/` passthrough, malformed URL passthrough.
- [x] `TestFetchManifest_RewritesBlobURL` — end-to-end: paste blob URL → server receives Contents API path + bearer + `application/vnd.github.raw` Accept.
- [x] `TestDownloadWasm_RewritesBlobURL` — same end-to-end check for WASM downloads.
- [x] `go test ./marketplace/... -race` passes; `go vet`, `go build ./...`, and `golangci-lint run ./marketplace/...` all clean.
- [x] Unrelated pre-existing `browser.TestPage_Evaluate` failure on `main` is **not** introduced by this change.

### DIFF
--- a/marketplace/github_url.go
+++ b/marketplace/github_url.go
@@ -38,10 +38,14 @@ func normalizeGitHubURL(rawURL string) (newURL string, useRawAccept bool) {
 	owner, repo, ref := parts[0], parts[1], parts[3]
 	path := strings.Join(parts[4:], "/")
 
-	api := &url.URL{
-		Scheme: "https",
-		Host:   "api.github.com",
-		Path:   "/repos/" + owner + "/" + repo + "/contents/" + path,
+	apiURL, err := url.JoinPath("https://api.github.com",
+		"repos", owner, repo, "contents", path)
+	if err != nil {
+		return rawURL, false
+	}
+	api, err := url.Parse(apiURL)
+	if err != nil {
+		return rawURL, false
 	}
 	q := api.Query()
 	q.Set("ref", ref)
--- a/marketplace/github_url.go
+++ b/marketplace/github_url.go
@@ -1,0 +1,51 @@
+package marketplace
+
+import (
+	"net/url"
+	"strings"
+)
+
+// normalizeGitHubURL rewrites GitHub HTML URLs (the kind a user copies from
+// the browser) into the GitHub Contents API form so the file's raw bytes can
+// be fetched with an Authorization header. This is the only way to retrieve
+// files from a private repository over HTTPS without using `git` itself.
+//
+// Supported input forms:
+//
+//   - https://github.com/{owner}/{repo}/blob/{ref}/{path}   →  Contents API
+//   - https://github.com/{owner}/{repo}/raw/{ref}/{path}    →  Contents API
+//
+// Any URL on raw.githubusercontent.com / api.github.com / non-GitHub hosts is
+// passed through unchanged. Malformed URLs are also passed through so the
+// caller can attempt the request and surface a normal HTTP error.
+//
+// useRawAccept is true when the rewritten URL is the Contents API form, in
+// which case the request must carry `Accept: application/vnd.github.raw` to
+// receive raw file bytes instead of the default JSON metadata wrapper.
+func normalizeGitHubURL(rawURL string) (newURL string, useRawAccept bool) {
+	u, err := url.Parse(rawURL)
+	if err != nil || u.Host == "" {
+		return rawURL, false
+	}
+	if strings.ToLower(u.Host) != "github.com" {
+		return rawURL, false
+	}
+	// /{owner}/{repo}/blob|raw/{ref}/{path...}
+	parts := strings.Split(strings.TrimPrefix(u.Path, "/"), "/")
+	if len(parts) < 5 || (parts[2] != "blob" && parts[2] != "raw") {
+		return rawURL, false
+	}
+	owner, repo, ref := parts[0], parts[1], parts[3]
+	path := strings.Join(parts[4:], "/")
+
+	api := &url.URL{
+		Scheme: "https",
+		Host:   "api.github.com",
+		Path:   "/repos/" + owner + "/" + repo + "/contents/" + path,
+	}
+	q := api.Query()
+	q.Set("ref", ref)
+	api.RawQuery = q.Encode()
+
+	return api.String(), true
+}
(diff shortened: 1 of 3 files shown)


marketplace/github_url_test.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+func (t *rewriteTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if req.URL.Host == t.match {
+		targetURL, err := req.URL.Parse(t.target)
+		if err != nil {
+			return nil, err
+		}
+		req.URL.Scheme = targetURL.Scheme
+		req.URL.Host = targetURL.Host
```
acmacalister, line 179: `RoundTrip` modifies `req.URL.Scheme`, `req.URL.Host`, and `req.Host` directly on the incoming request. The `http.RoundTripper` contract says RoundTrip must not modify the original request — `http.Client` may inspect or re-use it after the call returns (e.g. on redirect or retry). Using `req.Clone()` keeps this correct:

```go
func (t *rewriteTransport) RoundTrip(req *http.Request) (*http.Response, error) {
	if req.URL.Host == t.match {
		targetURL, err := req.URL.Parse(t.target)
		if err != nil {
			return nil, err
		}
		req = req.Clone(req.Context())
		req.URL.Scheme = targetURL.Scheme
		req.URL.Host = targetURL.Host
		req.Host = targetURL.Host
	}
	return t.wrapped.RoundTrip(req)
}
```

marketplace/github_url.go
```diff
@@ -0,0 +1,51 @@
...
+	if len(parts) < 5 || (parts[2] != "blob" && parts[2] != "raw") {
+		return rawURL, false
+	}
+	owner, repo, ref := parts[0], parts[1], parts[3]
+	path := strings.Join(parts[4:], "/")
+
+	api := &url.URL{
+		Scheme: "https",
+		Host:   "api.github.com",
+		Path:   "/repos/" + owner + "/" + repo + "/contents/" + path,
```
acmacalister: The `owner`, `repo`, and `path` segments come from `url.Parse` and are already percent-decoded. Concatenating them back into a URL path raw means any characters that need encoding (spaces, `#`, `?`, etc.) would silently produce a malformed URL. `url.JoinPath` handles this correctly:

```go
apiURL, _ := url.JoinPath(
	"https://api.github.com",
	"repos", owner, repo, "contents", path,
)
api, _ := url.Parse(apiURL)
q := api.Query()
q.Set("ref", ref)
api.RawQuery = q.Encode()
return api.String(), true
```

In practice GitHub names rarely contain such characters, but this is safer and only a few extra lines.

acmacalister (approved):
Nice, clean implementation — the normalizer is well-isolated, test coverage is thorough, and the auth wiring through the existing `GitHubTokenFunc` just works. Two small things inline.

marketplace/github_url.go
daltoniam: Good catch — switched to `url.JoinPath` in ad70320.

marketplace/github_url_test.go
daltoniam: Right, the `RoundTripper` contract is real even in a test helper. Now cloning the request before mutating in ad70320.

marketplace/github_url.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	if strings.ToLower(u.Host) != "github.com" {
+		return rawURL, false
+	}
+	// /{owner}/{repo}/blob|raw/{ref}/{path...}
+	parts := strings.Split(strings.TrimPrefix(u.Path, "/"), "/")
+	if len(parts) < 5 || (parts[2] != "blob" && parts[2] != "raw") {
+		return rawURL, false
+	}
+	owner, repo, ref := parts[0], parts[1], parts[3]
```
acmacalister, line 38: Branch names with slashes (e.g. `feature/my-branch`) will be parsed incorrectly here — `parts[3]` captures only the first segment as `ref`, and the remaining segments become part of `path`. So `blob/feature/my-branch/manifest.json` would produce `ref=feature` and `path=my-branch/manifest.json`, which is wrong.

This is an inherent ambiguity in the GitHub HTML URL format (the Contents API itself has the same problem), so there's no clean fix without a GitHub API round-trip to resolve the ref. Worth a comment in the godoc so users who configure slash-containing branches know to use the `raw.githubusercontent.com` or direct Contents API form instead:

```go
// Note: branch names containing slashes (e.g. feature/my-branch) cannot be
// reliably parsed from the HTML URL form — parts[3] is treated as the full
// ref. Users with such branches should use a raw.githubusercontent.com URL.
```

acmacalister (approved):
Clean implementation — the `normalizeGitHubURL` abstraction is well-scoped, both previous comments are properly addressed, and the end-to-end tests are thorough. One edge case to be aware of inline.

PR #173, opened 2026-05-18
Surface plugin live-load failures in marketplace UI

daltoniam:
## Summary

- Marketplace install/upload/update handlers in `web/web_marketplace.go` called `liveLoadPlugin` and silently discarded the error (`_ = w.liveLoadPlugin(...)`), so the UI always showed "Installed and loaded" success — even when the WASM module failed to load (missing required export, ABI mismatch, configure error). The plugin then never appeared at `/integrations` and the user had no signal as to why.
- Forward the load error to the existing flash banner so the install page surfaces what went wrong (e.g. `Installed here but load failed: wasm: module does not export 'metadata'`).

Discovered while debugging a `here` plugin that installed but never registered because its `here.wasm` was missing the required `metadata()` export.

## Test plan

- [x] `go vet ./web/...`
- [x] `go test ./web/...`
- [x] `golangci-lint run ./web/...`
- [ ] Manual: install a plugin whose `.wasm` is missing a required export and confirm the error appears in the flash banner instead of a misleading success.

### DIFF
--- a/web/web_marketplace.go
+++ b/web/web_marketplace.go
@@ -99,7 +99,10 @@ func (w *WebServer) handlePluginInstall(rw http.ResponseWriter, r *http.Request)
 		return
 	}
 
-	_ = w.liveLoadPlugin(r.Context(), ip.Path, "")
+	if err := w.liveLoadPlugin(r.Context(), ip.Path, ""); err != nil {
+		http.Redirect(rw, r, fmt.Sprintf("/plugins?error=Installed+%s+but+load+failed:+%s", ip.Name, urlEncode(err.Error())), http.StatusSeeOther)
+		return
+	}
 	http.Redirect(rw, r, fmt.Sprintf("/plugins?success=Installed+and+loaded+%s@%s.", ip.Name, ip.Version), http.StatusSeeOther)
 }
 
@@ -127,7 +130,10 @@ func (w *WebServer) handlePluginInstallURL(rw http.ResponseWriter, r *http.Reque
 		return
 	}
 
-	_ = w.liveLoadPlugin(r.Context(), ip.Path, "")
+	if err := w.liveLoadPlugin(r.Context(), ip.Path, ""); err != nil {
+		http.Redirect(rw, r, fmt.Sprintf("/plugins?error=Installed+%s+but+load+failed:+%s", ip.Name, urlEncode(err.Error())), http.StatusSeeOther)
+		return
+	}
 	http.Redirect(rw, r, fmt.Sprintf("/plugins?success=Installed+and+loaded+%s.", ip.Name), http.StatusSeeOther)
 }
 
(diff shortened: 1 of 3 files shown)


web/web_marketplace.go
```diff
@@ -162,7 +168,10 @@ func (w *WebServer) handlePluginUpload(rw http.ResponseWriter, r *http.Request)
 		return
 	}
 
-	_ = w.liveLoadPlugin(r.Context(), ip.Path, "")
+	if err := w.liveLoadPlugin(r.Context(), ip.Path, ""); err != nil {
+		http.Redirect(rw, r, fmt.Sprintf("/plugins?error=Uploaded+%s+but+load+failed:+%s", ip.Name, urlEncode(err.Error())), http.StatusSeeOther)
```
acmacalister: `ip.Name` here comes straight from `r.FormValue("name")` (via `InstallFromBytes`) and isn't URL-encoded, while `err.Error()` correctly goes through `urlEncode`. A plugin name like `foo&success=Installed+backdoor!` would split into two query params and silently override the flash state.

Same pattern on lines 103, 134, and 225 — all four new redirect lines have this. `urlEncode(ip.Name)` everywhere fixes it:

```go
http.Redirect(rw, r, fmt.Sprintf("/plugins?error=Uploaded+%s+but+load+failed:+%s",
    urlEncode(ip.Name), urlEncode(err.Error())), http.StatusSeeOther)
```

web/web_marketplace.go
```diff
@@ -99,7 +99,10 @@ func (w *WebServer) handlePluginInstall(rw http.ResponseWriter, r *http.Request)
 		return
 	}
 
-	_ = w.liveLoadPlugin(r.Context(), ip.Path, "")
+	if err := w.liveLoadPlugin(r.Context(), ip.Path, ""); err != nil {
```
acmacalister, line 110: The new error-surface behavior (what this whole PR is about) isn't tested. A test that stubs `wasmLoader` to return an error would confirm the redirect goes to `/plugins?error=...` instead of `/plugins?success=...`.

Right now `wasmLoader` is a concrete `*wasmmod.Loader`, so mocking it requires extracting a small interface — something like:

```go
type pluginLoader interface {
    LoadPlugin(ctx context.Context, path, nameOverride string) error
    UnloadPlugin(ctx context.Context, name string) error
}
```

Swap the field type to that interface and you can inject an error-returning stub in tests.

acmacalister (commented):
Good fix — silently discarding liveLoadPlugin errors was a real papercut. Two things worth addressing before merge, both inline.

web/web_marketplace.go
daltoniam: Good catch — switched `urlEncode` to `url.QueryEscape` so reserved chars are actually escaped, and wrapped `ip.Name` (and `ip.Version`) at all four redirect sites. Added `TestPluginUpload_NameEscaping` to lock in the regression. Fixed in 68495083761c457f7038f9b5fee40f3f0a09586e.

web/web_marketplace.go
daltoniam: Done — extracted a `pluginLoader` interface (`LoadPlugin`/`UnloadPlugin`), switched the `WebServer` field to it, and guarded the constructor against the typed-nil-interface pitfall when callers pass a nil `*wasm.Loader`. Added `TestPluginUpload_LiveLoadError` and `TestPluginUpload_LiveLoadSuccess` using an `errLoader` stub. Fixed in 68495083761c457f7038f9b5fee40f3f0a09586e.

acmacalister (approved):
Solid fix — the `pluginLoader` interface extraction is clean, the nil-guard in `New()` correctly handles the typed-nil pitfall, and `url.QueryEscape` is the right replacement for the old space-only encoding. One thing to follow up on: `handlePluginUninstall` at line 205 has the same unencoded-name-in-redirect issue the PR fixed in the other four handlers — `fmt.Sprintf("/plugins?success=Uninstalled+%s.", name)` where `name` comes straight from `r.FormValue`. Worth a one-line `urlEncode(name)` there before shipping.

PR #175, opened 2026-05-19
Defer browser startup until needed

daltoniam:
## Summary

- Defer Playwright startup until a browser-backed integration actually requests a session so server boot doesn't crash in environments without a usable browser runtime.
- Add lazy browser service lifecycle coverage and make the browser evaluate assertion tolerant of numeric return type differences.

## Test plan

- [x] make ci

💘 Generated with Crush

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"sync"
 	"syscall"
 	"time"
 
@@ -201,16 +202,18 @@ func runServer(stdioMode bool, port int, discoverAll bool) {
 		log.Fatalf("Failed to load config: %v", err)
 	}
 
-	browserDone := make(chan struct{})
-	var browserSvc mcp.BrowserService
-	go func() {
-		defer close(browserDone)
+	browserSvc := newLazyBrowserService(func(ctx context.Context) (mcp.BrowserService, error) {
 		svc, err := browser.New(true /* headless */)
 		if err != nil {
-			log.Printf("browser service unavailable (%v) — browser-based integrations disabled", err)
-			return
+			return nil, fmt.Errorf("browser service unavailable: %w", err)
 		}
-		browserSvc = svc
+		log.Println("browser service ready — Amazon browser features enabled")
+		return svc, nil
+	})
+	defer func() {
+		if err := browserSvc.Close(); err != nil {
+			log.Printf("browser service close failed: %v", err)
+		}
 	}()
 
 	gmailIntegration := gmail.New()
@@ -344,13 +347,7 @@ func runServer(stdioMode bool, port int, discoverAll bool) {
 	gchat.SetConfigService(gchatIntegration, cfgMgr)
 	gpeople.SetConfigService(gpeopleIntegration, cfgMgr)
 	gmeet.SetConfigService(gmeetIntegration, cfgMgr)
-	go func() {
-		<-browserDone
-		if browserSvc != nil {
-			amazon.SetBrowserService(amazonIntegration, browserSvc)
-			log.Println("browser service ready — Amazon browser features enabled")
-		}
-	}()
+	amazon.SetBrowserService(amazonIntegration, browserSvc)
 
 	var serverOpts []server.Option
 	if discoverAll {
@@ -482,10 +479,49 @@ func runServer(stdioMode bool, port int, discoverAll bool) {
 		log.Fatalf("HTTP server error: %v", err)
 	}
 
-	<-browserDone
-	if browserSvc != nil {
-		browserSvc.Close() //nolint:errcheck
+}
+
+type lazyBrowserService struct {
+	mu      sync.Mutex
+	new     func(ctx context.Context) (mcp.BrowserService, error)
+	service mcp.BrowserService
+	closed  bool
+}
+
+func newLazyBrowserService(newFn func(ctx context.Context) (mcp.BrowserService, error)) *lazyBrowserService {
+	return &lazyBrowserService{new: newFn}
+}
+
+func (s *lazyBrowserService) NewSession(ctx context.Context) (mcp.BrowserSession, error) {
+	s.mu.Lock()
+	if s.closed {
+		s.mu.Unlock()
+		return nil, fmt.Errorf("browser service closed")
 	}
+	if s.service == nil {
+		svc, err := s.new(ctx)
+		if err != nil {
+			s.mu.Unlock()
+			return nil, err
+		}
+		s.service = svc
+	}
+	svc := s.service
+	s.mu.Unlock()
+	return svc.NewSession(ctx)
+}
+
+func (s *lazyBrowserService) Close() error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if s.closed {
+		return nil
+	}
+	s.closed = true
+	if s.service == nil {
+		return nil
+	}
+	return s.service.Close()
 }
 
 // metricsPath returns the on-disk path for persisted lifetime metrics.
--- a/cmd/server/main_test.go
+++ b/cmd/server/main_test.go
@@ -1,0 +1,157 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"testing"
+	"time"
+
+	mcp "github.com/daltoniam/switchboard"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestLazyBrowserService_DoesNotStartUntilFirstSession(t *testing.T) {
+	starts := 0
+	svc := newLazyBrowserService(func(context.Context) (mcp.BrowserService, error) {
+		starts++
+		return &fakeBrowserService{}, nil
+	})
+
+	require.NoError(t, svc.Close())
+	assert.Equal(t, 0, starts)
+}
+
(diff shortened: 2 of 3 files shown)


cmd/server/main_test.go
```diff
@@ -0,0 +1,70 @@
...
+func TestLazyBrowserService_NewSessionAfterClose(t *testing.T) {
+	svc := newLazyBrowserService(func(context.Context) (mcp.BrowserService, error) {
+		return &fakeBrowserService{}, nil
+	})
+
+	require.NoError(t, svc.Close())
+	_, err := svc.NewSession(context.Background())
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "browser service closed")
+}
```
acmacalister, line 52: The three tests cover no-start-before-use, reuse, and post-close error, but none of them verify that `Close()` actually forwards to the inner service. `fakeBrowserService` has a `closed` field that's never asserted — if `s.service.Close()` were accidentally dropped from the implementation, Playwright processes would leak silently.

Worth adding a fourth case:

```go
func TestLazyBrowserService_CloseForwardsToInner(t *testing.T) {
	inner := &fakeBrowserService{}
	svc := newLazyBrowserService(func(context.Context) (mcp.BrowserService, error) {
		return inner, nil
	})

	_, err := svc.NewSession(context.Background())
	require.NoError(t, err)

	require.NoError(t, svc.Close())
	assert.True(t, inner.closed)
}
```

acmacalister (commented):
Lazy init pattern is clean — the goroutine + channel dance is gone and the mutex usage is sound. One test path missing inline.

cmd/server/main_test.go
daltoniam: Added `TestLazyBrowserService_CloseForwardsToInner` and re-ran `make ci`.

cmd/server/main_test.go
```diff
@@ -0,0 +1,83 @@
...
+	svc := newLazyBrowserService(func(context.Context) (mcp.BrowserService, error) {
+		return inner, nil
+	})
+
+	_, err := svc.NewSession(context.Background())
+	require.NoError(t, err)
+
+	require.NoError(t, svc.Close())
+	assert.True(t, inner.closed)
+}
```
acmacalister, line 65: We have tests for no-start-before-use, reuse, post-close rejection, and close forwarding — but the factory error path isn't covered. If `s.new(ctx)` returns an error, `s.service` stays nil and the next call retries rather than permanently failing. That retry behaviour is worth pinning:

```go
func TestLazyBrowserService_FactoryRetryAfterTransientError(t *testing.T) {
	attempts := 0
	inner := &fakeBrowserService{}
	svc := newLazyBrowserService(func(context.Context) (mcp.BrowserService, error) {
		attempts++
		if attempts == 1 {
			return nil, fmt.Errorf("transient error")
		}
		return inner, nil
	})

	_, err := svc.NewSession(context.Background())
	require.Error(t, err)
	assert.Contains(t, err.Error(), "transient error")

	// second call should succeed — factory is retried, not permanently failed
	_, err = svc.NewSession(context.Background())
	require.NoError(t, err)
	assert.Equal(t, 2, attempts)
}
```

Without this, someone could accidentally change the failure mode (e.g. set `s.closed = true` on init error, permanently disabling the service) without a test catching it.

acmacalister (approved):
Clean fix — the lazy wrapper is well-structured, mutex discipline is correct, and the `EqualValues` change in `browser_test.go` is a nice robustness improvement. One gap in the test suite worth filling inline.

cmd/server/main_test.go
daltoniam: Added `TestLazyBrowserService_FactoryRetryAfterTransientError` to pin retry-after-init-error behavior and re-ran `make ci`.

cmd/server/main_test.go
```diff
@@ -0,0 +1,104 @@
...
+	})
+
+	_, err := svc.NewSession(context.Background())
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "transient error")
+
+	_, err = svc.NewSession(context.Background())
+	require.NoError(t, err)
+	assert.Equal(t, 2, attempts)
+}
```
acmacalister, line 99: The five tests cover all the sequential behaviors really well, but none of them call `NewSession` concurrently — which means the `sync.Mutex` isn't actually exercised by the test suite. You could remove the lock entirely and every test would still pass; the race detector only fires when concurrent access happens in the test run itself.

A sixth test would pin this and make the mutex load-bearing:

```go
func TestLazyBrowserService_ConcurrentInitOnce(t *testing.T) {
	var (
		mu     sync.Mutex
		starts int
	)
	svc := newLazyBrowserService(func(context.Context) (mcp.BrowserService, error) {
		mu.Lock()
		starts++
		mu.Unlock()
		return &fakeBrowserService{}, nil
	})

	var wg sync.WaitGroup
	for range 10 {
		wg.Add(1)
		go func() {
			defer wg.Done()
			_, err := svc.NewSession(context.Background())
			assert.NoError(t, err)
		}()
	}
	wg.Wait()
	assert.Equal(t, 1, starts)
}
```

With `-race` this would also catch any data races if the mutex were weakened in a future refactor.

acmacalister (approved):
Solid fix — lazy init cleanly eliminates the boot-time crash and the implementation handles all the edge cases. Previous round of review feedback was fully addressed. One small thought inline about concurrent test coverage for the mutex.

cmd/server/main_test.go
daltoniam: Added `TestLazyBrowserService_ConcurrentInitOnce` to exercise concurrent initialization under `-race` and re-ran `make ci`.

cmd/server/main.go
```diff
@@ -442,10 +439,43 @@ func runServer(stdioMode bool, port int, discoverAll bool) {
...
+		return nil, fmt.Errorf("browser service closed")
+	}
+	if s.service == nil {
+		svc, err := s.new(ctx)
+		if err != nil {
+			return nil, err
+		}
+		s.service = svc
+	}
+	return s.service.NewSession(ctx)
```
acmacalister: The `defer s.mu.Unlock()` at the top of the function means the mutex is still held here, which serializes every `NewSession` call after the lazy init is done. If Playwright session creation is slow (or if the Amazon tools are ever called concurrently), all callers queue up behind each other unnecessarily.

The init guard only needs the lock; the delegate call to `s.service.NewSession` doesn't:

```go
func (s *lazyBrowserService) NewSession(ctx context.Context) (mcp.BrowserSession, error) {
	s.mu.Lock()
	if s.closed {
		s.mu.Unlock()
		return nil, fmt.Errorf("browser service closed")
	}
	if s.service == nil {
		svc, err := s.new(ctx)
		if err != nil {
			s.mu.Unlock()
			return nil, err
		}
		s.service = svc
	}
	svc := s.service
	s.mu.Unlock()
	return svc.NewSession(ctx)
}
```

This also makes the `TestLazyBrowserService_ConcurrentInitOnce` assertions stronger — with the lock released before the delegate, the race detector would catch a non-thread-safe underlying service rather than silently serializing it away.

acmacalister (approved):
Clean fix — the lazy init approach is exactly right and the test coverage is excellent. One thing worth considering inline regarding lock scope.

cmd/server/main.go
daltoniam: Released the lazy wrapper mutex before delegating to the underlying browser service and kept the fake service thread-safe for the race test. Re-ran `make ci`.

cmd/server/main.go
```diff
@@ -442,10 +439,46 @@ func runServer(stdioMode bool, port int, discoverAll bool) {
...
+	}
+	svc := s.service
+	s.mu.Unlock()
+	return svc.NewSession(ctx)
+}
+
+func (s *lazyBrowserService) Close() error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.closed = true
```
acmacalister, line 520: `s.closed` is set here but never checked at the top of this function, so `Close()` isn't idempotent. A second call with a started service goes straight through to `s.service.Close()` again. The real Playwright backend calls `s.browser.Close()` and `s.pw.Stop()` — both would fail on double-invocation.

Moving the check before the assignment fixes it:

```go
func (s *lazyBrowserService) Close() error {
	s.mu.Lock()
	defer s.mu.Unlock()
	if s.closed {
		return nil
	}
	s.closed = true
	if s.service == nil {
		return nil
	}
	return s.service.Close()
}
```

Worth adding a `TestLazyBrowserService_DoubleCloseIsNoop` to pin it — call `NewSession`, then `Close` twice, assert both return nil and that the inner service's close count is 1.

acmacalister (approved):
Clean PR — build passes, tests pass with -race, lint is clean. One thing to look at inline: `Close()` isn't idempotent as written.

cmd/server/main.go
daltoniam: Idempotency check is in place at `main.go:517-519` and `TestLazyBrowserService_DoubleCloseIsNoop` pins the behavior (asserts `inner.closeCount == 1` after two `Close()` calls). Re-ran `make ci`.

acmacalister (approved):
Clean PR — builds, tests pass (race-clean), lint clean. The lazy wrapper design is solid: factory called once under the lock, lock released before the delegate call, and all seven lifecycle states are covered by tests. Previous feedback was all addressed correctly. LGTM.

PR #176, opened 2026-05-21
Show tool descriptions on integration pages

daltoniam:
## Summary

- The Available Tools section on the generic integration detail page (and the Postgres / X setup pages) previously rendered bare tool-name chips, giving users no indication of what each operation actually does.
- Each tool now renders as a name + description row sourced from the integration's `mcp.ToolDefinition.Description`, so the UI stays in sync as integrations evolve.
- Mirrors the change shipped in hosted Switchboard ([Vluxe/Switchboard-hosted#37](https://github.com/Vluxe/Switchboard-hosted/pull/37)), scoped to the OSS pages (no policy editor exists here).

## Test plan

- [x] `make generate`
- [x] `make ci` (build, vet, test-race, lint, gosec, govulncheck)
- [x] New render tests in `web/templates/pages/integration_test.go` cover the description, the empty fallback, and the omitted-section case
- [ ] Spot-check `/integrations/<name>` in a local server

💘 Generated with Crush


web/templates/components/components.templ
```diff
@@ -55,3 +55,14 @@ templ StatCardColor(value int, label string, variant string) {
 templ ToolTag(name string) {
 	<span class="tool-tag">{ name }</span>
 }
+
+templ ToolRow(name string, description string) {
```
acmacalister, line 55: Since this PR replaces all `ToolTag` call sites with `ToolRow`, the `ToolTag` component above (and its generated counterpart in `components_templ.go`) is now unused. The `.tools-list` and `.tool-tag` CSS rules in `base.templ` are also orphaned. Might be worth a small follow-up to clean those up before they accumulate confusion.

acmacalister (approved):
Clean PR — build passes, tests green, lint clean. Nice UX improvement surfacing descriptions rather than bare chip names. One small cleanup opportunity inline.

acmacalister (approved):
Clean PR — builds, tests pass, lint clean. One optional follow-up: a `GET /integrations/testint` handler test in `web_test.go` would close the gap between the template-level tests and the handler wiring (the mock already has `Description: "List things"` set up). Not blocking.

PR #179, opened 2026-05-30
feat(server): add StatelessHandler for replica-safe deployments

daltoniam:
## Summary

- Adds `Server.StatelessHandler()` exposing the SDK's `Stateless: true` mode so consumers deploying behind a load balancer with multiple replicas don't need session affinity.
- Mirrors the existing internal usage in `project_server.go` (which already enables `Stateless: true`).

## Why

The session-based `Handler()` keeps `Mcp-Session-Id` state in-process. When a downstream deployment runs multiple replicas without session affinity, a follow-up SSE GET / `tools/call` POST that hashes to a different pod fails with `503 "Failed to open SSE stream"` or `404 "session not found"`, eventually timing out the client. This was just observed in production with the Claude Desktop install flow against the switchboard-hosted runtime (HPA, 1–5 pods, no affinity). The hosted side will switch its mcpd to `StatelessHandler()` in a companion PR.

## Test plan

- [x] `go build ./...`
- [x] `go test ./server/...`
- [ ] Verify a downstream consumer using `StatelessHandler()` works across replicas

💘 Generated with Crush


server/server.go
```diff
@@ -1493,6 +1493,22 @@ func (s *Server) Handler() http.Handler {
 	)
 }
 
+// StatelessHandler returns an http.Handler that serves MCP over streamable HTTP
+// transport in stateless mode. Every request is handled independently with no
+// session state retained across calls, which makes it safe to deploy behind a
+// load balancer with multiple replicas (no session affinity required).
+func (s *Server) StatelessHandler() http.Handler {
```
acmacalister, line 1500: There's a `TestHandler` at `server_test.go:206` that asserts the stateful handler isn't nil — `StatelessHandler` should have the same minimal coverage. The project convention is "new code must include tests, no exceptions", and the parallel test is one-liner:

```go
func TestStatelessHandler(t *testing.T) {
	s := setupTestServer()
	handler := s.StatelessHandler()
	assert.NotNil(t, handler)
}
```

acmacalister (approved):
Build, tests, and lint all pass. Clean implementation that mirrors the existing stateless pattern in `project_server.go`. One small gap inline.

PR #180, opened 2026-06-01
feat(notion): add v1 OAuth backend alongside v3 cookie path

daltoniam:
## Summary

The current notion integration talks to `www.notion.so/api/v3` with a browser session cookie (`token_v2`). That path still works for users who set it up before OAuth existed, but it's brittle (cookies expire, Notion can change v3 without notice) and unusable for hosted SaaS.

This adds a parallel `notionV1` backend that uses Notion's documented public REST API (`api.notion.com/v1`) with an OAuth bearer token and the date-stamped `Notion-Version` header. I picked `2025-09-03` — that's the version that introduced the `data_sources` model, so the public surface matches the v3 backend's database/data-source split naturally and we don't need a translation layer.

Backend selection happens in `Configure()` based on which credential key is set:

- `access_token` → v1 (OAuth, hosted-friendly)
- `token_v2` → v3 (cookie, self-host legacy)

If both are present, v1 wins. `Execute()` and `Healthy()` delegate to v1 when it's configured; otherwise the existing v3 path runs unchanged. **No behavior change for existing `token_v2` users.**

All tools the v3 backend exposes have a v1 equivalent (data sources, pages, blocks, search, users, comments, plus the convenience helpers `notion_get_page_content` / `notion_create_page_with_content`). Routing lives in `dispatchV1`. If a tool ends up v3-only in the future, `Execute` returns a clear "not supported on the Notion OAuth (v1) backend" error rather than a generic unknown-tool error.

## Test plan

- [x] `go test ./integrations/notion/...` — new `v1_test.go` covers the configure handshake (auth + version headers), and each handler family with httptest servers asserting URL paths, request bodies, and response shaping.
- [x] `go test ./...` — full suite green.
- [x] `golangci-lint run ./integrations/notion/...` — clean.
- [ ] Smoke-test end-to-end against a real Notion OAuth install (wired in Switchboard-hosted, separate PR).

### DIFF
--- a/integrations/notion/v1_blocks.go
+++ b/integrations/notion/v1_blocks.go
@@ -1,0 +1,181 @@
+package notion
+
+import (
+	"context"
+	"fmt"
+
+	mcp "github.com/daltoniam/switchboard"
+)
+
+// --- Blocks (v1) ---
+
+func v1RetrieveBlock(ctx context.Context, n *notionV1, args map[string]any) (*mcp.ToolResult, error) {
+	blockID, err := mcp.ArgStr(args, "block_id")
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	if blockID == "" {
+		return mcp.ErrResult(fmt.Errorf("block_id is required"))
+	}
+	data, err := n.get(ctx, "/blocks/%s", blockID)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	return mcp.RawResult(data)
+}
+
+// v1UpdateBlock translates the v3 `type_content` payload into the v1 shape.
+// v3 callers send: {"properties": {"title": [["new text"]]}}
+// v1 expects: {"<block_type>": {"rich_text": [{"type":"text","text":{"content":"new text"}}]}}
+//
+// To translate we need to know the block's current type, so we fetch it
+// first. Power-users can bypass the translation by providing a pre-shaped
+// v1 payload with the block type as the top-level key (e.g.
+// {"paragraph": {"rich_text": [...]}}) — that pass-through preserves the
+// full v1 expressiveness (annotations, links, mentions, etc).
+func v1UpdateBlock(ctx context.Context, n *notionV1, args map[string]any) (*mcp.ToolResult, error) {
+	r := mcp.NewArgs(args)
+	blockID := r.Str("block_id")
+	typeContent := r.Map("type_content")
+	archived := r.Bool("archived")
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	if blockID == "" {
+		return mcp.ErrResult(fmt.Errorf("block_id is required"))
+	}
+
+	body := map[string]any{}
+	if archived {
+		body["archived"] = true
+	}
+
+	if typeContent != nil {
+		// Pass-through: caller supplied a v1-shaped payload.
+		passedThrough := false
+		for k, v := range typeContent {
+			if k == "properties" || k == "format" {
+				continue
+			}
+			body[k] = v
+			passedThrough = true
+		}
+
+		// Legacy v3 path: translate properties.title to v1 rich_text.
+		if !passedThrough {
+			blockType, err := v1FetchBlockType(ctx, n, blockID)
+			if err != nil {
+				return mcp.ErrResult(err)
+			}
+			text := v1ExtractV3TitleText(map[string]any{"properties": typeContent["properties"]})
+			payload := map[string]any{"rich_text": v1RichTextFromString(text)}
+			body[blockType] = payload
+		}
+	}
+
+	if len(body) == 0 {
+		return mcp.ErrResult(fmt.Errorf("at least one of type_content or archived must be set"))
+	}
+
+	data, err := n.patch(ctx, "/blocks/"+blockID, body)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	return mcp.RawResult(data)
+}
+
+func v1DeleteBlock(ctx context.Context, n *notionV1, args map[string]any) (*mcp.ToolResult, error) {
+	blockID, err := mcp.ArgStr(args, "block_id")
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	if blockID == "" {
+		return mcp.ErrResult(fmt.Errorf("block_id is required"))
+	}
+	data, err := n.del(ctx, "/blocks/%s", blockID)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	return mcp.RawResult(data)
+}
+
+func v1GetBlockChildren(ctx context.Context, n *notionV1, args map[string]any) (*mcp.ToolResult, error) {
+	r := mcp.NewArgs(args)
+	blockID := r.Str("block_id")
+	startCursor := r.Str("start_cursor")
+	pageSize := r.Int("page_size")
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	if blockID == "" {
+		return mcp.ErrResult(fmt.Errorf("block_id is required"))
+	}
+	if pageSize <= 0 || pageSize > 100 {
+		pageSize = 100
+	}
+	qs := queryEncode(map[string]string{
+		"start_cursor": startCursor,
+		"page_size":    fmt.Sprintf("%d", pageSize),
+	})
+	data, err := n.get(ctx, "/blocks/%s/children%s", blockID, qs)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	return mcp.RawResult(data)
+}
+
+func v1AppendBlockChildren(ctx context.Context, n *notionV1, args map[string]any) (*mcp.ToolResult, error) {
+	r := mcp.NewArgs(args)
+	blockID := r.Str("block_id")
+	after := r.Str("after")
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	if blockID == "" {
+		return mcp.ErrResult(fmt.Errorf("block_id is required"))
+	}
+	childrenAny, ok := args["children"].([]any)
+	if !ok || len(childrenAny) == 0 {
+		return mcp.ErrResult(fmt.Errorf("children is required"))
+	}
+	if len(childrenAny) > 100 {
+		return mcp.ErrResult(fmt.Errorf("v1 append_block_children accepts up to 100 children per call; got %d. Split into multiple calls", len(childrenAny)))
+	}
+
+	normalized := make([]any, 0, len(childrenAny))
+	for _, c := range childrenAny {
+		if cm, ok := c.(map[string]any); ok {
+			normalized = append(normalized, v1NormalizeBlock(cm))
+		}
+	}
+
+	body := map[string]any{"children": normalized}
+	if after != "" {
+		body["after"] = after
+	}
+	data, err := n.patch(ctx, "/blocks/"+blockID+"/children", body)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	return mcp.RawResult(data)
+}
+
+// v1FetchBlockType issues a GET /blocks/{id} and returns the block's
+// "type" field. Used by update_block to translate the v3 properties.title
+// payload into a v1 rich_text payload keyed by the block's type.
+func v1FetchBlockType(ctx context.Context, n *notionV1, blockID string) (string, error) {
+	data, err := n.get(ctx, "/blocks/%s", blockID)
+	if err != nil {
+		return "", err
+	}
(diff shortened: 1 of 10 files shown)


integrations/notion/v1_dispatch.go
```diff
@@ -0,0 +1,51 @@
...
+
+// dispatchV1 is the v1 backend's tool routing table. Mirrors the v3
+// `dispatch` map; same tool names, different handlers. When the
+// integration is configured with an OAuth access token, notion.Execute
+// looks tools up here.
+//
+// Tools that have no v1 equivalent (none today, but reserved for future
+// v3-only surfaces) should be left out — notion.Execute returns a
+// "not supported on v1 backend" error for missing entries.
+var dispatchV1 = map[mcp.ToolName]v1HandlerFunc{
```
acmacalister, line 13: `dispatchV1` is missing the two parity tests every dispatch map in this project is required to have. The v3 dispatch has `TestDispatchMap_EveryToolHasHandler` and `TestDispatchMap_NoOrphanHandlers` in `notion_test.go` — `dispatchV1` needs equivalents.

Right now, nothing enforces that every entry in `dispatchV1` has a corresponding `ToolDefinition` in `Tools()`, or vice versa. If a tool name is misspelled here or a new tool is added to `Tools()` without a v1 handler, it silently routes to the "not yet supported on v1 backend" error at runtime instead of failing at test time.

```go
func TestDispatchV1_AllToolsCovered(t *testing.T) {
    i := New()
    for _, tool := range i.Tools() {
        _, ok := dispatchV1[tool.Name]
        assert.True(t, ok, "tool %s has no v1 dispatch handler", tool.Name)
    }
}

func TestDispatchV1_NoOrphanHandlers(t *testing.T) {
    i := New()
    toolNames := make(map[mcp.ToolName]bool)
    for _, tool := range i.Tools() {
        toolNames[tool.Name] = true
    }
    for name := range dispatchV1 {
        assert.True(t, toolNames[name], "v1 dispatch handler %s has no tool definition", name)
    }
}
```

integrations/notion/v1_data_sources.go
```diff
@@ -0,0 +1,209 @@
...
+	if properties != nil {
+		body["properties"] = properties
+	}
+	if len(body) == 0 {
+		return mcp.ErrResult(fmt.Errorf("at least one of title or properties must be set"))
+	}
+
+	// Prefer data_sources; fall back to databases for older IDs.
+	data, err := n.patch(ctx, "/data_sources/"+id, body)
+	if err != nil {
```
acmacalister, line 129: This fallback triggers on any error from `/data_sources/`, including transient ones. If Notion rate-limits the request (429), `doRequest` returns a `*mcp.RetryableError` — but instead of propagating it so the runtime can back off, we swallow it and immediately fire a second request to `/databases/`. Same issue exists in `v1QueryDataSource` (line 164) and `v1FetchDatabaseOrDataSource` (line 199).

The fallback only makes sense for 404-style "this ID is not a data source" responses. A guard like this preserves the intent while letting retryable errors escape:

```go
data, err := n.patch(ctx, "/data_sources/"+id, body)
if err != nil {
    var re *mcp.RetryableError
    if errors.As(err, &re) {
        return mcp.ErrResult(err)
    }
    data, err = n.patch(ctx, "/databases/"+id, body)
    if err != nil {
        return mcp.ErrResult(err)
    }
}
```

integrations/notion/v1_pages.go
```diff
@@ -0,0 +1,463 @@
...
+		return map[string]any{"type": "workspace", "workspace": true}, nil
+	}
+	return nil, fmt.Errorf("parent must contain page_id, database_id, data_source_id, or workspace:true")
+}
+
+// v1TitleKeyForParent returns the property key under which to write a
+// convenience title. Page-parented pages use the literal "title"; rows
+// in a database use whatever the database's title column is named, but
+// the caller must specify that — we default to "title" if unknown.
+func v1TitleKeyForParent(parent map[string]any) string {
```
acmacalister: Both branches of `v1TitleKeyForParent` return `"title"`, so the condition is never evaluated. The comment above acknowledges that database rows use a user-defined title column name, but the code doesn't implement that distinction — it just returns `"title"` unconditionally.

Either inline the literal `"title"` at the call site and delete this function, or implement the differentiation the comment describes. As written, the function misleads readers into thinking there's conditional logic here.

integrations/notion/v1_other.go
```diff
@@ -0,0 +1,163 @@
...
+	return mcp.RawResult(data)
+}
+
+// --- Tiny shared helper ---
+//
+// jsonUnmarshalLite is the smallest possible wrapper around json.Unmarshal
+// so v1 handlers don't have to reach for the legacy unmarshalJSON helper
+// (which exists in the v3 codepath but has no special semantics worth
+// sharing).
+func jsonUnmarshalLite(data []byte, v any) error {
```
acmacalister: `jsonUnmarshalLite` is a one-liner wrapper around `json.Unmarshal` that adds no behaviour and is only called once (in `v1FetchBlockType`). Worth replacing the call site with `json.Unmarshal` directly and removing this.

acmacalister (approved):
Good stuff overall — the dual-backend design is clean, the test coverage is thorough, build and lint are both green. Four things worth tidying up inline.

integrations/notion/v1_dispatch.go
daltoniam: Done in 785138e — added both `TestDispatchV1_EveryToolHasHandler` and `TestDispatchV1_NoOrphanHandlers` in v1_test.go, matching the v3 pattern.

integrations/notion/v1_data_sources.go
daltoniam: Good catch. Fixed in 785138e — added an `errors.As` guard on `*mcp.RetryableError` to all three fallback sites (`v1UpdateDataSource`, `v1QueryDataSource`, `v1FetchDatabaseOrDataSource`). Rate-limits and 5xx now propagate so the runtime can back off; only non-retryable errors fall through to the `/databases/` retry.

integrations/notion/v1_pages.go
daltoniam: You're right, the function was dead. Removed in 785138e and inlined `const titleKey = "title"` at the call site, with the comment updated to reflect what actually happens (default to `title`; callers can override via the properties map for database rows with a renamed title column).

integrations/notion/v1_other.go
daltoniam: Removed in 785138e. `v1FetchBlockType` now calls `json.Unmarshal` directly; `encoding/json` moved into v1_blocks.go and out of v1_other.go.

integrations/notion/v1_blocks.go
```diff
@@ -0,0 +1,182 @@
...
+// v1UpdateBlock translates the v3 `type_content` payload into the v1 shape.
+// v3 callers send: {"properties": {"title": [["new text"]]}}
+// v1 expects: {"<block_type>": {"rich_text": [{"type":"text","text":{"content":"new text"}}]}}
+//
+// To translate we need to know the block's current type, so we fetch it
+// first. Power-users can bypass the translation by providing a pre-shaped
+// v1 payload with the block type as the top-level key (e.g.
+// {"paragraph": {"rich_text": [...]}}) — that pass-through preserves the
+// full v1 expressiveness (annotations, links, mentions, etc).
+func v1UpdateBlock(ctx context.Context, n *notionV1, args map[string]any) (*mcp.ToolResult, error) {
```
acmacalister, line 37: `v1UpdateBlock` is the most complex handler in this PR — it makes an extra `GET /blocks/{id}` to discover the block type, then translates the v3 `properties.title` shape to a v1 `rich_text` payload. The `passedThrough` detection (skip "properties"/"format", treat anything else as a v1 payload) has branching that's easy to accidentally break.

Every other non-trivial handler has at least one `httptest` case; this one doesn't. The dispatch parity tests confirm it's wired up, but they don't exercise any of the translation logic. Worth adding cases for at least the v3 legacy path (where `v1FetchBlockType` is called) and the v1 pass-through path.

integrations/notion/v1_data_sources.go
```diff
@@ -0,0 +1,228 @@
...
+		body["start_cursor"] = startCursor
+	}
+	if pageSize > 0 {
+		if pageSize > 100 {
+			pageSize = 100
+		}
+		body["page_size"] = pageSize
+	}
+
+	data, err := n.post(ctx, "/data_sources/"+id+"/query", body)
```
acmacalister, line 172: The `errors.As` guard here is the right fix, but there's no test proving a 429 from `/data_sources/` actually stops the fallback. `TestV1_QueryDataSourceFallsBackToDatabases` only exercises the 404 path — if the guard is ever removed, the suite stays green.

Something like:

```go
func TestV1_QueryDataSourceDoesNotFallBackOn429(t *testing.T) {
    callCount := 0
    v1, _ := newV1WithServer(t, func(w http.ResponseWriter, r *http.Request) {
        callCount++
        w.Header().Set("Retry-After", "5")
        w.WriteHeader(http.StatusTooManyRequests)
        _, _ = w.Write([]byte(`{"object":"error","code":"rate_limited","message":"slow down"}`))
    })
    _, err := v1QueryDataSource(context.Background(), v1, map[string]any{"data_source_id": "ds-1"})
    require.Error(t, err)
    var re *mcp.RetryableError
    require.ErrorAs(t, err, &re)
    assert.Equal(t, 1, callCount, "429 should not trigger the /databases/ fallback")
}
```

Same test is worth having for `v1UpdateDataSource` and `v1FetchDatabaseOrDataSource` (used by `v1RetrieveDataSource`).

integrations/notion/v1_pages.go
```diff
@@ -0,0 +1,452 @@
...
+		if cm, ok := c.(map[string]any); ok {
+			children = append(children, v1NormalizeBlock(cm))
+		}
+	}
+
+	body, err := v1BuildPageBody(parent, properties, title, children)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	if len(children) > 100 {
```
acmacalister, line 224: The `> 100` guard fires after `v1BuildPageBody` has already normalized all N children and constructed the full body. No API call is made so the behavior is correct, but for a 200-child input we're doing twice the normalization work before surfacing the error. Moving the check before the `v1BuildPageBody` call (or even right after the normalization loop) keeps validation upfront:

```go
if len(childrenAny) > 100 {
    return mcp.ErrResult(fmt.Errorf("v1 create_page_with_content accepts up to 100 children..."))
}
children := make([]map[string]any, 0, len(childrenAny))
...
body, err := v1BuildPageBody(...)
```

acmacalister (commented):
Solid PR — build passes, tests pass, lint clean. The retryable-error guards and dispatch parity tests are both in good shape. Three things inline, two of which are about test coverage.

arjansingh (commented):
Existing code looks fine. it's worth adding `compact.yml` specs because IIRC, even the v1 api has some extraneous info.

arjansingh (approved):

daltoniam:
Done in 4eb2ba7 — added `compact_v1.yaml` with v1-shape specs and wired `CompactSpec` / `MaxBytes` / `Views` to switch on `n.v1 != nil` so the v3 specs (parent_id flat, properties.title flat string) keep working on the cookie backend while v1 gets its own paths (nested `parent.{type,page_id}`, typed `properties` map, etc.).

You were right about the noise — drops `created_by` / `last_edited_by` full user objects, `cover`, `icon`, `public_url`, `request_status`, the `is_archived` / `is_locked` duplicates, the deprecated `archived` alias, and the `object` / `type` wrappers. Multi-view tools (search, query_data_source, retrieve_data_source, get_page_content, retrieve_comments) get the same slim-default-+-full shape as v3.

Backed by a `TestV1FieldCompactionSpecs_ParityWithV3` test that fails if any tool is in one yaml but not the other, plus mirrored versions of all four structural tests (`AllParse`, `NoDuplicateTools`, `NoOrphanSpecs`, `NoMutationTools`) and switch-on-backend assertions for `CompactSpec` and `Views`.

One known gap: views are JSON-only on v1 for now — the existing markdown renderers are v3 record-map shaped and would mis-render v1 blocks. A v1 markdown renderer is worth a follow-up PR but felt out of scope here.

acmacalister (approved):
Build clean, full test suite green (race detector), lint at 0. The dual-backend design is well-executed and the parity/compaction tests are solid additions.

Two items from acmacalister's fixup review are still outstanding before merge: the missing 429-stops-fallback tests for `v1QueryDataSource`, `v1UpdateDataSource`, and `v1FetchDatabaseOrDataSource` (discussion_r3335291076), and the children > 100 guard placement in `v1CreatePageWithContent` (discussion_r3335291085). Both are real — I agree with that feedback. Nothing new blocking from this pass.

PR #181, opened 2026-06-02
fix(wasm): serialize Module calls and guard against use-after-close

daltoniam:
## Summary

- Fixes the two recent production crashes — a `nil pointer dereference` at `wasm/memory.go:51` and a `fatal error: traceback / unexpected SPWRITE function runtime.morestack` — both rooted in concurrent invocation of a single wazero `api.Module`.
- Adds a per-`Module` call mutex around every entry point that touches `m.mod` or `m.fn*` (`Execute`, `Configure`, `Tools`, `Name`, `Healthy`, `Metadata`) and an `atomic.Bool` `closed` flag so racing callers get a clean `ErrModuleClosed` instead of dereferencing a freed wazero memory instance.

## Why

Wazero documents (`api/wasm.go:378-381`) that `Function.Call` is **not goroutine-safe** and must not be reentered until the previous call returns. A single `Module` shares one linear memory and one malloc/free heap, so each `malloc -> Memory().Write -> Call -> Memory().Read -> free` sequence must run as one atomic critical section.

Goja scripts fire many `api.call()` invocations from a single tool call, and the MCP server handles concurrent JSON-RPC requests against shared wasm-backed integrations. Without serialization, malloc/free interleaved with `Memory().Read` and produced the nil-pointer panic in `readFromGuest`. When that panic crossed Goja's `runTryInner` protected boundary while another panic was active, the runtime re-panicked and aborted with the `morestack` traceback.

`Loader.LoadPlugin` also closes the old module while another goroutine may be mid-call. The new `closed` flag makes those callers fail with `ErrModuleClosed` rather than touching a freed memory instance.

## Test plan

- [x] New `TestModule_ConcurrentExecute` (32 goroutines × 8 iterations) — reproduces the panic without the fix; passes with `-race` after the fix.
- [x] New `TestModule_ConcurrentMixed` exercises `Execute`/`Tools`/`Name`/`Healthy` in parallel.
- [x] New `TestModule_ExecuteAfterClose` — calls after `Close` return error, never panic.
- [x] New `TestModule_ConcurrentExecuteAndClose` races `Close` against in-flight `Execute` calls; asserts no goroutine panics.
- [x] `make ci` (build + vet + test-race + lint + gosec + govulncheck) clean.

## Follow-ups (out of scope)

- For higher parallelism we could instantiate one wazero module per concurrent caller (pool). Correctness is now sound with serialization; per-instance pooling is a perf optimization for a future PR.

### DIFF
--- a/wasm/concurrent_test.go
+++ b/wasm/concurrent_test.go
@@ -1,0 +1,193 @@
+package wasm
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"testing"
+
+	mcp "github.com/daltoniam/switchboard"
+)
+
+// TestModule_ConcurrentExecute verifies that concurrent calls into a single
+// *Module do not panic or corrupt guest memory. Wazero's api.Function.Call is
+// not goroutine-safe, so Module must serialize calls internally.
+//
+// Reproduces the production panic at wasm/memory.go:51 where Goja scripts
+// firing many api.call() invocations in parallel hit nil pointer derefs
+// inside readFromGuest.
+func TestModule_ConcurrentExecute(t *testing.T) {
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`[{"id":"item-1","name":"widget"}]`))
+	}))
+	defer srv.Close()
+
+	mod := loadTestModule(t)
+	if err := mod.Configure(context.Background(), mcp.Credentials{
+		"base_url": srv.URL,
+		"api_key":  "test-key",
+	}); err != nil {
+		t.Fatalf("Configure: %v", err)
+	}
+
+	const goroutines = 32
+	const iterations = 8
+
+	var wg sync.WaitGroup
+	errCh := make(chan error, goroutines*iterations)
+	for g := range goroutines {
+		wg.Add(1)
+		go func(id int) {
+			defer wg.Done()
+			for i := range iterations {
+				result, err := mod.Execute(context.Background(), "example_echo", map[string]any{
+					"message": "hello",
+				})
+				if err != nil {
+					errCh <- err
+					return
+				}
+				if result.IsError {
+					errCh <- &execErr{data: result.Data}
+					return
+				}
+				if !strings.Contains(result.Data, "hello") {
+					errCh <- &execErr{data: result.Data}
+					return
+				}
+				_ = id
+				_ = i
+			}
+		}(g)
+	}
+	wg.Wait()
+	close(errCh)
+	for err := range errCh {
+		t.Errorf("concurrent Execute error: %v", err)
+	}
+}
+
+// TestModule_ConcurrentMixed exercises Execute, Tools, Name, Healthy in parallel.
+// All of these touch m.mod and must be serialized.
+func TestModule_ConcurrentMixed(t *testing.T) {
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path == "/health" {
+			_, _ = w.Write([]byte(`{"status":"ok"}`))
+			return
+		}
+		_, _ = w.Write([]byte(`[{"id":"item-1","name":"widget"}]`))
+	}))
+	defer srv.Close()
+
+	mod := loadTestModule(t)
+	if err := mod.Configure(context.Background(), mcp.Credentials{
+		"base_url": srv.URL,
+		"api_key":  "test-key",
+	}); err != nil {
+		t.Fatalf("Configure: %v", err)
+	}
+
+	const goroutines = 16
+	var wg sync.WaitGroup
+	for g := range goroutines {
+		wg.Add(1)
+		go func(id int) {
+			defer wg.Done()
+			switch id % 4 {
+			case 0:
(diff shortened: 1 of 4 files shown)


wasm/concurrent_test.go
```diff
@@ -0,0 +1,193 @@
...
+		t.Fatalf("Configure: %v", err)
+	}
+
+	if err := mod.Close(ctx); err != nil {
+		t.Fatalf("Close: %v", err)
+	}
+
+	// Subsequent calls must not panic.
+	result, err := mod.Execute(ctx, "example_echo", map[string]any{"message": "x"})
+	if err == nil && (result == nil || !result.IsError) {
```
acmacalister: The condition `err == nil && (result == nil || !result.IsError)` would silently pass if `Execute` ever returned `(nil, someErr)` instead of the expected `{IsError: true}` result — the test would see `err != nil` and short-circuit without checking the result. Since the behaviour here is deterministic (closed module always returns `{IsError: true}, nil`), it doesn't matter today, but it's a subtle trap for anyone who later changes the error path.

Splitting the assertion makes the intent unambiguous:

```go
if err != nil {
	t.Fatalf("Execute after Close returned unexpected Go error: %v", err)
}
if result == nil || !result.IsError {
	t.Fatal("expected Execute after Close to return an IsError result")
}
```

wasm/concurrent_test.go
```diff
@@ -0,0 +1,193 @@
...
+		t.Fatal("expected Execute after Close to return an error")
+	}
+
+	if mod.Healthy(ctx) {
+		t.Error("Healthy() should return false after Close")
+	}
+	if tools := mod.Tools(); tools != nil {
+		t.Errorf("Tools() should return nil after Close, got %d tools", len(tools))
+	}
```
acmacalister, line 157: `TestModule_ExecuteAfterClose` checks `Execute`, `Healthy`, and `Tools`, but `Name()` and `Configure()` go through the same double-check pattern and aren't covered here. Two more assertions would complete the after-close coverage:

```go
if got := mod.Name(); got != "unknown" {
	t.Errorf("Name() after Close = %q, want "unknown"", got)
}
if err := mod.Configure(ctx, mcp.Credentials{}); !errors.Is(err, ErrModuleClosed) {
	t.Errorf("Configure after Close: got %v, want ErrModuleClosed", err)
}
```

acmacalister (approved):
Solid fix — the double-checked lock + CAS pattern is correct, the concurrency comment in the struct doc is exactly what future readers need, and the test suite is comprehensive. Two minor suggestions inline.

wasm/concurrent_test.go
daltoniam: Done in c21b6c8 — split into separate Go-error and IsError-result assertions.

wasm/concurrent_test.go
daltoniam: Done in c21b6c8 — added Name() and Configure() after-close assertions.

wasm/concurrent_test.go
```diff
@@ -0,0 +1,206 @@
...
+			defer wg.Done()
+			switch id % 4 {
+			case 0:
+				_, _ = mod.Execute(context.Background(), "example_echo", map[string]any{"message": "x"})
+			case 1:
+				_ = mod.Tools()
+			case 2:
+				_ = mod.Name()
+			case 3:
+				_ = mod.Healthy(context.Background())
```
acmacalister, line 109: `Metadata()` acquires `callMu` through the same double-check pattern added here, but it's not in the switch. Since we're already spinning up 16 goroutines, adding a fifth case costs nothing and makes sure the race detector actually sees it:

```go
const goroutines = 20
// ...
case 4:
	_ = mod.Metadata(context.Background())
```

(Also update `switch id % 4` → `id % 5`.)

wasm/concurrent_test.go
```diff
@@ -0,0 +1,206 @@
...
+
+	if got := mod.Name(); got != "unknown" {
+		t.Errorf("Name() after Close = %q, want %q", got, "unknown")
+	}
+	if mod.Healthy(ctx) {
+		t.Error("Healthy() should return false after Close")
+	}
+	if tools := mod.Tools(); tools != nil {
+		t.Errorf("Tools() should return nil after Close, got %d tools", len(tools))
+	}
```
acmacalister, line 157: The PR summary lists `Metadata` alongside `Execute`, `Configure`, `Tools`, `Name`, and `Healthy` as methods protected by the double-check pattern, but it's the only one without an after-close assertion here. Worth rounding it out:

```go
if got := mod.Metadata(); got != nil {
	t.Errorf("Metadata() should return nil after Close, got %v", got)
}
```

acmacalister (approved):
Solid fix. The double-checked locking pattern is correct — setting `closed` before acquiring `callMu` ensures in-flight calls drain before wazero teardown, and the re-check under the lock closes the TOCTOU window. Just two small gaps in `Metadata()` coverage inline.

wasm/concurrent_test.go
daltoniam: Done in 1e22593 — added Metadata() as case 4 (goroutines bumped to 20, % 5).

wasm/concurrent_test.go
daltoniam: Done in 1e22593 — added Metadata() nil assertion after Close.

acmacalister (approved):
Solid fix. The double-check locking pattern is the right approach here, the concurrency model is clearly documented on the struct, and the Close() CAS-before-lock ordering correctly handles the race between in-flight callers and a concurrent teardown. Four tests covering concurrent execution, mixed method calls, after-close semantics across every entry point, and the Close-vs-Execute race — that is thorough coverage. Build, race-detected tests, and lint all clean.

PR #182, opened 2026-06-03
feat(server): allow integrations to raise the response cap per-tool

daltoniam:
## Summary

- Adds `PerToolMaxResponseBytesIntegration`, an optional interface that lets an integration raise the 50KB default response cap for a single tool without loosening the safety net for the rest of its surface. The integration-wide `MaxResponseBytesIntegration` override remains the fallback when no per-tool value is declared.
- Threads the tool name through both `responseLimitFor` callsites (`server.handleExecute` and `ProjectRouter.makeExecuteHandler`) so the new interface runs on both the standard and project-scoped paths.
- Puts the interface to use on `github_get_pull_diff` (raw unified diff, no projection/pagination knob, 1MB cap). The agents PR review flow has been hitting "Response exceeded 50KB" for normal-sized PRs (a 21-file/5k-line PR is ~200KB); this unblocks it without touching the other 195 GitHub tools.

## Motivation

The model kept calling `github_get_pull_diff`, Switchboard kept rejecting the 193KB diff response, and the agent ran out of recovery moves. Switchboard already had a per-integration cap override (Confluence at 256KB), but raising the cap integration-wide for GitHub would weaken the safety net for the other 195 tools that should stay compact. The per-tool override targets the one tool that legitimately needs more headroom.

## Test plan

- [x] `gofmt -w . && go vet ./...`
- [x] `go build ./...`
- [x] `go test ./... -count=1`
- [x] `golangci-lint run` (0 issues)
- [x] New unit tests pin both callsites: `TestHandleExecute_PerToolCap{Honored,DoesNotLeakToOtherTools,WinsOverIntegrationCap}` on the server path; `TestProjectRouter_ExecutePerToolCap` on the project router path; `TestMaxResponseBytesForTool` on the GitHub integration.

💘 Generated with Crush

### DIFF
--- a/integrations/github/github.go
+++ b/integrations/github/github.go
@@ -20,10 +20,19 @@ var maxBytesByTool = compactResult.MaxBytes
 
 // Compile-time interface assertions.
 var (
-	_ mcp.Integration                = (*integration)(nil)
-	_ mcp.FieldCompactionIntegration = (*integration)(nil)
-	_ mcp.ToolMaxBytesIntegration    = (*integration)(nil)
+	_ mcp.Integration                        = (*integration)(nil)
+	_ mcp.FieldCompactionIntegration         = (*integration)(nil)
+	_ mcp.ToolMaxBytesIntegration            = (*integration)(nil)
+	_ mcp.PerToolMaxResponseBytesIntegration = (*integration)(nil)
 )
+
+// githubPullDiffMaxResponseBytes raises the integration-wide response cap for
+// the github_get_pull_diff tool only. Raw unified diffs for real-world PRs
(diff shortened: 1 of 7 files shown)


acmacalister (approved):

acmacalister (approved):
Clean PR — builds, tests pass (race detector clean), lint 0 issues. The layered priority logic in `responseLimitFor` is exactly right, the compile-time assertion in the GitHub adapter is a nice safety net, and the test matrix covers all three meaningful scenarios (honored, isolation, wins-over-integration-cap) on both the main and project-router paths. LGTM.

PR #183, opened 2026-06-05
feat(cloudflare): add AI Gateway tools for token usage and logs

daltoniam:
## Summary

- Adds 6 read tools to the Cloudflare integration so callers can answer **"how many tokens did this AI Gateway burn, and on what?"** without leaving Switchboard. Today the integration covers Zones/DNS/Workers/Pages/R2/KV/D1/WAF/LBs/Accounts but has no AI Gateway surface at all, so token usage, cost, and prompt logs are invisible.
- The new tools wrap the documented `GET /accounts/{id}/ai-gateway/gateways[...]` endpoints; per-request `tokens_in`, `tokens_out`, `cost`, `model`, `provider`, `cached`, and `duration` are surfaced via `cloudflare_list_ai_gateway_logs`, with drill-down tools for the full log entry and the raw upstream request/response bodies.

## Tools added

| Tool | Endpoint | Purpose |
|---|---|---|
| ``cloudflare_list_ai_gateways`` | `GET /gateways` | Discover gateway IDs (entry point — "Start here") |
| ``cloudflare_get_ai_gateway`` | `GET /gateways/{id}` | Settings: cache TTL, rate limits, spend limits, DLP, guardrails |
| ``cloudflare_list_ai_gateway_logs`` | `GET /gateways/{id}/logs` | Per-request token usage, cost, model, provider, cache hit, success — filterable by date, provider, model, success, cached, feedback, free-text, plus sort + paging |
| ``cloudflare_get_ai_gateway_log`` | `GET /gateways/{id}/logs/{log_id}` | Single log entry |
| ``cloudflare_get_ai_gateway_log_request`` | `GET .../logs/{log_id}/request` | Raw upstream prompt body |
| ``cloudflare_get_ai_gateway_log_response`` | `GET .../logs/{log_id}/response` | Raw upstream completion body |

Field-compaction specs are added for both list tools so `tokens_in`, `tokens_out`, `cost`, `cached`, `duration`, etc. survive compaction while incidental fields are stripped — keeps log listings well under the per-tool response cap.

## Test plan

- [x] ``go vet ./...``
- [x] ``go tool golangci-lint run`` (0 issues)
- [x] ``go test ./...`` (all packages green, including the existing ``TestDispatchMap_AllToolsCovered`` / ``TestDispatchMap_NoOrphanHandlers`` / ``TestFieldCompactionSpecs_NoOrphanSpecs`` parity checks)
- [x] New ``ai_gateway_test.go`` with ``httptest`` coverage for each handler (request path, query-string filters, response shape) plus a compaction-spec presence check
- [ ] Smoke test against a real Cloudflare account (reviewer): `cloudflare_list_ai_gateways` → `cloudflare_list_ai_gateway_logs` → `cloudflare_get_ai_gateway_log_response`

## Follow-up

A second PR will close the remaining gaps in the Cloudflare integration (Workers AI inference, Vectorize, Queues, R2 object-level ops, Stream, Images, Access/Zero Trust, Tunnels, Email Routing, Logpush, Workers secrets/deploys/tail, Pages create-deploy, Hyperdrive, Page Rules, Notifications, API Tokens).

💘 Generated with Crush


integrations/cloudflare/compact.yaml
```diff
@@ -30,6 +30,33 @@ tools:
...
+            - result[].path
+            - result[].success
+            - result[].cached
+            - result[].cost
+            - result[].tokens_in
+            - result[].tokens_out
+            - result[].duration
+            - result[].status_code
+            - result[].step
+            - result[].created_at
```
acmacalister, line 60: `feedback` is stored per-log (the -1/0/1 rating that gets set after a request) and since it's exposed as a filter param it'll appear in real responses. Without it in the spec it gets silently stripped during compaction — a caller filtering by `feedback=-1` would get the right rows back but the `feedback` values themselves would be gone from the output.

Worth adding `- result[].feedback` here alongside `result[].created_at`.

acmacalister (approved):
Solid implementation — the filter passthrough test, compaction specs, and entry-point guidance are all done well. One thing worth adding before merge.

integrations/cloudflare/compact.yaml
daltoniam: Good catch — added `result[].feedback` to the `cloudflare_list_ai_gateway_logs` compaction spec in 90a7e84 so the rating survives compaction alongside the other per-log metrics.

acmacalister (approved):
Clean PR — builds, tests pass with race detector, lint clean, go.mod tidy. The 6-tool AI Gateway surface is well-structured and the filter handling in `listAIGatewayLogs` is particularly tidy. One minor note: the logs test uses `cost` as the `order_by` value but the Cloudflare API's documented valid set is `created_at/provider/model/model_type/success/cached` — worth swapping to a documented value so the test doubles as a usage example. Not blocking. LGTM.

PR #184, opened 2026-06-05
feat(cloudflare): expand coverage with Workers AI, Vectorize, Queues, Zero Trust, and more

daltoniam:
## Summary

Stacks on top of #183 (AI Gateway). Adds **50 new tools** across 13 Cloudflare API surfaces so agents can drive the rest of the platform without falling back to raw HTTP.

### New surfaces

| Surface | Tools |
|---|---|
| Workers AI | `list_ai_models`, `run_ai_model` |
| Vectorize v2 | `list/get/create/delete_vectorize_index`, `query_vectorize_index` |
| Queues | `list/get/create/delete_queue`, `send_queue_messages` |
| Hyperdrive | `list/get/create/delete_hyperdrive_config` |
| Workers extras | `list_worker_secrets`, `list_worker_deployments`, `get_worker_subdomain`, `list_worker_tails` |
| Pages extras | `create_pages_project`, `create_pages_deployment`, `list_pages_domains` |
| KV bulk | `bulk_delete_kv_values` |
| Stream | `list/get/delete_stream_video` |
| Images | `list/get/delete_image` |
| Zero Trust Access | `list_access_apps`, `list_access_app_policies`, `list_access_identity_providers` |
| Cloudflare Tunnel | `list/get/delete_tunnel` |
| Email Routing | `list_email_routing_rules`, `list_email_routing_addresses`, `get_email_routing_settings` |
| Logpush | `list/get/create_logpush_job` |
| Page Rules | `list/get/delete_page_rule` |
| Notifications | `list_notification_policies`, `list_notification_webhooks` |
| API Tokens | `list/get/delete_api_token` |

### Conventions followed

- Every list tool ships a `compact.yaml` field whitelist so multi-tool agent loops do not blow the response cap.
- Tool descriptions follow the three-tier discoverability pattern from `add-integration` SKILL.md (entry points say \"Start here\", drill-down tools say \"Use after X\", mutations include workflow hints).
- All handlers use the shared `mcp.NewArgs` / `mcp.ArgStr` / `mcp.OptInt` helpers (no local arg-coercion).
- Endpoint paths verified against Cloudflare API docs + `cf-terraforming` (e.g. `/access/apps` not `/applications`, `/cfd_tunnel` not `/tunnels`).

### Deferred (call out)

**R2 object-level operations** (put/get/delete object, multipart upload) are intentionally not included. They require S3-style SigV4 signing against a different host (`{account_id}.r2.cloudflarestorage.com`) with separate access key credentials. That is a much bigger surface (new credential fields, a SigV4 signer, separate transport) and deserves its own PR.

## Test plan

- [x] `go vet ./...` — clean
- [x] `golangci-lint run ./integrations/cloudflare/...` — 0 issues
- [x] `go test ./...` — all packages pass
- [x] Httptest-backed unit tests for at least one tool from every new surface, plus required-field error paths
- [x] Dispatch + compact-spec parity tests (`TestDispatchMap_*`, `TestFieldCompactionSpecs_*`) still pass

💘 Generated with Crush

### DIFF
--- a/integrations/cloudflare/compact.yaml
+++ b/integrations/cloudflare/compact.yaml
@@ -326,7 +326,6 @@ tools:
             - result[].dataset
             - result[].enabled
             - result[].name
-            - result[].destination_conf
             - result[].logpull_options
             - result[].last_complete
             - result[].last_error
--- a/integrations/cloudflare/compact.yaml
+++ b/integrations/cloudflare/compact.yaml
@@ -175,3 +175,199 @@ tools:
             - result[].plan.name
             - result[].activated_on
             - result[].modified_on
+    cloudflare_list_ai_models:
+        spec:
+            - result[].id
+            - result[].source
+            - result[].name
+            - result[].description
+            - result[].task.id
+            - result[].task.name
+            - result[].properties[].property_id
+            - result[].properties[].value
+    cloudflare_list_vectorize_indexes:
+        spec:
+            - result[].id
+            - result[].name
+            - result[].description
+            - result[].config.dimensions
+            - result[].config.metric
+            - result[].created_on
+            - result[].modified_on
+    cloudflare_list_queues:
+        spec:
+            - result[].queue_id
+            - result[].queue_name
+            - result[].created_on
+            - result[].modified_on
+            - result[].producers_total_count
+            - result[].consumers_total_count
+    cloudflare_list_hyperdrive_configs:
+        spec:
+            - result[].id
+            - result[].name
+            - result[].origin.host
+            - result[].origin.port
+            - result[].origin.database
+            - result[].origin.scheme
+            - result[].caching.disabled
+            - result[].caching.max_age
+            - result[].caching.stale_while_revalidate
+    cloudflare_list_worker_secrets:
+        spec:
+            - result[].name
+            - result[].type
+    cloudflare_list_worker_deployments:
+        spec:
+            - result[].id
+            - result[].source
+            - result[].strategy
+            - result[].author_email
+            - result[].created_on
+            - result[].versions[].version_id
+            - result[].versions[].percentage
+    cloudflare_list_worker_tails:
+        spec:
+            - result[].id
+            - result[].url
+            - result[].expires_at
+    cloudflare_list_pages_domains:
+        spec:
+            - result[].id
+            - result[].name
+            - result[].status
+            - result[].verification_data.status
+            - result[].validation_data.status
+            - result[].certificate_authority
+            - result[].created_on
+    cloudflare_list_stream_videos:
+        spec:
+            - result[].uid
+            - result[].meta.name
+            - result[].duration
+            - result[].size
+            - result[].status.state
+            - result[].readyToStream
+            - result[].thumbnail
+            - result[].preview
+            - result[].playback.hls
+            - result[].playback.dash
+            - result[].created
+            - result[].modified
+    cloudflare_list_images:
+        spec:
+            - result.images[].id
+            - result.images[].filename
+            - result.images[].meta
+            - result.images[].requireSignedURLs
+            - result.images[].variants
+            - result.images[].uploaded
+    cloudflare_list_access_apps:
+        spec:
+            - result[].id
+            - result[].uid
+            - result[].name
+            - result[].domain
+            - result[].type
+            - result[].session_duration
+            - result[].auto_redirect_to_identity
+            - result[].allowed_idps
+            - result[].created_at
+            - result[].updated_at
+    cloudflare_list_access_app_policies:
+        spec:
+            - result[].id
+            - result[].name
+            - result[].decision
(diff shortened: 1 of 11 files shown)


integrations/cloudflare/governance.go
```diff
@@ -0,0 +1,137 @@
...
+func listAPITokens(ctx context.Context, c *cloudflare, args map[string]any) (*mcp.ToolResult, error) {
+	acct, err := c.acctID(args)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	q := queryEncode(map[string]string{
+		"page":     fmt.Sprintf("%d", mcp.OptInt(args, "page", 1)),
+		"per_page": fmt.Sprintf("%d", mcp.OptInt(args, "per_page", 20)),
+	})
+	data, err := c.get(ctx, "/accounts/%s/tokens%s", acct, q)
```
acmacalister: These three token functions (`listAPITokens`, `getAPIToken`, `deleteAPIToken`) all use an `/accounts/%s/tokens` path that doesn't exist in the Cloudflare v4 API. Token management is user-scoped — the correct endpoints are `GET /user/tokens`, `GET /user/tokens/{token_id}`, and `DELETE /user/tokens/{token_id}`. The `account_id` parameter should be dropped from all three; it isn't used by the API at all.

```go
// listAPITokens
data, err := c.get(ctx, "/user/tokens%s", q)

// getAPIToken
data, err := c.get(ctx, "/user/tokens/%s", tokenID)

// deleteAPIToken
data, err := c.del(ctx, "/user/tokens/%s", tokenID)
```

As-is these will 404 on any live account.

integrations/cloudflare/expanded_coverage_test.go
```diff
@@ -0,0 +1,425 @@
...
+	assert.False(t, result.IsError)
+	assert.Contains(t, result.Data, "prod-alerts")
+}
+
+// --- API Tokens ---
+
+func TestListAPITokens(t *testing.T) {
+	c, ts := newTestClient(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, "GET", r.Method)
+		assert.Contains(t, r.URL.Path, "/accounts/test-acct/tokens")
```
acmacalister: This assertion validates the wrong path (same issue as the handler above) so it passes today but won't catch the live 404. Once the handler is fixed to `/user/tokens`, this should be:

```go
assert.True(t, strings.HasSuffix(r.URL.Path, "/user/tokens"))
```

Same applies to `TestDeleteAPIToken` below at the `/tokens/tok1` assertion.

Also: there's no `TestGetAPIToken` — each new tool should have at least a happy-path test covering the URL routing.

integrations/cloudflare/email_logpush.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	data, err := c.get(ctx, "/accounts/%s/logpush/jobs", acct)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	return mcp.RawResult(data)
+}
+
+func getLogpushJob(ctx context.Context, c *cloudflare, args map[string]any) (*mcp.ToolResult, error) {
+	jobID := mcp.OptInt(args, "job_id", 0)
```
acmacalister: This uses `OptInt` with a sentinel-zero check instead of the `r.Int()` + `r.Err()` pattern everything else in the adapter uses. The difference: `OptInt` also returns the default for negative values, so `job_id: -1` would silently produce `jobID == 0` and surface as "required" rather than forwarding -1 to the API. Probably harmless in practice (Logpush job IDs are always positive), but it's inconsistent and the error message would be confusing.

```go
r := mcp.NewArgs(args)
jobID := r.Int("job_id")
    return mcp.ErrResult(err)
}
```

integrations/cloudflare/compact.yaml
```diff
@@ -175,3 +175,199 @@ tools:
...
+            - result[].verified
+            - result[].created
+            - result[].modified
+    cloudflare_list_logpush_jobs:
+        spec:
+            - result[].id
+            - result[].dataset
+            - result[].enabled
+            - result[].name
+            - result[].destination_conf
```
acmacalister: Worth a conscious decision here: `destination_conf` can embed credentials in the URI — e.g. `s3://bucket/path?access-key-id=AKIA...&secret-access-key=...` or `https://collector?header_Authorization=Bearer+xyz`. Whitelisting it means those values flow into LLM context verbatim. If the primary use case is auditing job destinations (not managing credentials), keeping only the scheme+host portion would be safer. If agents do need to inspect or clone full configs, the current spec is reasonable — just flagging the trade-off.

acmacalister (commented):
Great work on this — 50 tools across 13 surfaces with clean code, full compaction specs, and solid test coverage throughout. Found one live-failure bug in the API token endpoints and a few smaller things inline.

PR #185, opened 2026-06-06
feat(github): add runner registration token tools

daltoniam:
## Summary

- Adds repository and organization GitHub Actions runner registration-token tools so Switchboard can bootstrap self-hosted runners.
- Wires the tools through GitHub dispatch/metadata and covers both REST endpoints with integration tests.

## Test plan

- [x] go test ./integrations/github
- [x] go test ./...
- [x] make ci

💘 Generated with Crush


acmacalister (approved):

integrations/github/tools.go
```diff
@@ -799,11 +799,21 @@ var tools = []mcp.ToolDefinition{
 		Parameters: map[string]string{"owner": "Repository owner", "repo": "Repository name", "page": "Page number", "per_page": "Results per page"},
 		Required:   []string{"owner", "repo"},
 	},
+	{
+		Name: mcp.ToolName("github_create_runner_registration_token"), Description: "Create a short-lived registration token for adding a self-hosted runner to a repository",
```
acmacalister: The description is accurate but an LLM receiving the response won't automatically know what to do with the returned `token` value. Worth adding a short note — something like: "...returns a short-lived token and its expiry time; pass the token to the runner's `./config.sh --token <value>` command to complete registration. Token is valid for ~1 hour." Same applies to the org variant below on line 813.

acmacalister (approved):
Clean PR — builds, all tests pass, lint is clean. One small suggestion inline about making the tool description more self-contained for LLM use.

integrations/github/tools.go
daltoniam: Updated both repo and org runner token tool descriptions with the config.sh --token usage and ~1 hour validity note.

acmacalister (approved):
Clean PR — builds, all tests pass with the race detector, lint is clean. Both tools follow the established handler pattern precisely, dispatch parity tests cover them automatically, and the descriptions now include the config.sh usage guidance. LGTM.

PR #186, opened 2026-06-06
feat(digitalocean): support droplet user data

daltoniam:
## Summary

- Adds optional `user_data` support to DigitalOcean droplet creation so Switchboard can pass cloud-init bootstrap scripts.
- Documents the new parameter in the droplet creation tool metadata and covers the request payload with a regression test.

## Test plan

- [x] go test ./integrations/digitalocean
- [x] go test ./...
- [x] make ci

💘 Generated with Crush

### DIFF
--- a/integrations/digitalocean/digitalocean_test.go
+++ b/integrations/digitalocean/digitalocean_test.go
@@ -2,8 +2,10 @@ package digitalocean
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
+	"net/http/httptest"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


acmacalister (approved):
Clean PR — builds, all tests pass (race detector clean), lint passes, go mod tidy. The `user_data` plumbing follows the exact same optional-arg pattern as `vpc_uuid`, `tags`, and `ssh_keys` already in `createDroplet`, and the `httptest` regression test properly captures the outgoing payload to verify it round-trips. LGTM.

PR #190, opened 2026-06-16
fix: allow longer HTTP calls from WASM plugins

daltoniam:
- Increase the WASM host HTTP timeout from 30s to 60s so slower internal tools can return normally instead of failing at the transport layer.

Tests: `gofmt -w . && go test ./...`

💘 Generated with Crush


acmacalister (approved):

PR #192, opened 2026-06-24
Add multi-cluster ClickHouse connections

daltoniam:
- Enables ClickHouse users to configure multiple cluster aliases and choose a target cluster per tool call without changing existing default-connection behavior.
- Adds the setup UI, config defaults, tool descriptions, compaction, docs, and tests needed to make multi-cluster use discoverable and safe.

💘 Generated with Crush


acmacalister (approved):
Clean PR — CI/status checks are passing (build, test, lint, security all green) and no blocking issues found. Multi-cluster ClickHouse support follows established patterns (dispatch parity, compaction, setup UI). LGTM.

PR #193, opened 2026-06-24
Add read-only Kubernetes integration

daltoniam:
- Adds a read-only Kubernetes integration for local cluster discovery, workload inspection, logs, events, services, ingresses, deployments, and nodes using the Kubernetes API.
- Refreshes search after hot credential updates so newly configured integrations are discoverable immediately, without requiring a server restart.

💘 Generated with Crush


acmacalister (approved):
Clean PR — CI/status checks are passing (build, test, lint, security all green). New kubernetes adapter follows all project conventions (dispatch parity, CredentialDetector, compaction). RefreshSearchIndex hook correctly enables hot credential updates. LGTM.

PR #196, opened 2026-06-26
feat(postgres): query databases through a hosted agent tunnel

daltoniam:
## Summary
- Adds an **agent connection mode** to the Postgres integration so it can run read-only queries against a customer's database without ever holding the database credentials. The real DSN stays on the customer-side agent; queries are forwarded to a hosted relay that proxies to that agent and returns rows.
- All existing Postgres tools (`query`, `select`, `explain`, `list_tables`, etc.) work identically in agent mode. Direct connections are unchanged. Writes are rejected over the tunnel because the relay enforces read-only access.

## How it works
A new `runner` interface abstracts statement execution. `sqlRunner` is the existing direct `*sql.DB` path (byte-identical behavior, including read-only transaction wrapping). `tunnelRunner` speaks plain HTTP/JSON to the hosted relay's query endpoint and converts its column/row-array response into the same row-object JSON the tools already emit, so output is identical regardless of mode.

Agent mode is selected with credentials `mode=agent`, `tunnel_url`, `org`, and optional `tunnel_auth_token` — no DSN.

## Tests
- `runner_test.go`: tunnel request/response transform, empty results, single-row, server-error propagation, read-only write rejection, ping, endpoint normalization, and agent-mode `Configure`/`Healthy` against an httptest relay.

### DIFF
--- a/integrations/postgres/postgres.go
+++ b/integrations/postgres/postgres.go
@@ -35,9 +35,11 @@ var (
 	_ mcp.ToolMaxBytesIntegration    = (*postgres)(nil)
 )
 
-// pgConn holds a single Postgres connection and its metadata.
+// pgConn holds a single Postgres connection and its metadata. Statement
+// execution goes through runner, which is either a direct *sql.DB (sqlRunner)
+// or a hosted tunnel relay (tunnelRunner) when mode=agent.
 type pgConn struct {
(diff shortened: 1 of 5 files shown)


acmacalister (approved):
CI/status checks are passing. Clean PR — no blocking issues found.

PR #197, opened 2026-06-26
Move integration tool descriptions out of Go and into YAML

arjansingh:
## Summary

Every integration's tool descriptions lived inline in Go, about 1,400 tools across the adapters. Editing the wording for one tool was a code change, reviewed next to dispatch maps, compaction specs, and everything else in the file. Prose was slow to write and the reviews were noisy.

This PR moves all of it into a `tools.yaml` file per adapter. Editing a description is now a content change you can review on its own, with no Go in the diff. The tool list the server hands to the model stays byte-for-byte the same as before.

## What was built

**tools.yaml per adapter.** Each adapter's tool code is now a few lines that load a YAML file. Descriptions, parameters, and required flags all live in the YAML. Every adapter with a fixed tool set moved over.

**A stricter parameter type.** Parameters and required-ness used to be two separate collections, cross-referenced by string name. A typo in the required list compiled fine and quietly changed the tool's contract. Now each parameter carries its own required flag, so a parameter that is required but never declared is no longer something you can write down.

**Prose moved for the two meta-tools too.** The long search and execute descriptions that teach the model how to discover and call tools moved into markdown files, same idea: code owns behavior, content owns the words.

**A loud loader.** A misspelled key fails at startup instead of quietly shipping a tool with a blank description.

## The wire lock

The thing that makes this safe is a lock file on the tool list the server hands the model. It works like `go.sum`. The `tools.yaml` files are the source of truth, and `server/tools_list.lock.json` is the generated, committed snapshot of the exact wire output they produce. A test, `TestToolsList_MatchesWireLock`, compares the live output against the lock byte-for-byte on every run, and CI runs it on every PR.

That is what lets you trust this migration. Across all the adapters, the model sees the same tool list it saw before. Nothing was dropped or reworded by accident.

It also stays useful after the migration. The lock catches any future change to what the model sees, intended or not: a dropped description, a renamed parameter, a flipped required flag all turn the test red. When the change is intentional, you regenerate the lock and commit it with the YAML:

```
go test ./server -run TestToolsList_MatchesWireLock -update
```

You don't hand-edit the lock, the same way you don't hand-edit `go.sum`. An accidental change shows up as a lock diff in review instead of shipping quietly.

## Design decisions

We kept the wire output identical on purpose. Nothing downstream of the server had to change, and the lock holds us to it.

Adapters that build their tools at runtime from a remote server keep doing that. Only the ones with a fixed tool set moved to YAML.

Required-ness now lives on each parameter. That also fixed a real bug: when search lifts a shared parameter into a shared block, it no longer loses whether a given tool needed it.

## Test plan

- [x] `make build`, `go vet`, and the full `go test -race` suite green
- [x] wire lock test passes across all adapters, and `-update` regenerates it cleanly
- [ ] CI lint and security (run on the PR)

## Note

This branch was rebased onto the latest main, which had added twelve adapters since it diverged (the Google Workspace suite, Kubernetes, and Vercel). Those are migrated to YAML here too, so the whole adapter set lands on one shape. There is also a small unrelated fix in here for health probes hanging on an unreachable integration.

Co-Authored-By: claude-opus-4-7 <noreply@anthropic.com>


acmacalister (approved):
CI all green (build, test, lint, security). PR moves 1,400+ tool descriptions to YAML with byte-identity snapshot test and no Must Fix or Should Fix issues found. LGTM.

acmacalister (approved):
Clean PR. All CI checks passed (build, test, lint, security). The YAML migration is thorough, the wire lock enforces parity, and the architecture patterns are followed correctly. LGTM.

PR #199, opened 2026-07-06
Fix RWX integration against live API schema

aleksclark:
## Summary

The RWX read-path tools were validated only against fixtures that used a flat status schema the real API never returns — so every tool failed in production despite green tests. Discovered while running a full monitor/troubleshoot pass against a real failed run.

- **`rwx_get_run_results`**: parse nested `run_status`/`task_status`/`polling` instead of flat `result_status`/`completed`. Failed runs were misreported as `running` and never flagged as errors.
- **`parseResultsPrompt`**: match singular `# Failed task:` headers so `failed_tasks` (and their task-ids) surface for single-failure runs — previously agents got no task-id to pull logs.
- **`downloadLogArchive`**: POST `application/x-www-form-urlencoded` (the archive endpoint rejects multipart), fixing all log tools (`get`/`grep`/`tail`/`head`) which were 500ing.
- **`rwx_get_artifacts`**: artifacts are task-scoped; require `task_id` or `run_id`+`task_key` and clarify the description (run_id-only returned a confusing "Task not found").

## Test plan

- [x] `make ci` (build, vet, race tests, lint, security) passes
- [x] Tests updated to use the real nested schema and assert the urlencoded log-download request
- [x] Validated live end-to-end through the local switchboard MCP: recent runs → run results (`failure`/`completed`, populated `failed_tasks`) → task logs pinpointing the root cause

💘 Generated with Crush

### DIFF
--- a/integrations/rwx/api.go
+++ b/integrations/rwx/api.go
@@ -7,7 +7,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"mime/multipart"
 	"net/http"
 	"net/url"
 	"path"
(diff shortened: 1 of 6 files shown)


acmacalister:
Clean PR — all CI checks passed (build ✓, test ✓, lint ✓, security ✓). Reviewed the diff: proper nested status schema handling, urlencoded form fix for log downloads, task-scoped artifact validation, and comprehensive test coverage for real API shapes. No blocking issues found. LGTM.

PR #202, opened 2026-07-10
fix(search): return matches when the tool catalog is small

daltoniam:
## Problem

Tool search ranks candidates with TF-IDF. Plain IDF (`log(N/df)`) is zero for any term that appears in **every** tool — `log(N/N) = 0`. In a small catalog every query term is effectively universal, so all candidates score zero, get filtered by the `score > 0` cut, and **search returns nothing even when the query obviously matches a tool.**

This bites two real deployments:
- a gateway configured with a single integration (or a handful of tools), and
- a multi-tenant host whose per-user policy narrows the visible catalog down to a few tools.

## Fix

Smooth the IDF weight to `log(N/df) + 1`. A matched term now always contributes a positive floor, so a matching tool is never dropped. The `+1` is a constant shift, so **relative ranking is unchanged** — a rarer term still outranks a common one.

## Tests

- Unit: `computeIDF` keeps universal words positive (single-tool corpus, word shared by all tools).
- End-to-end: `handleSearch` on a one-tool and a two-tool catalog returns the matching tools. Both new tests were confirmed to **fail on the old formula** and pass with the fix.
- Full `./server` suite + race detector + `golangci-lint` clean.

💘 Generated with Crush

### DIFF
--- a/server/search.go
+++ b/server/search.go
@@ -179,7 +179,17 @@ func buildSynonymMap(groups [][]string) map[string][]string {
 // computeIDF builds the inverse document frequency map from the tool corpus
 // and pre-computes each tool's token set (stored on toolWithIntegration.tokens).
 // Only words appearing in at least one tool are indexed.
-// Formula: IDF(word) = log(totalTools / toolsContainingWord)
+//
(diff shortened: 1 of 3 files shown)


acmacalister (approved):

PR #203, opened 2026-07-10
feat(server): allow appending deployment-specific MCP instructions

daltoniam:
## Why

The MCP `initialize` response tells connecting clients how to use the gateway (search before execute, session context, pinning, etc.). That guidance was fixed, so a deployment had no way to add notes that only apply to some tenants — for example that a particular organization has private-resource tunnels and how to target a specific one.

## What changed

- Add a `WithExtraInstructions(text)` server option that appends deployment-specific guidance to the standard instructions.
- The open-source default is unchanged: with no option set, clients receive exactly the same instructions as before.

## Tests

Added a test that drives the initialize handshake over an in-memory transport and asserts the base instructions are always present and that supplied extra text is appended only when the option is used.

## Context

Enables a hosted deployment to tell assistants, per organization, that private tunnels exist and how to reach them — the client-facing half of making multi-tunnel querying discoverable.

💘 Generated with Crush

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -69,21 +69,36 @@ const (
 
 // Server wraps the MCP SDK server and exposes search/execute tools.
 type Server struct {
-	mcpServer        *mcpsdk.Server
-	services         *mcp.Services
-	scriptEngine     *script.Engine
-	sessionStore     SessionStore
-	retryBackoff     time.Duration
-	breakers         map[string]*breaker
-	breakerThreshold int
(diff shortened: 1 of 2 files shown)


acmacalister (approved):
All CI checks passed (build, test, lint, security). Clean diff — follows existing Option pattern, extracts baseInstructions cleanly, and the in-memory transport test verifies the initialize handshake. LGTM.

PR #204, opened 2026-07-12
docs: use full Google OAuth scope URLs in the setup guide

daltoniam:
## Summary

- The Google Workspace setup guide abbreviated each scope as `.../auth/...` with a footnote explaining `...` meant `https://www.googleapis.com`. Google's Cloud Console rejects those shortened strings as **invalid** when pasted into the "Manually add scopes" box, so users following the guide can't add the scopes.
- This lists the full scope URLs so they can be pasted directly, and adds a one-line note calling out the "Manually add scopes" gotcha.

Docs-only; the actual scopes requested by the code (`googleoauth/family.go`) were already the full URLs — this only fixes the documentation.

💘 Generated with Crush

### DIFF
--- a/docs/google-setup.md
+++ b/docs/google-setup.md
@@ -111,23 +111,26 @@ Then complete the sign-in from the web UI to obtain tokens.
 
 ## Scopes requested per service
 
+Enter these full scope strings when Google's console asks you to *Manually add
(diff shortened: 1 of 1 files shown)


acmacalister (approved):
CI passed (build, test, lint, security, rust-sdk). Clean docs-only change fixing the Google OAuth scope URLs. LGTM.

PR #205, opened 2026-07-16
Add bounded per-request timeouts for WASM HTTP calls

daltoniam:
## Summary

- WASM integrations currently share a hard 60-second HTTP client timeout, which forces long-running tools to poll even when their upstream API supports a longer blocking request.
- Add an opt-in `X-Host-Timeout-Seconds` control header. It applies only to that host request, defaults to 60 seconds when absent/invalid, and is capped at 300 seconds.
- Consume the control header in the host so it is never leaked to upstream services; both HTTP/1.1 and h2c transports retain their existing defaults.

## Why

Overmind's child-agent cohort endpoint can wait for specialists server-side, but must currently return after 55 seconds to stay below this 60-second host limit. A bounded per-request override lets that one tool wait nearly five minutes without making every GitHub, Slack, or other integration call slower to fail.

## Test plan

- [x] Override times out a deliberately slow request at one second
- [x] The same request succeeds with a two-second override
- [x] Control header is not forwarded upstream
- [x] Invalid/zero values retain the 60-second default
- [x] Values above 300 seconds are clamped
- [x] `make ci`

### DIFF
--- a/wasm/host.go
+++ b/wasm/host.go
@@ -11,6 +11,7 @@ import (
 	"log/slog"
 	"net"
 	"net/http"
+	"strconv"
 	"time"
 
 	"github.com/tetratelabs/wazero/api"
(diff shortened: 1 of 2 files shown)


acmacalister (approved):
CI passed (build, test, lint, security all SUCCESS). Clean PR — no blocking issues found.

PR #206, opened 2026-07-28
Add Ramp corporate card and bill pay integration

daltoniam:
## Summary
- Adds a Ramp adapter for corporate card spend, reimbursements, Bill Pay, cards, users, merchants/vendors, and org structure
- Uses Bearer auth (`RAMP_ACCESS_TOKEN`) against the Developer API with optional sandbox base URL override
- Includes field compaction specs, dispatch parity tests, and config/env wiring

## Test plan
- [x] `make build vet test-race lint` (govulncheck failures are pre-existing on main)
- [ ] Enable Ramp in config with a valid token and run `search` for `ramp transactions`
- [ ] `execute ramp_list_transactions` and `ramp_list_users` against sandbox or production


acmacalister (approved):
CI passed (build/test/lint/security all green). Clean adapter implementation following all patterns: dispatch parity, compaction, shared helpers, entry-point guidance. No issues found. LGTM.

integrations/ramp/tools.go
```diff
@@ -0,0 +1,177 @@
...
+		},
+	},
+	{
+		Name: mcp.ToolName("ramp_get_transaction"), Description: "Get details of a specific Ramp card transaction including merchant, amount, receipts, and accounting fields. Use after list_transactions.",
+		Parameters: map[string]string{"transaction_id": "Transaction ID"},
+		Required:   []string{"transaction_id"},
+	},
+	{
+		Name: mcp.ToolName("ramp_update_transaction"), Description: "Update a Ramp transaction memo or accounting fields. Use after get_transaction.",
+		Parameters: map[string]string{
```
acmacalister, line 40: `ramp_update_transaction` doesn't match the live Ramp Developer API.

Tool text (and the handler test) treat `data` as a free-form PATCH body for memo/accounting fields, e.g. `{"memo":"Client dinner"}`. Per the current OpenAPI spec, `PATCH /developer/v1/transactions/{transaction_id}` only accepts `line_items` for split/unsplit:

```json
{ "line_items": [ { "amount": 4000, "memo": "Case-67890" }, ... ] }
```

Memo updates are a different endpoint: `POST /developer/v1/memos/{transaction_id}`.

As written, a natural "set memo" call will 400 against production while unit tests still pass against the permissive mock. Worth either:

1. Retargeting this tool at the memos endpoint (and renaming/describing it accordingly), or
2. Keeping the PATCH but documenting/validating the `line_items` split body only, and adding a separate memo tool.

```go
// memo update
data, err := r.doRequest(ctx, http.MethodPost,
    fmt.Sprintf("/developer/v1/memos/%s", url.PathEscape(id)), body)

// split update (current path)
data, err := r.patch(ctx,
    fmt.Sprintf("/developer/v1/transactions/%s", url.PathEscape(id)), body)
```

integrations/ramp/handlers.go
```diff
@@ -0,0 +1,313 @@
...
+	dataStr := rd.Str("data")
+	if err := rd.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	var body any
+	if err := json.Unmarshal([]byte(dataStr), &body); err != nil {
+		return mcp.ErrResult(fmt.Errorf("invalid JSON for data: %w", err))
+	}
+	path := fmt.Sprintf("/developer/v1/transactions/%s", url.PathEscape(id))
+	data, err := r.patch(ctx, path, body)
```
acmacalister: This always PATCHes `/developer/v1/transactions/{id}` with whatever JSON the model put in `data`. That endpoint is split-only (`line_items`); top-level `memo` / accounting fields aren't valid there.

If we want "update memo" behavior from the tool description, this needs to hit `POST /developer/v1/memos/{transaction_id}` (or we should tighten the tool to splits only and reject other shapes before the API call).

integrations/ramp/tools.go
```diff
@@ -0,0 +1,177 @@
...
+			"card_id":       "Filter by card ID",
+			"department_id": "Filter by department ID",
+			"location_id":   "Filter by location ID",
+			"merchant_id":   "Filter by merchant ID",
+			"state":         "Transaction state filter",
+			"sync_status":   "Accounting sync status filter",
+			"entity_id":     "Filter by entity ID",
+			"page_size":     "Results per page (2-100, default 20)",
+			"start":         "Pagination cursor (ID of last entity from previous page)",
+		},
```
acmacalister, line 22: Pagination docs here don't quite match what Ramp returns.

`start` is described as "ID of last entity from previous page", but list responses use `DeveloperAPINestedPage`, where `page.next` is a full URL like `https://api.ramp.com/developer/v1/transactions?start=...&page_size=20`, not a bare entity ID.

An agent that copies `page.next` into `start` (or tries to reuse the full URL) will fail. We should either:

- Document that callers must pull the `start` query value out of `page.next`, or
- Accept a `next` arg and follow/parse that URL in the handler.

Compaction already surfaces `page.next`, so aligning the tool param story with that field would make multi-page workflows much more reliable.

config/config_test.go
```diff
@@ -32,7 +32,7 @@ func TestLoad_CreatesDefaultWhenMissing(t *testing.T) {
 	_, err = os.Stat(path)
 	assert.NoError(t, err)
 
-	assert.Len(t, m.cfg.Integrations, 48)
+	assert.Len(t, m.cfg.Integrations, 49)
 	for _, name := range []string{"github", "datadog", "linear", "sentry", "slack", "metabase", "aws", "posthog", "postgres", "clickhouse", "elasticsearch", "pganalyze", "rwx", "gmail", "gcal", "gdrive", "gdocs", "gsheets", "gslides", "gforms", "gchat", "gmeet", "gtasks", "gpeople", "notion", "ollama", "ynab", "stripe", "gcp", "suno", "amazon", "jira", "confluence", "salesforce", "cloudflare", "digitalocean", "fly", "kubernetes", "vercel", "snowflake", "acp", "web", "botidentity", "x", "signoz", "nomad", "agents", "switchboard"} {
```
acmacalister: Count went to 49, but `"ramp"` never got added to this name list (same gap in `TestDefaultConfig`'s `expected` credential map). The length assertion will pass even if ramp's default entry is missing, as long as something else fills the slot.

Worth adding `"ramp"` here and `"ramp": {"access_token", "base_url"}` to the expected map so the new integration is actually asserted by name.

acmacalister (commented):
Solid adapter skeleton — dispatch parity, compaction, shared helpers, config/env wiring, and CI (build/test/lint/security) are all green.

One correctness issue against the live Ramp API on transaction updates, plus a couple of pagination/test-coverage nits inline.

integrations/ramp/tools.go
```diff
@@ -0,0 +1,197 @@
...
+			"user_id":       "Filter by cardholder user ID",
+			"card_id":       "Filter by card ID",
+			"department_id": "Filter by department ID",
+			"location_id":   "Filter by location ID",
+			"merchant_id":   "Filter by merchant ID",
+			"state":         "Transaction state filter",
+			"sync_status":   "Accounting sync status filter",
+			"entity_id":     "Filter by entity ID",
+			"page_size":     "Results per page (2-100, default 20)",
+			"start":         "Pagination cursor (ID of last entity from previous page)",
```
acmacalister: Looks like the pagination cleanup in `17aef14` landed on most list tools but missed this one (and `ramp_list_users` below).

`pageParams` already accepts `next` and pulls `start`/`page_size` out of `page.next`, and `TestListTransactions_NextURL` covers that path — but the tool schema still only advertises `start` as "ID of last entity from previous page". Since this is the Start-here entry point, agents will keep guessing the wrong cursor shape here even though the handler can do the right thing.

Same shape as the other list tools would close the loop:

```go
"page_size": "Results per page (2-100, default 20)",
"start":     "Pagination start token from the previous page URL query string",
"next":      "Full page.next URL from a previous list response (preferred over start)",
```

integrations/ramp/tools.go
```diff
@@ -0,0 +1,197 @@
...
+		Parameters: map[string]string{"bill_id": "Bill ID"},
+		Required:   []string{"bill_id"},
+	},
+	{
+		Name: mcp.ToolName("ramp_list_users"), Description: "List Ramp employees, cardholders, and users for the business",
+		Parameters: map[string]string{
+			"department_id": "Filter by department ID",
+			"location_id":   "Filter by location ID",
+			"page_size":     "Results per page (2-100, default 20)",
+			"start":         "Pagination cursor",
```
acmacalister: Same leftover as `ramp_list_transactions`: still documents a bare "Pagination cursor" and doesn't expose `next`, while every other list tool (and `pageParams`) already support the full `page.next` URL flow.

Worth aligning so user pagination matches the rest of the adapter.

acmacalister (approved):
Nice follow-up on the earlier review — splitting memo vs splits onto the real Ramp endpoints, adding `next` parsing, and tightening the config tests were the right fixes. CI is green (build/test/lint/security/rust-sdk).

One leftover from the pagination pass: `ramp_list_transactions` and `ramp_list_users` still have the old `start`-only docs even though the handler already supports `next`. Two small schema nits inline; otherwise LGTM.

integrations/ramp/compact.yaml
```diff
@@ -0,0 +1,238 @@
...
+        spec:
+            - data[].id
+            - data[].amount
+            - data[].currency
+            - data[].state
+            - data[].user_id
+            - data[].merchant
+            - data[].transaction_date
+            - data[].created_at
+            - page.next
```
acmacalister, line 63: Checked these paths against the current Ramp OpenAPI (`Reimbursement`): there is no top-level `amount` or `currency` anymore. Amounts live on:

- `entity_amount` / `merchant_amount` → `ApiAmount` (`value`, `currency`)
- `payee_amount` → `CurrencyAmount` (`amount`, `currency_code`)

With the current include list, compacted list/get responses drop the money fields entirely (id/state/merchant survive, spend does not), which is rough for the main reimbursement workflows. Something closer to:

```yaml
- data[].id
- data[].entity_amount.value
- data[].entity_amount.currency
- data[].merchant_amount.value
- data[].merchant_amount.currency
- data[].payee_amount.amount
- data[].payee_amount.currency_code
- data[].state
- data[].user_id
- data[].merchant
- data[].transaction_date
- data[].created_at
- page.next
```

(and the same shape on `ramp_get_reimbursement`).

integrations/ramp/compact.yaml
```diff
@@ -0,0 +1,238 @@
...
+            - phone
+    ramp_list_virtual_cards:
+        spec:
+            - data[].id
+            - data[].display_name
+            - data[].last_four
+            - data[].state
+            - data[].cardholder_id
+            - data[].is_physical
+            - page.next
```
acmacalister, line 146: `VirtualCard` in the live Developer API is much thinner than this spec assumes. OpenAPI only exposes:

`id`, `user_id`, `fund_id`, `is_card_suspended`, `created_at`

So `display_name` / `last_four` / `state` / `cardholder_id` / `is_physical` / `spending_restrictions` all get stripped, and a list call effectively collapses to `{id}` plus `page.next`. Physical cards do have `display_name`/`last_four`/etc. — looks like the virtual specs were copied from that shape.

```yaml
ramp_list_virtual_cards:
  spec:
    - data[].id
    - data[].user_id
    - data[].fund_id
    - data[].is_card_suspended
    - data[].created_at
    - page.next
ramp_get_virtual_card:
  spec:
    - id
    - user_id
    - fund_id
    - is_card_suspended
    - created_at
```

integrations/ramp/compact.yaml
```diff
@@ -0,0 +1,238 @@
...
+    ramp_get_location:
+        spec:
+            - id
+            - name
+    ramp_list_merchants:
+        spec:
+            - data[].id
+            - data[].name
+            - data[].sk_category_name
+            - page.next
```
acmacalister, line 207: Same OpenAPI drift here: merchant list items use `merchant_name`, not `name` (`Merchant` schema is `id`, `merchant_name`, `sk_category_name`, `is_auto_approved`). Compacted merchant lists will keep id/category and drop the display name.

```yaml
- data[].id
- data[].merchant_name
- data[].sk_category_name
- data[].is_auto_approved
- page.next
```

integrations/ramp/handlers.go
```diff
@@ -0,0 +1,336 @@
...
+
+func updateTransactionSplits(ctx context.Context, r *ramp, args map[string]any) (*mcp.ToolResult, error) {
+	rd := mcp.NewArgs(args)
+	id := rd.Str("transaction_id")
+	lineItemsRaw := rd.Str("line_items")
+	if err := rd.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	var lineItems any
+	if err := json.Unmarshal([]byte(lineItemsRaw), &lineItems); err != nil {
```
acmacalister: `line_items` is forced through `rd.Str` + `json.Unmarshal([]byte(...))`. That works when the model stringifies the array, but `execute` arguments (and `api.call` from scripts) are a free-form object — clients often pass a real JSON array.

In that case `ArgStr` errors with `cannot convert []interface {} to string` before we ever hit Ramp. Worth accepting both shapes:

```go
var lineItems any
switch v := args["line_items"].(type) {
case nil:
    return mcp.ErrResult(fmt.Errorf("line_items is required"))
case string:
    if err := json.Unmarshal([]byte(v), &lineItems); err != nil {
        return mcp.ErrResult(fmt.Errorf("invalid JSON for line_items: %w", err))
    }
case []any:
    lineItems = v
default:
    return mcp.ErrResult(fmt.Errorf("line_items must be a JSON array"))
}
```

integrations/ramp/compact.yaml
```diff
@@ -0,0 +1,238 @@
...
+            - data[].cardholder_id
+            - page.next
+    ramp_get_physical_card:
+        spec:
+            - id
+            - display_name
+            - last_four
+            - state
+            - cardholder_id
+            - spending_restrictions
```
acmacalister: Minor: `PhysicalCard` also has no `spending_restrictions` field in OpenAPI (list fields look right otherwise — `display_name`, `last_four`, `state`, `cardholder_id` are real). Dropping that path on get avoids a dead include.

acmacalister (approved):
CI is green (build/test/lint/security/rust-sdk). Nice follow-through on the earlier memo/pagination/config feedback — those look fixed on HEAD.

A few compaction specs still don't match the live Ramp OpenAPI (reimbursements, virtual cards, merchants), plus one `line_items` typing edge case for native JSON arrays. Inline notes; nothing blocking merge if you want to land and clean those up after.

integrations/ramp/compact.yaml
```diff
@@ -0,0 +1,242 @@
...
+            - data[].merchant_name
+            - data[].sk_category_name
+            - data[].is_auto_approved
+            - page.next
+    ramp_list_vendors:
+        spec:
+            - data[].id
+            - data[].name
+            - data[].is_active
+            - data[].remote_id
```
acmacalister: Vendor list items don't expose a top-level `remote_id` in the live Developer API (`/developer/v1/vendors`). The closest fields are `accounting_vendor_remote_id` and `external_vendor_id`, so this include is a no-op and compacted vendor lists lose the ERP/external identifier.

```yaml
- data[].id
- data[].name
- data[].is_active
- data[].accounting_vendor_remote_id
- data[].external_vendor_id
- page.next
```

Same rename is worth applying on `ramp_get_vendor` if we want the remote id there too.

acmacalister (approved):
CI is green (build/test/lint/security/rust-sdk). Earlier memo/pagination/config/compaction feedback looks addressed on HEAD — solid adapter overall. One leftover field-name drift on vendor compaction inline.

integrations/ramp/ramp.go
```diff
@@ -0,0 +1,220 @@
...
+	_ = r.Err()
+	if start == "" && nextURL != "" {
+		if u, err := url.Parse(nextURL); err == nil {
+			start = u.Query().Get("start")
+			if pageSize == "" {
+				pageSize = u.Query().Get("page_size")
+			}
+		}
+	}
+	return map[string]string{
```
acmacalister: `next` is documented as the preferred pagination path, but we only pull `start` and `page_size` out of `page.next` and rebuild the query from the current call args.

Ramp's `page.next` is a full URL whose query already carries the original filters (`user_id`, `from_date`, `state`, etc.). If an agent does the natural thing — pass `next` alone on the follow-up call — those filters get dropped and page 2 is unfiltered.

Worth either following the parsed path+query directly (after allowing the configured host), or merging every query param from `next` into `params` instead of just the two pagination keys.

```go
if nextURL != "" {
    if u, err := url.Parse(nextURL); err == nil {
        // prefer full next query; only overlay explicit caller args if needed
        params := map[string]string{}
        for k, vs := range u.Query() {
            if len(vs) > 0 {
                params[k] = vs[0]
            }
        }
        return params
    }
}
```

integrations/ramp/handlers.go
```diff
@@ -0,0 +1,353 @@
...
+func setTransactionMemo(ctx context.Context, r *ramp, args map[string]any) (*mcp.ToolResult, error) {
+	rd := mcp.NewArgs(args)
+	id := rd.Str("transaction_id")
+	memo := rd.Str("memo")
+	if err := rd.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	body := map[string]any{"memo": memo}
+	if v := rd.Str("is_memo_recurring"); v != "" {
+		body["is_memo_recurring"] = strings.EqualFold(v, "true") || v == "1"
```
acmacalister: `is_memo_recurring` goes through `rd.Str` + string equality, so a native JSON bool from `execute`/`api.call` fails with `cannot convert bool to string` before we hit Ramp.

`mcp.ArgBool` already accepts bool and the usual string forms — same pattern as YNAB:

```go
if _, ok := args["is_memo_recurring"]; ok {
    recurring, err := mcp.ArgBool(args, "is_memo_recurring")
    if err != nil {
        return mcp.ErrResult(err)
    }
    body["is_memo_recurring"] = recurring
}
```

integrations/ramp/tools.go
```diff
@@ -0,0 +1,199 @@
...
+			"memo":              "Memo text (max 255 characters)",
+			"is_memo_recurring": "If true, apply this memo to similar future transactions (optional)",
+		},
+		Required: []string{"transaction_id", "memo"},
+	},
+	{
+		Name: mcp.ToolName("ramp_update_transaction_splits"), Description: "Split or unsplit a Ramp transaction into accounting line items. Empty line_items array unsplits. Use after get_transaction.",
+		Parameters: map[string]string{
+			"transaction_id": "Transaction ID",
+			"line_items":     `JSON array of split line items (e.g. [{"amount":4000,"memo":"Case-1"}]). Empty array removes splits.`,
```
acmacalister: OpenAPI has `line_items[].amount` as an integer in **minor units** (e.g. `4000` = $40.00), while list/get transaction `amount` is a settled number that often looks like major units in examples (`90.0`). Easy for a model to echo the wrong scale when building a split.

Worth calling that out in the param text the way YNAB does for milliunits, e.g. `amount in minor units (4000 = $40.00); must sum to the transaction amount`.

integrations/ramp/compact.yaml
```diff
@@ -0,0 +1,244 @@
...
+#
+# Full spec syntax (paths, arrays, aliases, exclusions): docs/field-compaction.md.
+
+version: 1
+tools:
+    ramp_list_transactions:
+        spec:
+            - data[].id
+            - data[].amount
+            - data[].currency_code
```
acmacalister, line 14: Transaction list compaction keeps `amount` but not `minor_unit_conversion_rate`. Ramp's own schema says to divide `amount` by that factor for the real currency value (USD example: `12034 / 100`), so compacted spend rows can look 100× off depending on how the tenant encodes amounts.

Including `data[].minor_unit_conversion_rate` (and the same on get) would make the money fields interpretable without a second uncompacted fetch.

acmacalister (approved):
CI is green (build/test/lint/security/rust-sdk). Earlier memo/pagination/config/compaction feedback looks addressed on HEAD — solid adapter overall.

A few leftover correctness nits inline around `next` filter preservation, bool arg parsing, and amount units.

integrations/ramp/compact.yaml
```diff
@@ -0,0 +1,246 @@
...
+            - memo
+            - receipts
+    ramp_list_bills:
+        spec:
+            - data[].id
+            - data[].invoice_number
+            - data[].status
+            - data[].status_summary
+            - data[].amount.amount
+            - data[].amount.currency_code
```
acmacalister, line 88: Same minor-units trap we fixed on transactions: bill `amount` is a `CurrencyAmount` (`amount` in smallest denomination + `minor_unit_conversion_rate`). List compaction keeps `amount.amount` / `amount.currency_code` but drops the rate, so compacted bill rows can look 100× off.

```yaml
- data[].amount.amount
- data[].amount.currency_code
- data[].amount.minor_unit_conversion_rate
```

`ramp_get_bill` is fine since it keeps the whole `amount` object.

integrations/ramp/compact.yaml
```diff
@@ -0,0 +1,246 @@
...
+            - page.next
+    ramp_get_reimbursement:
+        spec:
+            - id
+            - entity_amount.value
+            - entity_amount.currency
+            - merchant_amount.value
+            - merchant_amount.currency
+            - payee_amount.amount
+            - payee_amount.currency_code
```
acmacalister, line 72: `payee_amount` is also a `CurrencyAmount`, so without `payee_amount.minor_unit_conversion_rate` the compacted get response has the same scale ambiguity as bill/transaction amounts. `entity_amount` / `merchant_amount` are `ApiAmount` (`value` + `currency` only) and don't need it.

acmacalister (approved):
CI is green (build/test/lint/security/rust-sdk). Earlier memo/pagination/config/compaction feedback looks addressed on HEAD — solid adapter overall.

Two leftover money-field compaction nits inline on bills and reimbursement payee amounts; nothing blocking.

acmacalister (approved):
CI is green (build/test/lint/security/rust-sdk). Clean PR — earlier memo/pagination/compaction feedback looks addressed on HEAD, and I didn't find any new blocking issues against the live Ramp OpenAPI. LGTM.

PR #207, opened 2026-07-28
Add NetSuite ERP REST and SuiteQL integration

daltoniam:
## Summary
- Adds a NetSuite adapter for SuiteQL plus REST record access (customers, vendors, invoices, bills, POs, sales orders, employees, subsidiaries, departments, journal entries)
- Supports OAuth 2.0 bearer tokens or Token-Based Auth (OAuth 1.0 HMAC-SHA256)
- Includes field compaction specs, dispatch parity tests, and config/env wiring

## Test plan
- [x] `make build vet test-race lint` (govulncheck failures are pre-existing on main)
- [ ] Configure `NETSUITE_ACCOUNT_ID` with either OAuth2 or TBA credentials
- [ ] Run `netsuite_suiteql` with a simple customer query and list/get a known record type


acmacalister (approved):
CI passed (build, test, lint, security, rust-sdk all SUCCESS). Clean PR — no blocking issues found. LGTM.

integrations/netsuite/handlers.go
```diff
@@ -0,0 +1,253 @@
...
+	r := mcp.NewArgs(args)
+	selectType := r.Str("select")
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	q := ""
+	if selectType != "" {
+		q = queryEncode(map[string]string{"select": selectType})
+	}
+	data, err := n.get(ctx, "/services/rest/record/v1/metadata-catalog/%s", q)
```
acmacalister: `select` is documented as “record type to describe”, but this builds `metadata-catalog/?select=...`, which only filters the top-level catalog listing. NetSuite’s actual schema/describe endpoint is a path segment (`/metadata-catalog/{type}`) and typically wants `Accept: application/schema+json` (or `application/swagger+json`), not the default `application/json` set in `doRequest`.

As written, callers asking to describe `customer` won’t get field metadata. Something like:

```go
path := "/services/rest/record/v1/metadata-catalog"
headers := map[string]string{}
if selectType != "" {
    path += "/" + url.PathEscape(selectType)
    headers["Accept"] = "application/schema+json"
}
data, err := n.doRequest(ctx, http.MethodGet, path, nil, headers)
```

Worth a httptest assertion covering both full catalog and single-type describe.

config/config_test.go
```diff
@@ -32,7 +32,7 @@ func TestLoad_CreatesDefaultWhenMissing(t *testing.T) {
 	_, err = os.Stat(path)
 	assert.NoError(t, err)
 
-	assert.Len(t, m.cfg.Integrations, 48)
+	assert.Len(t, m.cfg.Integrations, 49)
 	for _, name := range []string{"github", "datadog", "linear", "sentry", "slack", "metabase", "aws", "posthog", "postgres", "clickhouse", "elasticsearch", "pganalyze", "rwx", "gmail", "gcal", "gdrive", "gdocs", "gsheets", "gslides", "gforms", "gchat", "gmeet", "gtasks", "gpeople", "notion", "ollama", "ynab", "stripe", "gcp", "suno", "amazon", "jira", "confluence", "salesforce", "cloudflare", "digitalocean", "fly", "kubernetes", "vercel", "snowflake", "acp", "web", "botidentity", "x", "signoz", "nomad", "agents", "switchboard"} {
```
acmacalister: Len went to 49 for NetSuite, but this name list still has 48 entries and never mentions `"netsuite"`. Same gap in `TestDefaultConfig`’s `expected` map — default credential keys for the new integration aren’t asserted.

Easy miss later if someone drops NetSuite from `defaultConfig` while another integration keeps the count at 49. Mind adding `"netsuite"` here and the credential keys in `expected`?

acmacalister (approved):
Solid NetSuite adapter — clean hexagonal shape, TBA/OAuth2 auth, dispatch/compaction parity, and SuiteQL Prefer: transient all look right. CI is green (build/test/lint/security/rust-sdk). Two non-blocking notes inline: metadata_catalog’s select path doesn’t match NetSuite’s describe API, and config tests never assert the new integration by name.

integrations/netsuite/netsuite.go
```diff
@@ -0,0 +1,380 @@
...
+	if resp.StatusCode == 429 || resp.StatusCode >= 500 {
+		re := &mcp.RetryableError{StatusCode: resp.StatusCode, Err: fmt.Errorf("netsuite API error (%d): %s", resp.StatusCode, string(data))}
+		re.RetryAfter = mcp.ParseRetryAfter(resp.Header.Get("Retry-After"))
+		return nil, re
+	}
+	if resp.StatusCode >= 400 {
+		return nil, fmt.Errorf("netsuite API error (%d): %s", resp.StatusCode, string(data))
+	}
+	if resp.StatusCode == 204 || len(data) == 0 {
+		return json.RawMessage(`{"status":"success"}`), nil
```
acmacalister: NetSuite’s REST create (and often PATCH) returns **204 with no body** and puts the new record URL in the `Location` header — e.g. `.../record/v1/customer/647`. We collapse every empty/204 response to `{"status":"success"}` and never read that header, so `netsuite_create_record` can’t hand the LLM an internal ID for a follow-up get/update.

Worth surfacing `Location` (and ideally the trailing id) when present:

```go
if resp.StatusCode == 204 || len(data) == 0 {
    out := map[string]any{"status": "success"}
    if loc := resp.Header.Get("Location"); loc != "" {
        out["location"] = loc
        if i := strings.LastIndex(loc, "/"); i >= 0 && i+1 < len(loc) {
            out["id"] = loc[i+1:]
        }
    }
    b, _ := json.Marshal(out)
    return json.RawMessage(b), nil
}
```

`TestCreateRecord` currently only checks `IsError == false` on a bare 204 — once this lands, asserting the response includes the Location/id would lock the contract in.

integrations/netsuite/netsuite_test.go
```diff
@@ -0,0 +1,345 @@
...
+}
+
+func TestCreateRecord(t *testing.T) {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, http.MethodPost, r.Method)
+		assert.Equal(t, "/services/rest/record/v1/customer", r.URL.Path)
+		var body map[string]any
+		_ = json.NewDecoder(r.Body).Decode(&body)
+		assert.Equal(t, "Acme", body["companyName"])
+		w.WriteHeader(204)
```
acmacalister, line 310: This create fixture returns 204 with no `Location` header and only asserts `!IsError`. That matches today’s handler, but it won’t catch the real NetSuite contract (204 + `Location: .../customer/<id>`).

Once `doRequest` preserves Location, something like this would lock it:

```go
w.Header().Set("Location", ts.URL+"/services/rest/record/v1/customer/647")
w.WriteHeader(204)
// ...
assert.Contains(t, result.Data, `"id":"647"`)
assert.Contains(t, result.Data, "location")
```

acmacalister (commented):
Solid NetSuite adapter — hexagonal shape, TBA/OAuth2, dispatch/compaction parity, and the earlier metadata-catalog + config coverage fixes all look good. CI is green (build/test/lint/security/rust-sdk). One real gap: create responses drop NetSuite’s Location header, so the LLM never gets the new record id. Notes inline.

integrations/netsuite/netsuite.go
```diff
@@ -0,0 +1,397 @@
...
+	}
+	return "?" + vals.Encode()
+}
+
+func pageParams(args map[string]any) map[string]string {
+	r := mcp.NewArgs(args)
+	limit := r.Str("limit")
+	offset := r.Str("offset")
+	q := r.Str("q")
+	_ = r.Err()
```
acmacalister: `pageParams` is the only `NewArgs` path that discards `.Err()` (`_ = r.Err()`). If a caller ever passes a non-stringable `limit`/`offset`/`q` (map/slice), we silently drop the bad value and hit NetSuite without the filter instead of failing closed.

Worth returning the error so list/SuiteQL handlers can surface it:

```go
func pageParams(args map[string]any) (map[string]string, error) {
	r := mcp.NewArgs(args)
	limit := r.Str("limit")
	offset := r.Str("offset")
	q := r.Str("q")
		return nil, err
	}
	return map[string]string{"limit": limit, "offset": offset, "q": q}, nil
}
```

integrations/netsuite/handlers.go
```diff
@@ -0,0 +1,290 @@
...
+
+func suiteQL(ctx context.Context, n *netsuite, args map[string]any) (*mcp.ToolResult, error) {
+	r := mcp.NewArgs(args)
+	q := r.Str("q")
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	params := pageParams(args)
+	delete(params, "q")
+	path := "/services/rest/query/v1/suiteql" + queryEncode(params)
```
acmacalister, line 34: Tool copy says SuiteQL defaults to `limit=1000`, but when `limit` is omitted we don't send one, so NetSuite's own default (up to 1000 wide rows) kicks in. After compaction the execute path still hard-caps at 50KB (`defaultMaxResponseBytes`), so a bare `SELECT * FROM transaction`-style query will often come back as a generic "Response exceeded 50KB" rather than a usable page.

Other list-heavy adapters default much lower client-side (GitHub `per_page` defaults to 10). Something like defaulting omitted limit to `100` (still well under NetSuite's max) and mentioning that in the tool description would make the entry-point tool succeed more often on first try:

```go
if params["limit"] == "" {
	params["limit"] = "100"
}
```

acmacalister (approved):
CI is green (build/test/lint/security/rust-sdk). Solid NetSuite adapter — hexagonal shape, TBA + OAuth2, dispatch/compaction parity, metadata-catalog describe path, and the Location/id fix on 204 creates all look good. Two non-blocking notes inline on pagination error handling and SuiteQL default page size vs the 50KB response cap.

acmacalister (approved):
CI is green (build/test/lint/security/rust-sdk all SUCCESS). Prior notes on metadata-catalog describe path, Location/id on 204 creates, pageParams error handling, and SuiteQL default limit=100 are all addressed on this tip. Clean NetSuite adapter — LGTM.

integrations/netsuite/handlers.go
```diff
@@ -0,0 +1,290 @@
...
+	r := mcp.NewArgs(args)
+	id := r.Str("id")
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	return getTyped(ctx, n, "invoice", id, args)
+}
+
+func listBills(ctx context.Context, n *netsuite, args map[string]any) (*mcp.ToolResult, error) {
+	return listTyped(ctx, n, "vendorbill", args)
```
acmacalister: The REST record type path segments here are lowercased (`vendorbill`, and same for `purchaseorder` / `salesorder` / `journalentry` below), but NetSuite’s REST API Browser and metadata catalog use camelCase: `vendorBill`, `purchaseOrder`, `salesOrder`, `journalEntry`.

Official metadata examples even filter with `?select=...journalEntry...`, and the browser only lists `/record/v1/vendorBill` (0 hits for `/vendorbill`). As written, the typed list/get tools for bills, POs, SOs, and journal entries will 404 against a real account.

```go
return listTyped(ctx, n, "vendorBill", args)
// purchaseOrder, salesOrder, journalEntry
```

Worth locking the path segment in the httptest fixtures too so this doesn’t regress.

integrations/netsuite/tools.go
```diff
@@ -0,0 +1,185 @@
...
+			"q":      "SuiteQL query string (e.g. SELECT id, companyname FROM customer WHERE isinactive = 'F')",
+			"limit":  "Max rows per page (default 1000, max 1000)",
+			"offset": "Pagination offset (must be divisible by limit)",
+		},
+		Required: []string{"q"},
+	},
+	{
+		Name: mcp.ToolName("netsuite_list_records"), Description: "List NetSuite REST records of any type with optional filter query. Prefer typed list tools when available.",
+		Parameters: map[string]string{
+			"record_type": "REST record type (e.g. customer, vendor, invoice, vendorbill, salesorder, purchaseorder, employee, journalentry)",
```
acmacalister: These examples teach the LLM the wrong casing for multi-word types (`vendorbill`, `salesorder`, `purchaseorder`, `journalentry`). NetSuite REST wants camelCase (`vendorBill`, `salesOrder`, `purchaseOrder`, `journalEntry`) — same issue as the typed handlers.

Updating the examples (and the `netsuite_list_bills` description that says “vendorbill records”) will keep SuiteQL/generic CRUD callers from copying a path that 404s.

acmacalister (commented):
Solid NetSuite adapter overall — hexagonal shape, TBA + OAuth2, SuiteQL Prefer: transient, dispatch/compaction parity, and the metadata-catalog path/`Accept` fix all look good. CI is green on the latest head (build/test/lint/security/rust-sdk).

Agree with the open notes about preserving `Location`/`id` on 204 create/update responses. One more correctness issue inline: multi-word REST record types need camelCase or the typed bill/PO/SO/journal tools will 404.

acmacalister (approved):
CI is green (build, test, lint, security, rust-sdk all SUCCESS). Prior review notes (metadata-catalog describe path + Accept, Location/id on 204 creates, pageParams errors, SuiteQL default limit, camelCase REST types) are addressed on this tip.

Solid NetSuite adapter — hexagonal shape, TBA + OAuth2, dispatch/compaction parity, and SuiteQL Prefer: transient all look right. LGTM.

PR #211, opened 2026-08-04
Decouple app sessions from MCP transport session ids

daltoniam:
## Summary
- Resolve pin/context/history keys from `X-Switchboard-Session-Id` first so Stateless MCP still isolates conversations when `Mcp-Session-Id` is empty (go-sdk 1.7+).
- Wrap HTTP handlers with `AppSessionMiddleware`; keep legacy `Mcp-Session-Id` and `"default"` fallbacks for older clients and one-shot scripts.

## Why
go-sdk 1.7 Stateless mode ignores transport session ids (SEP-2567 / sessionless MCP). Switchboard app sessions previously keyed off `ServerSession.ID()`, so every request collapsed onto `"default"`.

## API for clients
Send on every MCP HTTP request:
```
X-Switchboard-Session-Id: <uuid-per-conversation>
```

Priority: app header → `Mcp-Session-Id` → transport session id → `"default"`.

## Follow-up (hosted)
1. Redis `SessionStore` for multi-pod mcpd
2. Crush mints/sends the header
3. Bump go-sdk to 1.7.x

### DIFF
--- a/server/session_codec.go
+++ b/server/session_codec.go
@@ -1,8 +1,11 @@
 package server
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
+	"strconv"
+	"strings"
 	"time"
 )
 
@@ -21,32 +24,49 @@ type sessionSnapshot struct {
 }
 
 // EncodeSession serializes a Session (including pins) to JSON for durable stores.
+// Field snapshots are copied under the session lock so marshal does not block
+// concurrent PinResult/SetContext on large pin payloads.
 func EncodeSession(s *Session) ([]byte, error) {
 	if s == nil {
 		return nil, fmt.Errorf("encode session: nil")
 	}
 	s.mu.RLock()
-	defer s.mu.RUnlock()
-
 	snap := sessionSnapshot{
-		ID:          s.ID,
-		Context:     s.Context,
-		CreatedAt:   s.CreatedAt,
-		LastUsed:    s.LastUsed,
-		Breadcrumbs: s.Breadcrumbs,
-		NextSeq:     s.nextSeq,
-		NextHandle:  s.nextHandle,
-		PinnedSize:  s.pinnedSize,
+		ID:         s.ID,
+		CreatedAt:  s.CreatedAt,
+		LastUsed:   s.LastUsed,
+		NextSeq:    s.nextSeq,
+		NextHandle: s.nextHandle,
+		PinnedSize: s.pinnedSize,
+	}
+	if len(s.Context) > 0 {
+		snap.Context = make(map[string]any, len(s.Context))
+		for k, v := range s.Context {
+			snap.Context[k] = v
+		}
+	} else {
+		snap.Context = map[string]any{}
+	}
+	if len(s.Breadcrumbs) > 0 {
+		snap.Breadcrumbs = make([]Breadcrumb, len(s.Breadcrumbs))
+		copy(snap.Breadcrumbs, s.Breadcrumbs)
 	}
 	if len(s.pinned) > 0 {
 		snap.Pinned = make(map[string]*PinnedResult, len(s.pinned))
 		for k, v := range s.pinned {
-			snap.Pinned[k] = v
+			if v == nil {
+				continue
+			}
+			// Deep-copy pin payload so marshal is independent of the live map.
+			cp := *v
+			if len(v.Data) > 0 {
+				cp.Data = bytes.Clone(v.Data)
+			}
+			snap.Pinned[k] = &cp
 		}
 	}
-	if snap.Context == nil {
-		snap.Context = map[string]any{}
-	}
+	s.mu.RUnlock()
+
 	return json.Marshal(snap)
 }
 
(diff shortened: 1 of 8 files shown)


server/session_context.go
```diff
@@ -28,3 +106,11 @@ func sessionIDFromReq(ss *mcpsdk.ServerSession) string {
 	}
 	return defaultSessionID
 }
+
+// sessionFor returns the Session bound to this tool call, creating it if needed.
+func (s *Server) sessionFor(ctx context.Context, req *mcpsdk.CallToolRequest) *Session {
+	if sess := sessionFromCtx(ctx); sess != nil {
+		return sess
+	}
+	return s.sessionStore.GetOrCreate(resolveAppSessionID(ctx, req))
```
acmacalister, line 145: Client-supplied session ids flow straight into `GetOrCreate` with no length or charset check. Now that `X-Switchboard-Session-Id` is the primary key (and any unique value creates a store entry), a chatty client can mint unbounded keys and grow the in-memory map until eviction pressure/OOM — each session can also hold up to 5MB of pins.

Worth validating before the store lookup, e.g. max length + a conservative allow-list (UUID hex/dashes, or `[A-Za-z0-9._:-]`), and rejecting/falling back when invalid:

```go
func normalizeAppSessionID(id string) string {
    id = strings.TrimSpace(id)
    if id == "" || len(id) > 128 {
        return ""
    }
    for _, r := range id {
        if !((r >= 'a' && r <= 'z') || (r >= 'A' && r <= 'Z') ||
            (r >= '0' && r <= '9') || r == '-' || r == '_' || r == '.' || r == ':') {
            return ""
        }
    }
    return id
}
```

That also keeps Redis/file backends from getting awkward keys once the shared store lands.

server/session_codec.go
```diff
@@ -0,0 +1,102 @@
...
+	if len(s.pinned) > 0 {
+		snap.Pinned = make(map[string]*PinnedResult, len(s.pinned))
+		for k, v := range s.pinned {
+			snap.Pinned[k] = v
+		}
+	}
+	if snap.Context == nil {
+		snap.Context = map[string]any{}
+	}
+	return json.Marshal(snap)
```
acmacalister, line 70: `json.Marshal` runs while `s.mu` is still held, and `Context` / `Breadcrumbs` / pin payloads are shared references into the live session. Correctness is fine under the RLock, but a session with large pins (up to 5MB) will block `PinResult` / `SetContext` / breadcrumbs for the whole encode.

Safer pattern for the durable-store path: deep-copy (or at least snapshot maps/slices and pin bytes) under the lock, unlock, then marshal:

```go
s.mu.RLock()
snap := sessionSnapshot{ /* copy fields */ }
// copy Context, Breadcrumbs, Pinned entries
s.mu.RUnlock()
return json.Marshal(snap)
```

server/session_codec.go
```diff
@@ -0,0 +1,102 @@
...
+		pinnedSize:  snap.PinnedSize,
+	}
+	if s.Context == nil {
+		s.Context = make(map[string]any)
+	}
+	if s.pinned == nil {
+		s.pinned = make(map[string]*PinnedResult)
+	}
+	// Recompute pinned size if the snapshot omitted it or drifted.
+	if s.pinnedSize == 0 && len(s.pinned) > 0 {
```
acmacalister, line 108: We recompute `pinnedSize` when it's zero, but not `nextHandle`. A payload that has pins (or was written by something that omitted `next_handle`) decodes with `nextHandle == 0`, so the next `PinResult` issues `$1` and can overwrite an existing `$1`.

Same recovery idea as size:

```go
if s.nextHandle == 0 && len(s.pinned) > 0 {
    for h := range s.pinned {
        if strings.HasPrefix(h, "$") {
            if n, err := strconv.Atoi(h[1:]); err == nil && n > s.nextHandle {
                s.nextHandle = n
            }
        }
    }
}
```

server/session_context.go
```diff
@@ -17,9 +35,69 @@ func sessionFromCtx(ctx context.Context) *Session {
...
+			return id
+		}
+		if id := strings.TrimSpace(req.Extra.Header.Get(mcpSessionIDHeader)); id != "" {
+			return id
+		}
+	}
+	if req != nil {
+		return sessionIDFromMCPSession(req.Session)
+	}
+	return defaultSessionID
```
acmacalister, line 124: Falling back to the shared `"default"` key is fine for local stdio / one-shot scripts, but on a multi-tenant HTTP deployment any client that omits `X-Switchboard-Session-Id` (and has no transport session id) will read/write the same pin/context/history bag — cross-conversation leak.

Not a blocker for the OSS local path, but before hosted multi-replica we probably want either: require the app header when auth identity is present, or namespace the fallback as something like `default:<user/org>` instead of a global key. Might be worth a brief note in the hosted follow-up list so it doesn't get lost.

acmacalister (approved):
Nice decoupling of app sessions from transport sessions — priority resolution, middleware wiring, and the isolation tests are solid. CI is green (build/test/lint/security/rust-sdk). A few non-blocking notes inline on client-minted session id validation, codec durability edge cases, and the shared `default` fallback for multi-tenant.

daltoniam:
Addressed all review notes in 8ab9312 (id validation, codec lock/snapshot + nextHandle recovery, default-fallback note for hosted).

acmacalister (approved):
Re-reviewed after the follow-up commit. CI is green (build/test/lint/security/rust-sdk). Session-id validation, codec snapshot-then-marshal, and nextHandle recovery all look solid, and I don't have anything new blocking. LGTM.

PR #212, opened 2026-08-14
Stop advertising MCP tool list changes

daltoniam:
## Summary
- Advertise a stable MCP tool list so Crush 0.89 does not open `subscriptions/listen`.
- Cover the Crush handshake on both the in-memory and stateless HTTP transports.

Crush 0.89 / MCP SDK 1.7 starts a long-lived subscription whenever initialize reports `tools.listChanged`. Switchboard serves request-scoped MCP servers, so that stream closes and `tools/list` fails with it.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/daltoniam/switchboard
 
-go 1.26.5
+go 1.26.6
 
 require (
 	cloud.google.com/go/compute v1.38.0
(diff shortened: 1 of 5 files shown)


acmacalister (approved):
Clean PR — CI checks (build/test/lint/security/rust-sdk) all passed, and I didn't find any blocking issues. Shared staticMCPCapabilities helper matches the MCP Go SDK's recommended way to keep tools advertised while disabling listChanged, covers both main and project servers, and the Crush-style in-memory + StatelessHandler tests lock the regression in. LGTM.

PR #213, opened 2026-08-14
Allow hosted runtimes to register extra MCP prompts and resources

daltoniam:
## Summary

Hosted skill sharing needs a request-scoped way to expose org skills as standard MCP prompts and resources. This adds `server.WithMCPFeatures` so a hosted runtime can register those surfaces without changing Switchboard's search/execute tool model.

## Test plan

- [x] `go test ./server -run 'TestNew|TestWithMCPFeatures'`

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -84,6 +84,7 @@ type Server struct {
 	catalogBytes      int64                 // byte size of full tool catalog (for savings accounting)
 	discoverAll       bool
 	extraInstructions string // appended to the base MCP instructions
+	features          []func(*mcpsdk.Server)
 }
 
 // baseInstructions is the default guidance sent to clients in the MCP
@@ -129,6 +130,18 @@ func WithSessionStore(store SessionStore) Option {
 // The text is appended after the base instructions, separated by a space.
 func WithExtraInstructions(text string) Option {
 	return func(s *Server) { s.extraInstructions = strings.TrimSpace(text) }
+}
+
+// WithMCPFeatures registers additional MCP prompts or resources on the
+// underlying SDK server after the built-in tools are in place. Hosted
+// deployments use this to expose org-scoped skills as prompts/resources
+// without changing Switchboard's search/execute tool surface.
+func WithMCPFeatures(register func(*mcpsdk.Server)) Option {
+	return func(s *Server) {
+		if register != nil {
+			s.features = append(s.features, register)
+		}
+	}
 }
 
 // staticMCPCapabilities advertises a stable tool list. Crush 0.89 / MCP
@@ -177,6 +190,9 @@ func New(services *mcp.Services, opts ...Option) *Server {
 	s.scriptEngine = script.New(&toolExecutor{server: s})
 
 	s.registerTools()
+	for _, register := range s.features {
+		register(s.mcpServer)
+	}
 	return s
 }
 
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -142,6 +142,27 @@ func TestNew(t *testing.T) {
 	assert.NotNil(t, s.services)
 }
 
+func TestWithMCPFeaturesRegistersPrompt(t *testing.T) {
+	reg := newMockRegistry()
+	services := &mcp.Services{
+		Config:   newMockConfigService(map[string]*mcp.IntegrationConfig{}),
+		Registry: reg,
+	}
+	srv := New(services, WithMCPFeatures(func(s *mcpsdk.Server) {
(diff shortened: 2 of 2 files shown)


server/server.go
```diff
@@ -177,6 +190,9 @@ func New(services *mcp.Services, opts ...Option) *Server {
 	s.scriptEngine = script.New(&toolExecutor{server: s})
 
 	s.registerTools()
+	for _, register := range s.features {
+		register(s.mcpServer)
```
acmacalister, line 194: Once a register callback adds prompts/resources here, the go-sdk infers `prompts`/`resources` capabilities with `listChanged: true` — `staticMCPCapabilities` only pins tools:

```go
// go-sdk Server.capabilities()
if s.prompts.len() > 0 {
    if caps.Prompts == nil {
        caps.Prompts = &PromptCapabilities{ListChanged: true}
    }
}
```

That reopens the #212 Crush path this option is aimed at: Crush always installs `PromptListChangedHandler` / `ResourceListChangedHandler`, and MCP SDK 1.7 opens `subscriptions/listen` whenever any of those handlers is set. On a request-scoped/stateless server that stream still has nowhere to live.

Worth extending `staticMCPCapabilities` the same way as tools:

```go
func staticMCPCapabilities() *mcpsdk.ServerCapabilities {
        return &mcpsdk.ServerCapabilities{
                Logging:   &mcpsdk.LoggingCapabilities{},
                Tools:     &mcpsdk.ToolCapabilities{ListChanged: false},
                Prompts:   &mcpsdk.PromptCapabilities{ListChanged: false},
                Resources: &mcpsdk.ResourceCapabilities{ListChanged: false},
        }
}
```

(SDK still only advertises prompts/resources when something is actually registered; a non-nil field with `listChanged:false` just overrides the inferred default.)

server/server_test.go
```diff
@@ -140,6 +140,27 @@ func TestNew(t *testing.T) {
...
+				Description: "review",
+				Messages: []*mcpsdk.PromptMessage{{
+					Role:    "user",
+					Content: &mcpsdk.TextContent{Text: "review this"},
+				}},
+			}, nil
+		})
+	}))
+	require.NotNil(t, srv)
+	require.NotNil(t, srv.mcpServer)
```
acmacalister, line 163: This only asserts the server constructed successfully — `AddPrompt` never has to succeed for the test to pass, and we don't check that clients can actually see or fetch the prompt.

`TestWithExtraInstructions` / `TestStaticMCPCapabilities_DisableListChanged` already wire an in-memory client session; same shape here would catch both registration and the listChanged pitfall:

```go
clientTransport, serverTransport := mcpsdk.NewInMemoryTransports()
ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
defer cancel()

ss, err := srv.mcpServer.Connect(ctx, serverTransport, nil)
require.NoError(t, err)
defer ss.Close() //nolint:errcheck

client := mcpsdk.NewClient(&mcpsdk.Implementation{Name: "test", Version: "1.0"}, nil)
cs, err := client.Connect(ctx, clientTransport, nil)
require.NoError(t, err)
defer cs.Close() //nolint:errcheck

caps := cs.InitializeResult().Capabilities
require.NotNil(t, caps.Prompts)
assert.False(t, caps.Prompts.ListChanged)

prompts, err := cs.ListPrompts(ctx, nil)
require.NoError(t, err)
require.Len(t, prompts.Prompts, 1)
assert.Equal(t, "skill_review", prompts.Prompts[0].Name)

got, err := cs.GetPrompt(ctx, &mcpsdk.GetPromptParams{Name: "skill_review"})
require.NoError(t, err)
require.NotEmpty(t, got.Messages)
```

acmacalister (commented):
Nice small extension point for hosted skill sharing — option shape matches the existing `With*` pattern. Two things worth tightening before this lands on the Crush/stateless path: prompt/resource `listChanged` defaults (same class of bug as #212), and the registration test currently doesn't exercise the MCP surface at all.

CI: build, test, lint, security, rust-sdk all green.

acmacalister (commented):
Option shape looks good and matches the existing `With*` pattern. CI is green across build, test, lint, security, rust-sdk, and compose.

The two open threads from the earlier pass still block landing this on the Crush/stateless path:

1. `staticMCPCapabilities` still only pins `tools.listChanged=false`. Once a feature callback registers prompts/resources, go-sdk v1.5.0 infers those caps with `listChanged: true` (see `Server.capabilities()`), which reopens the #212 subscriptions/listen failure mode.
2. `TestWithMCPFeaturesRegistersPrompt` still only checks construction — it never asserts the prompt is listable/fetchable or that `prompts.listChanged` stays false. The in-memory client shape from `TestStaticMCPCapabilities_DisableListChanged` would cover both.

No new issues beyond those. Happy to approve once the threads are addressed.

PR #220, opened 2026-08-30
fix: stop skill prompts from hanging Crush on initialize

daltoniam:
## Summary
- Pin prompt and resource `listChanged` false alongside tools so Crush 0.89 does not open `subscriptions/listen` against request-scoped MCP servers.
- Hosted skills register prompts/resources via `WithMCPFeatures`; a nil capability field previously let the SDK advertise `listChanged=true` and left Crush stuck on starting.

## Test plan
- [x] `go test ./server/ -run 'TestStaticMCPCapabilities|TestWithMCPFeatures|TestStatelessHandler_Crush'`
- [ ] Hosted mcpd with a published skill initializes in Crush without hanging


acmacalister (approved):
## PR #220 Review: fix: stop skill prompts from hanging Crush on initialize

### Verification (existing CI — not re-run)
| Check | Result |
|-------|--------|
| build | Pass |
| test | Pass |
| lint | Pass |
| security | Pass |
| rust-sdk | Pass |
| compose | Pass |

### What's Good
- Correct root cause vs go-sdk v1.7.0: non-nil `Prompts`/`Resources` override inferred `listChanged=true` when skills call `AddPrompt`/`AddResource`.
- `WithMCPFeatures` is a clean extension point; registration after `registerTools` and before catalog attach is the right order.
- Regression coverage with in-memory + Crush-style list-changed handlers matches the hang path.

### Must Fix
No issues found.

### Should Fix
No issues found.

### Consider
- `TestStaticMCPCapabilities_DisableListChanged` could `require.NotNil` on prompts/resources now that static caps always set them (optional `if` is weaker than the skills test).
- Pre-existing: with Project Catalog enabled, `Resources.ListChanged=true` can still open `subscriptions/listen` for clients with a resource list-changed handler — outside this PR's skill-prompt fix.

LGTM — clean, targeted fix with the right SDK-level invariant.

PR #221, opened 2026-09-04
Add Paperless and Recoll document search integrations

aleksclark:
## Summary
- Add Paperless-ngx document search, metadata, download, update, deletion, and safe test-upload tools.
- Add Recoll WebUI search backed by its live HTML query endpoint.
- Document isolated development smoke testing and integration configuration.

## Verification
- `make ci`
- Isolated live smoke tests against Paperless-ngx and Recoll WebUI

💘 Generated with Crush

### DIFF
--- a/integrations/paperless/paperless.go
+++ b/integrations/paperless/paperless.go
@@ -1,0 +1,312 @@
+// Package paperless integrates Paperless-ngx document management instances.
+package paperless
+
+import (
+	"bytes"
+	"context"
+	_ "embed"
+	"encoding/json"
+	"fmt"
+	"io"
+	"mime/multipart"
+	"net/http"
+	"net/url"
+	"strconv"
+	"strings"
+
+	mcp "github.com/daltoniam/switchboard"
+	"github.com/daltoniam/switchboard/compact"
+)
+
+//go:embed compact.yaml
+var compactYAML []byte
+
+var compactResult = compact.MustLoadWithOverlay("paperless", compactYAML, compact.Options{Strict: false})
+var fieldCompactionSpecs = compactResult.Specs
+
+var (
+	_ mcp.Integration                = (*paperless)(nil)
+	_ mcp.FieldCompactionIntegration = (*paperless)(nil)
+	_ mcp.PlainTextCredentials       = (*paperless)(nil)
+	_ mcp.PlaceholderHints           = (*paperless)(nil)
+)
+
+type paperless struct {
+	token   string
+	baseURL string
+	client  *http.Client
+}
+
+// New creates a Paperless-ngx integration.
+func New() mcp.Integration {
+	return &paperless{client: &http.Client{}}
+}
+
+func (p *paperless) Name() string { return "paperless" }
+
+func (p *paperless) PlainTextKeys() []string { return []string{"url"} }
+
+func (p *paperless) Placeholders() map[string]string {
+	return map[string]string{"url": "https://paperless.example.com"}
+}
+
+func (p *paperless) Configure(_ context.Context, creds mcp.Credentials) error {
+	p.token = creds["token"]
+	p.baseURL = strings.TrimRight(creds["url"], "/")
+	if p.token == "" {
+		return fmt.Errorf("paperless: token is required")
+	}
+	if p.baseURL == "" {
+		return fmt.Errorf("paperless: url is required")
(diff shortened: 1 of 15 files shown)


integrations/paperless/paperless.go
```diff
@@ -0,0 +1,312 @@
...
+	req.Header.Set("Authorization", "Token "+p.token)
+	if body != nil {
+		req.Header.Set("Content-Type", "application/json")
+	}
+	resp, err := p.client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("make Paperless request: %w", err)
+	}
+	defer func() { _ = resp.Body.Close() }()
+	data, err := io.ReadAll(resp.Body)
```
acmacalister, line 110: These response bodies are read with unbounded `io.ReadAll`. That is especially painful on `paperless_download_document`, where a multi‑MB PDF can land entirely in memory, and it is out of step with newer adapters that cap reads (`io.LimitReader` + a max response size).

Worth applying the same limit in both `doRequest` and `doMultipartRequest` (and preferably a tighter cap on the download path).

integrations/paperless/paperless.go
```diff
@@ -0,0 +1,312 @@
...
+func downloadDocument(ctx context.Context, p *paperless, args map[string]any) (*mcp.ToolResult, error) {
+	id, err := documentID(args)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	data, err := p.get(ctx, fmt.Sprintf("/api/documents/%d/download/", id))
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	return &mcp.ToolResult{Data: string(data)}, nil
```
acmacalister, line 287: `string(data)` will mangle binary originals (PDFs, images) and dump raw bytes straight into the tool result. Drive solved this with a small envelope: size cap, `content_type`, inline text for `text/*`, and `content_base64` otherwise (`gdrive` `downloadEnvelope`).

Something in that shape would make downloads actually usable for agents instead of producing corrupted UTF‑8 noise.

integrations/paperless/paperless.go
```diff
@@ -0,0 +1,312 @@
...
+
+type paperless struct {
+	token   string
+	baseURL string
+	client  *http.Client
+}
+
+// New creates a Paperless-ngx integration.
+func New() mcp.Integration {
+	return &paperless{client: &http.Client{}}
```
acmacalister, line 42: `&http.Client{}` has no timeout, so a hung Paperless instance can block a tool call indefinitely. Recoll in this same PR sets `Timeout: 30 * time.Second` — matching that here would be consistent and safer.

integrations/recoll/tools.go
```diff
@@ -0,0 +1,145 @@
...
+	},
+}
+
+var dispatch = map[mcp.ToolName]handlerFunc{
+	"recoll_search": search,
+}
+
+func search(ctx context.Context, r *recoll, args map[string]any) (*mcp.ToolResult, error) {
+	reader := mcp.NewArgs(args)
+	query := reader.Str("query")
```
acmacalister, line 34: `Str("query")` accepts an empty string, so a missing/blank query still hits the WebUI. Paperless search rejects empty query explicitly; same guard here would avoid a pointless HTML round-trip and a confusing empty result set.

integrations/recoll/tools.go
```diff
@@ -0,0 +1,145 @@
...
+			return
+		}
+		if hasClass(n, "result") {
+			response.Results = append(response.Results, searchResult{
+				Path:     textForClass(n, "path"),
+				MIMEType: decodeMIMETypes(textForClass(n, "meta")),
+			})
+			return
+		}
+		if n.Data == "p" && response.Count == 0 {
```
acmacalister, line 82: If the count paragraph markup changes, `Count` stays 0 even when `Results` is non-empty. Falling back to `len(response.Results)` (or setting count from results when the regex misses) would keep the payload coherent across WebUI tweaks.

acmacalister (commented):
Nice pair of integrations — wiring, dispatch parity, compaction specs, and config/env defaults all look solid, and CI is green across build/test/lint/security. A few follow-ups on Paperless downloads and a couple of Recoll edge cases below (PR is already merged, so these are post-merge notes).

PR #222, opened 2026-09-05
Expand Paperless administration and document retrieval

aleksclark:
## Summary
- Add Paperless administration, workflow, task, and bulk-operation tools.
- Keep OCR text in a paginated drill-down tool and bound all upstream responses.
- Address Paperless and Recoll post-merge review feedback.

## Verification
- `make ci`
- Isolated Crush smoke test against Paperless-ngx, including paginated OCR retrieval

💘 Generated with Crush

### DIFF
--- a/integrations/paperless/configuration.go
+++ b/integrations/paperless/configuration.go
@@ -1,0 +1,178 @@
+package paperless
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/url"
+	"strconv"
+
+	mcp "github.com/daltoniam/switchboard"
+)
+
+// listResource returns a paginated Paperless collection. Resource-specific
+// filtering is deliberately passed through only as documented query values.
+func listResource(path string) handlerFunc {
+	return func(ctx context.Context, p *paperless, args map[string]any) (*mcp.ToolResult, error) {
+		r := mcp.NewArgs(args)
+		page := r.OptInt("page", 1)
+		pageSize := r.OptInt("page_size", 25)
+		if err := r.Err(); err != nil {
+			return mcp.ErrResult(err)
+		}
+		query := url.Values{"page": {strconv.Itoa(page)}, "page_size": {strconv.Itoa(pageSize)}}
+		data, err := p.get(ctx, path+"?"+query.Encode())
+		if err != nil {
+			return mcp.ErrResult(err)
+		}
+		return mcp.RawResult(data)
+	}
+}
+
+func getResource(path string) handlerFunc {
+	return func(ctx context.Context, p *paperless, args map[string]any) (*mcp.ToolResult, error) {
+		id, err := resourceID(args)
+		if err != nil {
+			return mcp.ErrResult(err)
+		}
+		data, err := p.get(ctx, fmt.Sprintf("%s%d/", path, id))
+		if err != nil {
+			return mcp.ErrResult(err)
+		}
+		return mcp.RawResult(data)
+	}
+}
+
+func createResource(path string) handlerFunc {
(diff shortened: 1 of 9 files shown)


integrations/paperless/tools.go
```diff
@@ -1,69 +1,110 @@
...
+		{Name: "paperless_delete_document", Description: "Permanently delete a Paperless-ngx document. Use after get_document when the document should be removed.", Parameters: map[string]string{"document_id": "Document ID"}, Required: []string{"document_id"}},
+		{Name: "paperless_download_document", Description: "Download the original file for a Paperless-ngx document as a JSON envelope with content type, byte count, and base64 content. Use after get_document when the document file is needed.", Parameters: map[string]string{"document_id": "Document ID"}, Required: []string{"document_id"}},
+
+		{Name: "paperless_list_tasks", Description: "List asynchronous Paperless-ngx background tasks, including document consumption and workflow jobs. Start here to monitor task execution.", Parameters: pageParameters},
+		{Name: "paperless_get_task", Description: "Get a specific Paperless-ngx asynchronous task by ID. Use after list_tasks or an upload returns a task ID.", Parameters: idParameter, Required: []string{"id"}},
```
acmacalister, line 49: Upload returns a Celery `task_id` UUID (`{"task_id":"..."}`), but `paperless_get_task` only accepts integer DB `id` and hits `/api/tasks/{id}/`. Following this description after `create_document` will 404.

Paperless list already supports `?task_id=<celery-uuid>`. Worth either accepting that UUID here (list/filter then return the match) or tightening the description and adding a dedicated lookup so the create → poll path works.

integrations/paperless/paperless.go
```diff
@@ -201,45 +240,116 @@ func createDocument(ctx context.Context, p *paperless, args map[string]any) (*mc
...
 	id, err := documentID(args)
 	if err != nil {
 		return mcp.ErrResult(err)
 	}
+	offset, limit, err := ocrTextPageArgs(args)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+
 	data, err := p.get(ctx, fmt.Sprintf("/api/documents/%d/", id))
```
acmacalister, line 269: Pagination only slices after we pull the full document JSON, including OCR `content`, under the shared 2MB response cap. A long OCR'd PDF can fail with "response exceeds 2MB bytes" before any page is returned, which undercuts the drill-down this tool is meant to enable.

If upstream has no partial-content endpoint, we may need a higher limit just for this path, or stream/decode only the `content` field so large docs still page instead of hard-failing.

integrations/paperless/configuration.go
```diff
@@ -0,0 +1,178 @@
...
+	mcp "github.com/daltoniam/switchboard"
+)
+
+// listResource returns a paginated Paperless collection. Resource-specific
+// filtering is deliberately passed through only as documented query values.
+func listResource(path string) handlerFunc {
+	return func(ctx context.Context, p *paperless, args map[string]any) (*mcp.ToolResult, error) {
+		r := mcp.NewArgs(args)
+		page := r.OptInt("page", 1)
+		pageSize := r.OptInt("page_size", 25)
```
acmacalister, line 19: `page_size` is passed straight through and Paperless allows very large values (up to 100000). Combined with the 2MB response ceiling, a high `page_size` on document-heavy collections will error out rather than truncate.

Capping this (e.g. 100) in the handler and documenting the max would keep list/search tools reliable for agents.

acmacalister (commented):
Solid expansion of Paperless admin surface area — the generic resource factory, OCR drill-down, and danger copy on mutations are in good shape, and CI is green across build/test/lint/security/compose/rust-sdk. A few correctness and reliability gaps on the task-poll and large-document paths worth fixing when you next touch this.

PR #224, opened 2026-09-06
Add vision-ready Paperless document retrieval

aleksclark:
## Summary
- expose Paperless thumbnails and previews as native MCP media so vision-capable models can inspect document orientation and read static images without decoding base64
- preserve PDFs as embedded resources and add isolated live coverage for thumbnail, preview, and OCR retrieval

## Test plan
- `PAPERLESS_LIVE_TEST=1 go test ./integrations/paperless -run '^TestLiveDocumentVision$' -v` against `paperless.fleet.clark.team`
- `make ci`

💘 Generated with Crush


server/server.go
```diff
@@ -896,22 +897,26 @@ func (s *Server) handleExecute(ctx context.Context, req *mcpsdk.CallToolRequest)
...
-	text := result.Data
-	var out *mcpsdk.CallToolResult
-	if handle != "" {
-		out = &mcpsdk.CallToolResult{
-			Content: []mcpsdk.Content{
-				&mcpsdk.TextContent{Text: text},
-				&mcpsdk.TextContent{Text: "pinned as " + handle},
-			},
+	content := []mcpsdk.Content{&mcpsdk.TextContent{Text: result.Data}}
+	for _, media := range result.Media {
```
acmacalister, line 923: This correctly maps `result.Media` into native MCP image/resource content on the main execute path, but `ProjectRouter.makeExecuteHandler` in `server/project_server.go` still only returns `TextContent{Text: result.Data}`. On `/mcp/{project}`, thumbnail/preview would succeed with metadata JSON while silently dropping the image/PDF bytes.

Worth extracting this loop into a shared helper used by both surfaces, plus a project-router test mirroring `TestHandleExecute_ReturnsNativeImageContent`.

server/server.go
```diff
@@ -896,22 +897,26 @@ func (s *Server) handleExecute(ctx context.Context, req *mcpsdk.CallToolRequest)
...
 	}
-	text := result.Data
-	var out *mcpsdk.CallToolResult
-	if handle != "" {
-		out = &mcpsdk.CallToolResult{
-			Content: []mcpsdk.Content{
-				&mcpsdk.TextContent{Text: text},
-				&mcpsdk.TextContent{Text: "pinned as " + handle},
-			},
+	content := []mcpsdk.Content{&mcpsdk.TextContent{Text: result.Data}}
```
acmacalister, line 922: The size check just above still only looks at `len(result.Data)`, and for these tools `Data` is just the tiny metadata JSON. The real payload is in `result.Media` (adapter-capped at 5MB), so multi-megabyte images/PDFs bypass the 50KB safety net and then expand further when the SDK base64-encodes them on the wire.

Either fold media bytes into the cap (or a dedicated media budget), or call out that media tools are intentionally exempt and rely only on the adapter limit. Right now the server check looks protective when it isn't.

integrations/paperless/paperless.go
```diff
@@ -256,6 +258,49 @@ func getDocument(ctx context.Context, p *paperless, args map[string]any) (*mcp.T
...
+		"document_id":  id,
+		"content_type": contentType,
+		"bytes":        len(data),
+	})
+	if err != nil {
+		return mcp.ErrResult(fmt.Errorf("encode Paperless image metadata: %w", err))
+	}
+	nameExtension := ".pdf"
+	if contentType != "application/pdf" {
+		extensions, _ := mime.ExtensionsByType(contentType)
```
acmacalister: `mime.ExtensionsByType("image/jpeg")` returns `.jfif` first on Go's registry, so JPEG previews can end up named `paperless-document-N.jfif` instead of `.jpg`/`.jpeg`. Harmless for vision models, but a little surprising if anything keys off the filename. A small allow-list (`.webp`/`.png`/`.jpg`/`.gif`/`.pdf`) would keep names predictable.

acmacalister (approved):
Nice addition — native ImageContent/EmbeddedResource for Paperless thumbnails and previews is the right shape, and the unit + live coverage around it looks solid. CI is green (build/test/lint/security/rust-sdk/compose). A couple of follow-ups inline: project-scoped execute still drops Media, and the response size check only sees the metadata JSON.

server/server.go
aleksclark: Addressed in 1d76269: both execute surfaces now share native media conversion, with project-router coverage.

server/server.go
aleksclark: Addressed in 1d76269: response limits now include metadata and binary media bytes, with an oversized-media regression test.

integrations/paperless/paperless.go
aleksclark: Addressed in 1d76269: common media MIME types now map to stable extensions, including image/jpeg to .jpg.

integrations/paperless/paperless.go
```diff
@@ -256,6 +258,48 @@ func getDocument(ctx context.Context, p *paperless, args map[string]any) (*mcp.T
...
+		return mcp.ErrResult(fmt.Errorf("encode Paperless image metadata: %w", err))
+	}
+	nameExtension := map[string]string{
+		"application/pdf": ".pdf",
+		"image/gif":       ".gif",
+		"image/jpeg":      ".jpg",
+		"image/png":       ".png",
+		"image/webp":      ".webp",
+	}[contentType]
+	return mcp.MediaResult(string(metadata), data, contentType, fmt.Sprintf("paperless-document-%d%s", id, nameExtension))
```
acmacalister, line 309: After 1d76269, `toolResultBytes` counts media into the server response cap, but Paperless still uses the default 50KB limit — there's no `PerToolMaxResponseBytesIntegration` (or integration-wide raise) for these tools. Adapter-side we allow up to 5MB via `paperlessImageSizeLimit`, so any real thumbnail/preview over ~50KB will fetch successfully and then get rejected with "Response exceeded" on both execute surfaces.

The unit/live tests call `integration.Execute` directly and never hit that server check, so they stay green while MCP execute is broken for typical documents. Same pattern as GitHub's pull-diff headroom:

```go
var (
	_ mcp.Integration                        = (*paperless)(nil)
	_ mcp.PerToolMaxResponseBytesIntegration = (*paperless)(nil)
	// ...
)

func (p *paperless) MaxResponseBytesForTool(toolName mcp.ToolName) (int, bool) {
	switch toolName {
	case "paperless_get_document_thumbnail", "paperless_get_document_preview":
		return paperlessImageSizeLimit, true
	}
	return 0, false
}
```

Worth a server-level regression that runs thumbnail/preview through `handleExecute` with a payload between 50KB and the raised cap so this doesn't regress again.

acmacalister (commented):
CI is green (build/test/lint/security/rust-sdk/compose) and the follow-ups from the earlier pass look solid — shared `toolResultContent`/`toolResultBytes`, project-router parity, JPEG extension map. One real functional gap left: media is now counted toward the 50KB server cap, but Paperless never raises that cap for thumbnail/preview, so typical documents will fail on the MCP path even though adapter + unit tests pass.

integrations/paperless/paperless.go
aleksclark: Addressed in 715521a: thumbnail and preview now declare 5MB per-tool response headroom, with parity coverage for both tools and the default-cap fallback.

acmacalister (approved):
Clean PR — CI is green across build/test/lint/security/rust-sdk/compose, and the earlier follow-ups (shared media conversion on both execute surfaces, media bytes in the response cap, stable JPEG extensions, and 5MB per-tool headroom for thumbnail/preview) all look correctly wired. LGTM.

PR #225, opened 2026-09-06
Add Zendesk Support ticketing integration

daltoniam:
## Summary
- Add a Zendesk Support adapter so agents can search, create, update, and comment on tickets without leaving Switchboard.
- Cover the adjacent support workflow too: users, organizations, groups, views, macros, Help Center articles, tags, and CSAT ratings.

## Test plan
- [x] Adapter unit tests (`go test ./integrations/zendesk/`)
- [x] Config defaults and env mapping tests
- [x] `make ci` (build, vet, race tests, lint, gosec); govulncheck still reports existing Go 1.26.5 stdlib findings
- [ ] Enable Zendesk in the web UI with subdomain + API token (or OAuth token) and run `search` / `execute` for `zendesk_search_tickets`


integrations/zendesk/handlers.go
```diff
@@ -0,0 +1,557 @@
...
+	query := r.Str("query")
+	sortBy := r.Str("sort_by")
+	sortOrder := r.Str("sort_order")
+	page := r.OptInt("page", 1)
+	perPage := r.OptInt("per_page", 25)
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	if !strings.Contains(query, "type:") {
+		query = "type:ticket " + query
```
acmacalister, line 25: `strings.Contains(query, "type:")` will false-positive on free-text that just happens to include that substring. Something like `prototype: broken login` skips the `type:ticket` prefix and searches across users/orgs too, which is the opposite of what this tool promises.

A token-aware check is safer:

```go
hasType := false
for _, tok := range strings.Fields(query) {
	if strings.HasPrefix(tok, "type:") {
		hasType = true
		break
	}
}
if !hasType {
	query = "type:ticket " + query
}
```

Worth a regression case in `TestSearchTickets` for a query containing `prototype:`.

integrations/zendesk/handlers.go
```diff
@@ -0,0 +1,557 @@
...
+	if groupID != "" {
+		ticket["group_id"] = groupID
+	}
+	if orgID != "" {
+		ticket["organization_id"] = orgID
+	}
+	if tags != "" {
+		ticket["tags"] = splitCSV(tags)
+	}
+	if customFieldsRaw != "" {
```
acmacalister: `custom_fields` only accepts a JSON string. When an LLM (or script) passes a native array — which is the natural shape for `[{id, value}, ...]` — `r.Str` fails with a type error and the whole create/update bails.

Other adapters that take structured blobs usually accept both string and already-parsed values. Something like:

```go
if raw, ok := args["custom_fields"]; ok && raw != nil {
	switch v := raw.(type) {
	case string:
		if v != "" {
			if err := json.Unmarshal([]byte(v), &fields); err != nil {
				return nil, fmt.Errorf("invalid custom_fields JSON: %w", err)
			}
		}
	case []any:
		fields = v
	default:
		return nil, fmt.Errorf("custom_fields must be a JSON array or string")
	}
	ticket["custom_fields"] = fields
}
```

integrations/zendesk/handlers.go
```diff
@@ -0,0 +1,557 @@
...
+		ticket["status"] = status
+	}
+	if priority != "" {
+		ticket["priority"] = priority
+	}
+	if typ != "" {
+		ticket["type"] = typ
+	}
+	if requesterID != "" {
+		ticket["requester_id"] = requesterID
```
acmacalister: `requester_id` / `assignee_id` / `group_id` / `organization_id` are typed as integers in the Zendesk Ticket API, but we always ship them as JSON strings via `r.Str`. Most tenants coerce, but some requests 422 on the type mismatch — especially once IDs come back as numbers from a prior tool call and we re-stringify them.

Prefer parsing to int64 when present so the outbound body matches the schema:

```go
if requesterID != "" {
	id, err := strconv.ParseInt(requesterID, 10, 64)
	if err != nil {
		return nil, fmt.Errorf("requester_id: %w", err)
	}
	ticket["requester_id"] = id
}
```

Same idea for assignee/group/org (and the empty-assignee unassign path can stay `nil`).

acmacalister (approved):
Solid new adapter — wiring, dispatch/compaction/markdown parity, and auth dual-path all look clean, and CI is green (build/test/lint/security/compose/rust-sdk). Three things worth tightening before this hits real tenants, mainly around search query shaping and create/update payloads.

integrations/zendesk/handlers.go
```diff
@@ -0,0 +1,627 @@
...
+	return mcp.RawResult(data)
+}
+
+func addTicketComment(ctx context.Context, z *zendesk, args map[string]any) (*mcp.ToolResult, error) {
+	r := mcp.NewArgs(args)
+	id := r.Str("ticket_id")
+	body := r.Str("body")
+	public := true
+	if v := r.Str("public"); v != "" {
+		public = v != "false"
```
acmacalister: `public` is parsed as a string and only treated as private when it equals exactly `"false"`. A native bool works because `ArgStr` lowercases it, but values like `"False"`, `"FALSE"`, `"0"`, or `"f"` all stay public — the opposite of what the agent intended for an internal note.

`r.Bool` / `mcp.ArgBool` already accept those forms. Presence-check + Bool keeps the default-true behavior:

```go
public := true
if _, ok := args["public"]; ok {
	public = r.Bool("public")
}
```

Same pattern would help `ticketPayload`'s `public` / `verified` fields below.

integrations/zendesk/handlers.go
```diff
@@ -0,0 +1,627 @@
...
+	} else if ok {
+		ticket["group_id"] = id
+	}
+	if id, ok, err := optionalIntID(args, "organization_id"); err != nil {
+		return nil, err
+	} else if ok {
+		ticket["organization_id"] = id
+	}
+	if tags != "" {
+		ticket["tags"] = splitCSV(tags)
```
acmacalister: `tags` only comes through `r.Str` + `splitCSV`, so a native array (the shape tickets already return) fails the type conversion and aborts create/update.

`r.StrSlice` accepts CSV strings and `[]string`/`[]any` the same way Datadog/GitHub do for labels/tags:

```go
if tags := r.StrSlice("tags"); len(tags) > 0 {
	ticket["tags"] = tags
}
```

Worth doing in `userPayload` too so update-user doesn't hit the same wall.

integrations/zendesk/handlers.go
daltoniam: Fixed in 92f1389: `hasTypeFilter` now tokenizes on whitespace and only matches `type:` prefixes, so `prototype: broken login` still gets `type:ticket`.

acmacalister (approved):
Nice adapter — dispatch/compaction/markdown parity, dual auth, and the follow-up hardening for search + ticket payloads all look solid. CI is green across build/test/lint/security/compose/rust-sdk. Two small LLM-input edges on `public` and `tags` inline; neither is blocking.

integrations/zendesk/handlers.go
daltoniam: Fixed in 92f1389: `parseCustomFields` accepts a JSON string or a native array/object so LLM/script payloads no longer 400 on type conversion.

integrations/zendesk/handlers.go
daltoniam: Fixed in 92f1389: `optionalIntID` now emits integer IDs for requester/assignee/group/organization so the outbound ticket body matches the Zendesk schema.

integrations/zendesk/handlers.go
daltoniam: Fixed in 4884e90: `public` and `verified` now use `r.Bool` with a presence check, so `False`/`FALSE`/`0`/`f` become private notes as intended.

integrations/zendesk/handlers.go
daltoniam: Fixed in 4884e90: tickets and users now take `r.StrSlice("tags")`, so native arrays and CSV both work.

integrations/zendesk/handlers.go
```diff
@@ -0,0 +1,617 @@
...
+	} else if ok {
+		ticket["group_id"] = id
+	}
+	if id, ok, err := optionalIntID(args, "organization_id"); err != nil {
+		return nil, err
+	} else if ok {
+		ticket["organization_id"] = id
+	}
+	if len(tags) > 0 {
+		ticket["tags"] = tags
```
acmacalister, line 505: `update_ticket` advertises tags as replacing the existing set, but we only send `tags` when `len(tags) > 0`. An explicit clear (`tags: []` or `tags: ""`) gets dropped, so the previous tags stick around — same shape as `assignee_id` empty-to-unassign, just without the presence check.

```go
if _, present := args["tags"]; present {
	if tags == nil {
		tags = []string{}
	}
	ticket["tags"] = tags
}
```

Worth the same treatment in `userPayload` a few lines down.

acmacalister (approved):
Solid Zendesk adapter — dual auth, dispatch/compaction/markdown parity, and the earlier search/payload hardening all look good. CI is green across build/test/lint/security/compose/rust-sdk. One non-blocking edge on clearing tags inline.

integrations/zendesk/handlers.go
daltoniam: Fixed in c41c0a0: ticket and user payloads now send `tags` whenever the argument is present, including empty lists, so an explicit clear replaces the existing set.

acmacalister (approved):
## PR #225 Review: Add Zendesk Support ticketing integration

### Verification (existing CI)
| Check | Result |
|-------|--------|
| build | Pass |
| test | Pass |
| lint | Pass |
| security | Pass |
| rust-sdk | Pass |
| compose | Pass |

### What'''s Good
- Clean hexagonal adapter: unexported struct, exported `New()`, dual auth (OAuth + API token Basic), dispatch map with full parity tests
- Compaction + markdown coverage for document-shaped tools, shared `mcp` helpers throughout
- Earlier review rounds landed solid hardening (token-aware `type:` filter, int IDs, native tags/custom_fields, empty-tag clears) with regression tests

### Must Fix (Blocking)
No issues found

### Should Fix (Non-Blocking)
No issues found

### Consider (Nice to Have)
- Tool param copy still says tags are comma-separated only; handlers also accept native arrays — a one-line description tweak would match behavior
- `subdomain` is required in the UI credential schema even when `base_url` alone is enough for Configure (test/proxy setups)

Clean PR — CI is green across build/test/lint/security/compose/rust-sdk and I didn'''t find any blocking issues. LGTM.

PR #226, opened 2026-09-06
Add Microsoft 365 Graph integration

daltoniam:
## Summary

- Add a Microsoft Graph adapter so agents can work with Outlook mail, calendar, OneDrive/SharePoint, Teams, To Do, and directory users through Switchboard.
- Include Entra ID OAuth (PKCE) setup in the web UI, token refresh, field compaction, and markdown rendering for Outlook messages.

## Test plan

- [ ] `make ci` (or at least `go test ./integrations/microsoft365/ ./config/ ./web/`)
- [ ] In the web UI, open Microsoft 365 setup, save an Entra app, and complete OAuth
- [ ] `search` for `outlook mail` / `onedrive` / `teams` and `execute` `microsoft365_get_me`

💘 Generated with Crush


acmacalister (commented):
Solid Microsoft 365 adapter — dispatch/compaction/markdown parity, PKCE OAuth, and the Graph response shaping all look in good shape. CI is green across build/test/lint/security/compose/rust-sdk. A few path-URL and next_link hardening notes inline; the OneDrive path colon issue looks like a real functional bug for path-based file ops.

integrations/microsoft365/files.go
```diff
@@ -0,0 +1,244 @@
...
+		path := r.Str("path")
+		params := graphListParams(r)
+		if err := r.Err(); err != nil {
+			return "", nil, err
+		}
+		item, err := driveItemPath(userID, driveID, itemID, path)
+		if err != nil {
+			return "", nil, err
+		}
+		return item + "/children" + queryEncode(params), nil, nil
```
acmacalister: Path-based folder listing is missing the trailing colon Graph requires before the relationship.

`driveItemPath` returns something like `/me/drive/root:/Documents/Reports`, and we then append `/children`, which becomes:

```
/me/drive/root:/Documents/Reports/children
```

Per the Graph docs it needs to be:

```
/me/drive/root:/Documents/Reports:/children
```

Same shape applies to download (`:/content`) and create-folder on a path parent. Item-id based paths are fine; path-based ones will 404 or resolve incorrectly without the colon.

```go
func driveChildrenPath(item string) string {
	if strings.Contains(item, "/root:/") && !strings.HasSuffix(item, ":") {
		return item + ":/children"
	}
	return item + "/children"
}
```

Worth a small httptest assertion for `path: "Documents/Reports"` so this doesn't regress.

integrations/microsoft365/files.go
```diff
@@ -0,0 +1,244 @@
...
+		return mcp.ErrResult(err)
+	}
+	if maxBytes > maxDownloadBytes {
+		maxBytes = maxDownloadBytes
+	}
+	item, err := driveItemPath(userID, driveID, itemID, path)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	data, ct, err := m.getRaw(ctx, item+"/content")
```
acmacalister: Same colon issue on download by path: `item+"/content"` yields `/me/drive/root:/Notes.txt/content` instead of `/me/drive/root:/Notes.txt:/content`.

Upload already gets this right (`:/name:/content`), so download/list/create-folder just need to match that convention when the item was resolved via path.

integrations/microsoft365/files.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+func driveItemPath(userID, driveID, itemID, path string) (string, error) {
+	root := driveRoot(userID, driveID)
+	if itemID != "" {
+		return root + "/items/" + url.PathEscape(itemID), nil
+	}
+	if path != "" {
+		path = strings.TrimPrefix(path, "/")
+		return root + "/root:/" + path, nil
```
acmacalister: Path segments go into the URL raw here. Spaces and reserved characters in something like `My Documents/Q1 report` will produce an invalid Graph path.

We should encode each segment (and keep `/` as the separator), e.g.:

```go
parts := strings.Split(strings.Trim(path, "/"), "/")
for i, p := range parts {
	parts[i] = url.PathEscape(p)
}
return root + "/root:/" + strings.Join(parts, "/"), nil
```

integrations/microsoft365/files.go
```diff
@@ -0,0 +1,244 @@
...
+	data, err := nextOrGet(ctx, m, args, func() (string, map[string]string, error) {
+		r := mcp.NewArgs(args)
+		userID := r.Str("user_id")
+		driveID := r.Str("drive_id")
+		q := r.Str("q")
+		params := graphListParams(r)
+		if err := r.Err(); err != nil {
+			return "", nil, err
+		}
+		return driveRoot(userID, driveID) + "/root/search(q='" + url.PathEscape(q) + "')" + queryEncode(params), nil, nil
```
acmacalister: `url.PathEscape(q)` inside the OData string literal turns multi-word queries into `budget%20report`, which Graph search usually does not want. It also is not the OData string escape (double single-quotes).

Safer approach is to escape `'` as `''` and leave the rest alone (or use a query builder):

```go
q = strings.ReplaceAll(q, "'", "''")
return driveRoot(userID, driveID) + "/root/search(q='" + q + "')" + queryEncode(params), nil, nil
```

integrations/microsoft365/microsoft365.go
```diff
@@ -0,0 +1,435 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if len(headers) > 0 {
+		return m.getWithHeaders(ctx, path, headers)
+	}
+	return m.get(ctx, path)
+}
+
+func allowedNextLink(next, baseURL string) bool {
```
acmacalister, line 401: Prefix matching on `next_link` is a bit brittle. With a custom `base_url` like `https://graph.microsoft.us` (no path), a value such as `https://graph.microsoft.us.evil.com/...` would pass `HasPrefix` and we would attach the bearer token to a non-Graph host.

Parsing the URL and allowlisting hosts is sturdier:

```go
u, err := url.Parse(next)
if err != nil || (u.Scheme != "https") {
	return false
}
host := strings.ToLower(u.Hostname())
switch host {
case "graph.microsoft.com", "graph.microsoft.us", "graph.microsoft.de", "microsoftgraph.chinacloudapi.cn":
	return true
}
if base, err := url.Parse(baseURL); err == nil && strings.EqualFold(base.Hostname(), host) {
	return true
}
return false
```

integrations/microsoft365/oauth.go
```diff
@@ -0,0 +1,300 @@
...
+	clientID := oauthState.clientID
+	clientSecret := oauthState.clientSecret
+	tokenURL := oauthState.tokenURL
+	oauthState.mu.Unlock()
+
+	return exchangeToken(oauthState, tokenURL, data, clientID, clientSecret)
+}
+
+func exchangeToken(oauthState *OAuthState, tokenURL string, data url.Values, clientID, clientSecret string) error {
+	req, err := http.NewRequest("POST", tokenURL, strings.NewReader(data.Encode()))
```
acmacalister: Token exchange uses `http.NewRequest` without a context, so a hung Microsoft identity endpoint cannot be canceled from the callback request.

`RefreshAccessToken` already uses `NewRequestWithContext` — worth matching that here (and threading `r.Context()` from the web callback into `HandleM365Callback`).

integrations/microsoft365/files.go
daltoniam: Fixed in d75a334. `driveRelation` now inserts Graph's `:/children` (and `:/content`) for path-based items, with httptest coverage for `Documents/Reports`.

integrations/microsoft365/files.go
daltoniam: Fixed in d75a334. Path downloads now use `driveRelation(item, "content")`, so `/root:/Notes.txt:/content` is asserted in tests.

integrations/microsoft365/files.go
daltoniam: Fixed in d75a334. Each path segment is `url.PathEscape`d in `encodeDrivePath`, with a test for `My Documents/Q1 report`.

integrations/microsoft365/files.go
daltoniam: Fixed in d75a334. Drive search now OData-escapes `'` as `''` instead of PathEscape, so multi-word queries stay intact.

integrations/microsoft365/microsoft365.go
daltoniam: Fixed in d75a334. `allowedNextLink` now parses the URL and allowlists Graph hosts (plus the configured base hostname), so `graph.microsoft.us.evil.com` is rejected.

integrations/microsoft365/oauth.go
daltoniam: Fixed in d75a334. Token exchange uses `NewRequestWithContext`, and the web callback passes `r.Context()` into `HandleM365Callback`.

integrations/microsoft365/microsoft365.go
```diff
@@ -0,0 +1,450 @@
...
+func graphListParams(r *mcp.Args) map[string]string {
+	top := fmt.Sprintf("%d", r.OptInt("top", defaultPageSize))
+	params := map[string]string{
+		"$filter":    r.Str("filter"),
+		"$select":    r.Str("select"),
+		"$search":    r.Str("search"),
+		"$orderby":   r.Str("orderby"),
+		"$top":       top,
+		"$skip":      r.Str("skip"),
+		"$skiptoken": r.Str("skiptoken"),
```
acmacalister: `graphListParams` always emits `$skiptoken` from a bare token, but Graph's `@odata.nextLink` usually carries opaque query state (`$skiptoken=...` or similar) as a full URL. The tools already expose `next_link` for that, and `nextOrGet` handles it correctly.

Having both `skiptoken` and `next_link` is a bit of a footgun for the LLM — agents often paste a fragment of the next link into `skiptoken` and get a 400. Worth either dropping the bare `skiptoken`/`skip` params from the tool schemas (prefer `next_link` only), or documenting that `skiptoken` is only the raw token value and `next_link` is preferred.

```go
// Prefer next_link for pagination; bare $skiptoken is easy to misuse.
params := map[string]string{
    "$filter":  r.Str("filter"),
    "$select":  r.Str("select"),
    "$search":  r.Str("search"),
    "$orderby": r.Str("orderby"),
    "$top":     top,
}
```

integrations/microsoft365/oauth.go
```diff
@@ -0,0 +1,303 @@
...
+	"net/url"
+	"strings"
+	"sync"
+	"time"
+)
+
+const (
+	msAuthorizeURL = "https://login.microsoftonline.com/%s/oauth2/v2.0/authorize"
+	msTokenURL     = "https://login.microsoftonline.com/%s/oauth2/v2.0/token"
+	msDefaultScope = "openid profile offline_access User.Read User.ReadBasic.All Mail.ReadWrite Mail.Send Calendars.ReadWrite Files.ReadWrite Team.ReadBasic.All Channel.ReadBasic.All ChannelMessage.Read.All ChannelMessage.Send Chat.ReadWrite Tasks.ReadWrite People.Read"
```
acmacalister: `list_users` hits `GET /users`, which typically needs `User.Read.All` or `Directory.Read.All` (delegated). The default scope string only requests `User.Read` + `User.ReadBasic.All`, and the setup page copy only mentions `User.Read`.

That mismatch means a first-time OAuth setup following the UI will often 403 on directory listing while mail/files/tasks work fine. Either widen the default scopes (and the Entra permission checklist on the setup page) to cover directory list, or tone down `list_users` / point agents at `search_people` (`People.Read`) as the coworker-lookup path.

Same kind of gap may apply to some Teams message scopes (`ChannelMessage.*` often needs admin consent) — worth calling out in the setup instructions so consent failures aren't mysterious.

integrations/microsoft365/files.go
```diff
@@ -0,0 +1,269 @@
...
+		payload = []byte(content)
+		if contentType == "" {
+			contentType = "text/plain"
+		}
+	}
+	parent, err := driveItemPath(userID, driveID, itemID, path)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	data, err := m.put(ctx, uploadContentPath(parent, encodeDrivePath(name)), payload, contentType)
```
acmacalister, line 244: Simple PUT upload is fine for small files, but Graph rejects simple uploads over ~4 MB and wants an upload session instead. The tool description mentions "under ~4 MB", yet there's no client-side size check before the PUT — agents will only learn via a Graph 400/413.

A cheap guard (and a clearer error) would help:

```go
const maxSimpleUpload = 4 * 1024 * 1024
if len(payload) > maxSimpleUpload {
    return mcp.ErrResult(fmt.Errorf(
        "content is %d bytes; simple upload supports up to 4 MB — split the file or use a smaller payload",
        len(payload),
    ))
}
```

web/web.go
```diff
@@ -3785,3 +3793,173 @@ func (w *WebServer) handleXSaveOAuthCredentials(rw http.ResponseWriter, r *http.
...
+	if ic == nil {
+		ic = &mcp.IntegrationConfig{Credentials: mcp.Credentials{}}
+	}
+	ic.Enabled = true
+	ic.Credentials["access_token"] = result.AccessToken
+	if result.RefreshToken != "" {
+		ic.Credentials["refresh_token"] = result.RefreshToken
+	}
+	ic.Credentials[mcp.CredKeyTokenSource] = "oauth"
+	_ = w.services.Config.SetIntegration("microsoft365", ic)
```
acmacalister: After OAuth completes we persist tokens + `Enabled=true`, but we never call `Configure` / `ConfigureIntegration` on the live registry adapter here (same pattern as Gmail/X today).

If the process was already running with microsoft365 disabled or with an empty token, MCP `execute` won't pick up the new credentials until restart, even though the setup page health check configures a temporary instance and looks green. Not unique to this PR, but this is a large surface (mail send, file write, Teams) where "Connected" in the UI vs "not configured" on the wire will confuse people.

Worth wiring a `ConfigureIntegration` after a successful save (callback + manual token), or documenting that a restart is required.

acmacalister (approved):
Nice Microsoft 365 adapter — dispatch/compaction/markdown parity looks solid, and the follow-ups on path colon encoding, next_link host checks, and OAuth context all landed cleanly. CI is green (build/test/lint/security/compose/rust-sdk). A few remaining notes on Graph pagination ergonomics, default scopes vs list_users, upload size guard, and live reconfigure after OAuth.

integrations/microsoft365/microsoft365.go
daltoniam: Fixed in 519fc05. Dropped bare `skiptoken`/`skip` from `graphListParams` and the tool schemas so pagination goes through `next_link` only.

integrations/microsoft365/oauth.go
daltoniam: Fixed in 519fc05. Default OAuth scope now includes `User.Read.All`, `list_users` points agents at `search_people` when that consent is missing, and the setup page calls out admin-consent scopes.

integrations/microsoft365/files.go
daltoniam: Fixed in 519fc05. Simple PUT uploads over 4 MB now fail client-side with a clear error before hitting Graph.

web/web.go
daltoniam: Fixed in 519fc05. OAuth callback and manual token save now `ConfigureIntegration` the live registry adapter, persist, and notify — with rollback if the config write fails.

daltoniam:
Retriggering CI for 519fc05.

acmacalister (commented):
Solid Microsoft 365 adapter — dispatch/compaction/markdown parity, path encoding, next_link host checks, upload guard, and live Configure-after-OAuth all look in good shape, and prior review threads are resolved. CI is green (build/test/lint/security/compose/rust-sdk).

A few remaining notes around Graph `$search` quoting (mail/people/directory) and SharePoint scope vs tool copy — details inline.

integrations/microsoft365/microsoft365.go
```diff
@@ -0,0 +1,449 @@
...
+	}
+	return "/users/" + url.PathEscape(userID)
+}
+
+func graphListParams(r *mcp.Args) map[string]string {
+	top := fmt.Sprintf("%d", r.OptInt("top", defaultPageSize))
+	params := map[string]string{
+		"$filter":  r.Str("filter"),
+		"$select":  r.Str("select"),
+		"$search":  r.Str("search"),
```
acmacalister: Graph's `$search` values need to be double-quoted. From the docs, message search is `$search="pizza"` / `$search="from:ada"`, not a bare token — same for people and directory queries.

Right now `graphListParams` passes the raw `search` arg straight through, and the tool examples (`from:ada@contoso.com subject:budget`) steer agents toward the unquoted form. That usually 400s or silently misbehaves once this hits a real tenant.

Worth normalizing here (or at each call site) so we accept either form:

```go
func quoteGraphSearch(q string) string {
	q = strings.TrimSpace(q)
	if q == "" {
		return ""
	}
	if len(q) >= 2 && q[0] == '"' && q[len(q)-1] == '"' {
		return q
	}
	q = strings.ReplaceAll(q, `\`, `\\`)
	q = strings.ReplaceAll(q, `"`, `\"`)
	return `"` + q + `"`
}
```

And flip `TestListMessages` to assert the encoded quotes land on the wire.

integrations/microsoft365/users.go
```diff
@@ -0,0 +1,84 @@
...
+func searchPeople(ctx context.Context, m *m365, args map[string]any) (*mcp.ToolResult, error) {
+	data, err := nextOrGet(ctx, m, args, func() (string, map[string]string, error) {
+		r := mcp.NewArgs(args)
+		q := r.Str("search")
+		params := graphListParams(r)
+		if err := r.Err(); err != nil {
+			return "", nil, err
+		}
+		if q != "" {
+			params["$search"] = q
```
acmacalister: Same `$search` quoting issue on people search — docs want `$search="Irene McGowan"`, not a bare string.

For `list_users` the bar is higher still: directory `$search` wants property clauses like `"displayName:Ada" OR "mail:Ada"` plus `ConsistencyLevel: eventual` (which we already set). A bare name will not do what the tool description implies.

If we only auto-quote here, people search gets fixed; for directory search we should either rewrite a bare term into `displayName`/`mail` clauses or tighten the tool parameter docs so agents pass the Graph form.

integrations/microsoft365/tools.go
```diff
@@ -0,0 +1,369 @@
...
+		Parameters: map[string]string{
+			"select": "OData $select fields (e.g. id,displayName,mail,userPrincipalName)",
+		},
+	},
+	{
+		Name: mcp.ToolName("microsoft365_list_users"), Description: "List Microsoft Entra ID directory users. Needs User.Read.All (admin consent). Prefer search_people for coworker lookup with People.Read.",
+		Parameters: map[string]string{
+			"filter":    "OData $filter (e.g. startswith(displayName,'Ada'))",
+			"select":    "OData $select fields",
+			"search":    "OData $search across displayName and mail (requires ConsistencyLevel)",
```
acmacalister: This description undersells the Graph requirement. Directory `$search` is not a free-text bag — it needs quoted property clauses, e.g. `"displayName:Ada" OR "mail:Ada"`, with `ConsistencyLevel: eventual`.

Something like:

```text
search: OData $search clause(s), e.g. "displayName:Ada" OR "mail:Ada" (requires ConsistencyLevel; admin consent for User.Read.All)
```

Otherwise agents will try plain names and bounce off Graph 400s while `search_people` was the easier path anyway.

integrations/microsoft365/oauth.go
```diff
@@ -0,0 +1,303 @@
...
+	"net/url"
+	"strings"
+	"sync"
+	"time"
+)
+
+const (
+	msAuthorizeURL = "https://login.microsoftonline.com/%s/oauth2/v2.0/authorize"
+	msTokenURL     = "https://login.microsoftonline.com/%s/oauth2/v2.0/token"
+	msDefaultScope = "openid profile offline_access User.Read User.ReadBasic.All User.Read.All Mail.ReadWrite Mail.Send Calendars.ReadWrite Files.ReadWrite Team.ReadBasic.All Channel.ReadBasic.All ChannelMessage.Read.All ChannelMessage.Send Chat.ReadWrite Tasks.ReadWrite People.Read"
```
acmacalister: Default file scope is `Files.ReadWrite`, which covers the signed-in user's OneDrive. Tool copy and compaction talk about SharePoint too — site libraries the user can access generally need `Files.ReadWrite.All` (and sometimes `Sites.Read.All`).

Not a blocker if we're intentionally starting with personal OneDrive, but then the drive tool descriptions should say so, or we should widen the default scope / setup checklist so SharePoint paths don't 403 after a "successful" OAuth.

integrations/microsoft365/microsoft365.go
daltoniam: Fixed in bd97480. `graphListParams` now wraps `$search` with `quoteGraphSearch`, and `TestListMessages` asserts the quoted value on the wire.

integrations/microsoft365/users.go
daltoniam: Fixed in bd97480. People search uses the same quoting helper. Bare `list_users` terms become `"displayName:Ada" OR "mail:Ada"`; already-quoted or OR/AND clauses pass through.

integrations/microsoft365/tools.go
daltoniam: Fixed in bd97480. `list_users` search docs now show the Graph clause form and note that a bare name is rewritten.

integrations/microsoft365/oauth.go
daltoniam: Fixed in bd97480. Default OAuth scope includes `Files.ReadWrite.All`, and the setup page checklist calls out OneDrive and SharePoint.

integrations/microsoft365/microsoft365.go
```diff
@@ -0,0 +1,478 @@
...
+	r := mcp.NewArgs(args)
+	next := r.Str("next_link")
+	if err := r.Err(); err != nil {
+		return nil, err
+	}
+	if next != "" {
+		if !allowedNextLink(next, m.baseURL) {
+			return nil, fmt.Errorf("next_link must be a Microsoft Graph URL")
+		}
+		return m.get(ctx, next)
```
acmacalister, line 380: When agents page a `list_users` (or other directory) `$search` result via `next_link`, we drop the `ConsistencyLevel: eventual` header that the first page set.

Graph's paging docs call this out for directory resources: custom headers like `ConsistencyLevel` are not carried on `@odata.nextLink`, so subsequent pages need the header re-applied or the request fails / returns incomplete results. `list_messages` `$search` has the same shape if we keep sending the header on page 1.

Something like detecting directory advanced-query next links (or always attaching `ConsistencyLevel` when the next URL still has `$search` / `$count`) would close the gap:

```go
if next != "" {
	if !allowedNextLink(next, m.baseURL) {
		return nil, fmt.Errorf("next_link must be a Microsoft Graph URL")
	}
	headers := map[string]string{}
	if needsEventualConsistency(next) {
		headers["ConsistencyLevel"] = "eventual"
	}
	if len(headers) > 0 {
		return m.getWithHeaders(ctx, next, headers)
	}
	return m.get(ctx, next)
}
```

Worth a small httptest case that follows a `list_users` next_link and asserts the header is still present.

acmacalister (approved):
Solid Microsoft 365 adapter — prior review threads look resolved, and CI is green (build/test/lint/security/compose/rust-sdk). One remaining note on directory `$search` pagination keeping `ConsistencyLevel` on `next_link` follow-ups; details inline.

integrations/microsoft365/microsoft365.go
daltoniam: Fixed in ba02f1e. `nextOrGet` re-sends `ConsistencyLevel: eventual` when the next URL still has `$search` or `$count`, with a `list_users` httptest asserting the header on page 2.

integrations/microsoft365/microsoft365.go
```diff
@@ -0,0 +1,490 @@
...
+	}
+	if resp.StatusCode == 429 || resp.StatusCode >= 500 {
+		re := &mcp.RetryableError{StatusCode: resp.StatusCode, Err: fmt.Errorf("microsoft365 API error (%d): %s", resp.StatusCode, string(data))}
+		re.RetryAfter = mcp.ParseRetryAfter(resp.Header.Get("Retry-After"))
+		return nil, contentType, re
+	}
+	if resp.StatusCode >= 400 {
+		return nil, contentType, fmt.Errorf("microsoft365 API error (%d): %s", resp.StatusCode, string(data))
+	}
+	if resp.StatusCode == 202 || resp.StatusCode == 204 || len(data) == 0 {
```
acmacalister: `getRaw` (used by `download_drive_item`) goes through this path too, so a real empty file (`200` + zero-length body) gets rewritten to `{"status":"success"}` before `downloadEnvelope` runs. Agents then see a fake success payload instead of empty content.

Worth treating empty bodies as success only for the write/delete statuses (202/204), or branching in `getRaw` so binary/text downloads keep a true empty slice:

```go
if resp.StatusCode == 202 || resp.StatusCode == 204 {
	return json.RawMessage(`{"status":"success"}`), contentType, nil
}
if len(data) == 0 && strings.Contains(contentType, "json") {
	return json.RawMessage(`{"status":"success"}`), contentType, nil
}
return json.RawMessage(data), contentType, nil
```

A small httptest case for `download_drive_item` with an empty body would lock this in.

integrations/microsoft365/mail.go
```diff
@@ -0,0 +1,168 @@
...
+func listMessages(ctx context.Context, m *m365, args map[string]any) (*mcp.ToolResult, error) {
+	data, err := nextOrGet(ctx, m, args, func() (string, map[string]string, error) {
+		r := mcp.NewArgs(args)
+		userID := r.Str("user_id")
+		folderID := r.Str("folder_id")
+		params := graphListParams(r)
+		if err := r.Err(); err != nil {
+			return "", nil, err
+		}
+		if params["$orderby"] == "" && params["$search"] == "" {
```
acmacalister: Default `$orderby` is correctly skipped when `$search` is set, but a caller-supplied `orderby` still rides along with `$search` via `graphListParams`. Graph rejects that combination on messages (`$orderby` and `$search` cannot be used together), so agents that pass both get a 400.

Dropping `$orderby` whenever `$search` is present (not only when it's empty) would match the Graph constraint:

```go
if params["$search"] != "" {
	delete(params, "$orderby")
	headers["ConsistencyLevel"] = "eventual"
} else if params["$orderby"] == "" {
	params["$orderby"] = "receivedDateTime desc"
}
```

acmacalister (approved):
Solid Microsoft 365 adapter — prior review threads look resolved, and CI is green (build/test/lint/security/compose/rust-sdk). A couple of leftover edge cases on empty file downloads and message `$search`+`$orderby` inline; nothing blocking.

integrations/microsoft365/microsoft365.go
daltoniam: Fixed in 68613e8. Empty 200 bodies are no longer rewritten to `{"status":"success"}` unless the content type is JSON; `download_drive_item` now keeps a true empty file.

integrations/microsoft365/mail.go
daltoniam: Fixed in 68613e8. `list_messages` drops `$orderby` whenever `$search` is set, including caller-supplied orderby, so Graph does not 400 on the combination.

acmacalister (approved):
Clean PR — prior review threads are resolved, and CI is green on this head (build/lint/security/compose/rust-sdk, with the matching CI run succeeding). Dispatch/compaction/markdown parity, Graph path encoding, next_link host checks, OAuth PKCE + live Configure-after-save, and the empty-download / search+orderby fixes all look solid. LGTM.

PR #228, opened 2026-09-06
Add Intercom customer support integration

daltoniam:
## Summary
- Add an Intercom adapter so inbox conversations, contacts, tickets, and help-center articles are available through search and execute.
- Support Bearer token auth with optional EU/AU base URLs, plus markdown rendering for conversation and article bodies.

## Test plan
- [ ] `go test ./integrations/intercom/ ./config/`
- [ ] Enable Intercom with `INTERCOM_ACCESS_TOKEN` and search for `intercom conversations`
- [ ] Execute `intercom_search_conversations` then `intercom_get_conversation` on a real inbox thread

💘 Generated with Crush

### DIFF
--- a/integrations/intercom/handlers.go
+++ b/integrations/intercom/handlers.go
@@ -437,7 +437,7 @@ func searchTickets(ctx context.Context, c *intercom, args map[string]any) (*mcp.
 		}
 		query = q
 	} else if state != "" {
-		query = equalityQuery("ticket_state", state)
+		query = equalityQuery("state", state)
 	} else {
 		query = equalityQuery("open", true)
 	}
@@ -669,17 +669,29 @@ func listTags(ctx context.Context, c *intercom, _ map[string]any) (*mcp.ToolResu
 	return mcp.RawResult(data)
 }
 
+func listTicketTypes(ctx context.Context, c *intercom, _ map[string]any) (*mcp.ToolResult, error) {
+	data, err := c.get(ctx, "/ticket_types")
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	return mcp.RawResult(data)
+}
+
 func applyReplyActor(payload map[string]any, replyType, adminID, userID string) error {
-	if replyType == "admin" {
+	switch replyType {
+	case "admin":
 		if adminID == "" {
 			return fmt.Errorf("admin_id is required when type is admin")
 		}
 		payload["admin_id"] = adminID
 		return nil
+	case "user":
+		if userID == "" {
+			return fmt.Errorf("intercom_user_id is required when type is user")
+		}
+		payload["intercom_user_id"] = userID
+		return nil
+	default:
+		return fmt.Errorf("type must be admin or user, got %q", replyType)
 	}
-	if userID == "" {
-		return fmt.Errorf("intercom_user_id is required when type is user")
-	}
-	payload["intercom_user_id"] = userID
-	return nil
 }
--- a/integrations/intercom/handlers.go
+++ b/integrations/intercom/handlers.go
@@ -356,8 +356,8 @@ func createContact(ctx context.Context, c *intercom, args map[string]any) (*mcp.
 	if err := r.Err(); err != nil {
 		return mcp.ErrResult(err)
 	}
-	if len(payload) == 0 {
-		return mcp.ErrResult(fmt.Errorf("provide at least one of email, external_id, or name"))
+	if payload["email"] == nil && payload["external_id"] == nil {
+		return mcp.ErrResult(fmt.Errorf("provide email or external_id"))
 	}
 	data, err := c.post(ctx, "/contacts", payload)
 	if err != nil {
--- a/integrations/intercom/handlers.go
+++ b/integrations/intercom/handlers.go
@@ -152,13 +152,8 @@ func replyConversation(ctx context.Context, c *intercom, args map[string]any) (*
 		"type":         replyType,
 		"body":         body,
 	}
-	if replyType == "admin" {
-		if adminID == "" {
-			return mcp.ErrResult(fmt.Errorf("admin_id is required when type is admin"))
-		}
-		payload["admin_id"] = adminID
-	} else if userID != "" {
-		payload["intercom_user_id"] = userID
+	if err := applyReplyActor(payload, replyType, adminID, userID); err != nil {
+		return mcp.ErrResult(err)
 	}
 	if len(attachments) > 0 {
 		payload["attachment_urls"] = attachments
@@ -488,15 +483,23 @@ func createTicket(ctx context.Context, c *intercom, args map[string]any) (*mcp.T
 	} else if contactID != "" {
 		payload["contacts"] = []any{map[string]any{"id": contactID}}
 	}
+	if _, ok := payload["contacts"]; !ok {
+		return mcp.ErrResult(fmt.Errorf("provide contact_id or contacts"))
+	}
 	if title != "" {
 		payload["ticket_attributes"] = map[string]any{"_default_title_": title, "_default_description_": description}
 	} else if description != "" {
 		payload["ticket_attributes"] = map[string]any{"_default_description_": description}
 	}
+	assignment := map[string]any{}
 	if adminAssignee != "" {
-		payload["assignment"] = map[string]any{"admin_assignee_id": adminAssignee, "team_assignee_id": teamAssignee}
-	} else if teamAssignee != "" {
-		payload["assignment"] = map[string]any{"team_assignee_id": teamAssignee}
+		assignment["admin_assignee_id"] = adminAssignee
+	}
+	if teamAssignee != "" {
+		assignment["team_assignee_id"] = teamAssignee
+	}
+	if len(assignment) > 0 {
+		payload["assignment"] = assignment
 	}
 	data, err := c.post(ctx, "/tickets", payload)
 	if err != nil {
@@ -508,6 +511,7 @@ func createTicket(ctx context.Context, c *intercom, args map[string]any) (*mcp.T
 func updateTicket(ctx context.Context, c *intercom, args map[string]any) (*mcp.ToolResult, error) {
 	r := mcp.NewArgs(args)
 	id := r.Str("ticket_id")
+	assigneeID := r.Str("assignee_id")
 	adminAssignee := r.Str("admin_assignee_id")
 	teamAssignee := r.Str("team_assignee_id")
 	attrsRaw := r.Str("ticket_attributes")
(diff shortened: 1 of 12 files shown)


integrations/intercom/handlers.go
```diff
@@ -0,0 +1,666 @@
...
+	payload := map[string]any{}
+	if _, ok := args["open"]; ok {
+		open, err := mcp.ArgBool(args, "open")
+		if err != nil {
+			return mcp.ErrResult(err)
+		}
+		payload["open"] = open
+	}
+	if adminAssignee != "" {
+		payload["assignment"] = map[string]any{"admin_assignee_id": adminAssignee, "team_assignee_id": teamAssignee}
```
acmacalister: `updateTicket` builds a nested `assignment` object with `admin_assignee_id` / `team_assignee_id`, but with `Intercom-Version: 2.16` the Update Ticket API expects top-level `assignee_id` (admin or team ID, or `"0"` to unassign) plus optional `admin_id` for the actor. The nested form is the Create Ticket shape and was replaced for updates in 2.14+.

As written, assign-on-update will 400 or no-op against real workspaces. Something closer to:

```go
if adminAssignee != "" || teamAssignee != "" {
    // Intercom takes a single assignee_id (admin OR team).
    if adminAssignee != "" {
        payload["assignee_id"] = adminAssignee
    } else {
        payload["assignee_id"] = teamAssignee
    }
}
```

Tool params probably want a single `assignee_id` (and maybe `admin_id`) to match the API, rather than two create-style fields.

integrations/intercom/handlers.go
```diff
@@ -0,0 +1,666 @@
...
+	payload := map[string]any{"ticket_type_id": ticketTypeID}
+	if contactsRaw != "" {
+		contacts, err := parseJSONArray(contactsRaw)
+		if err != nil {
+			return mcp.ErrResult(err)
+		}
+		payload["contacts"] = contacts
+	} else if contactID != "" {
+		payload["contacts"] = []any{map[string]any{"id": contactID}}
+	}
```
acmacalister, line 485: Create Ticket requires `contacts` (exactly one today per Intercom). We accept `ticket_type_id` alone and only attach contacts when provided, so the LLM can hit a guaranteed 400.

Worth failing early when both `contacts` and `contact_id` are empty, same pattern as `searchContacts` requiring email/query:

```go
if _, ok := payload["contacts"]; !ok {
    return mcp.ErrResult(fmt.Errorf("provide contact_id or contacts"))
}
```

integrations/intercom/handlers.go
```diff
@@ -0,0 +1,666 @@
...
+		"type":         replyType,
+		"body":         body,
+	}
+	if replyType == "admin" {
+		if adminID == "" {
+			return mcp.ErrResult(fmt.Errorf("admin_id is required when type is admin"))
+		}
+		payload["admin_id"] = adminID
+	} else if userID != "" {
+		payload["intercom_user_id"] = userID
```
acmacalister: When `type` is `user` and `intercom_user_id` is missing we still POST the reply. Admin path correctly requires `admin_id`; the user path should mirror that (or require email/`user_id` if we want those Intercom variants) instead of sending a body Intercom will reject.

integrations/intercom/handlers.go
```diff
@@ -0,0 +1,666 @@
...
+	} else if contactID != "" {
+		payload["contacts"] = []any{map[string]any{"id": contactID}}
+	}
+	if title != "" {
+		payload["ticket_attributes"] = map[string]any{"_default_title_": title, "_default_description_": description}
+	} else if description != "" {
+		payload["ticket_attributes"] = map[string]any{"_default_description_": description}
+	}
+	if adminAssignee != "" {
+		payload["assignment"] = map[string]any{"admin_assignee_id": adminAssignee, "team_assignee_id": teamAssignee}
```
acmacalister: When only `admin_assignee_id` is set this still sends `"team_assignee_id": ""` inside `assignment`. Safer to omit empty keys so Intercom doesn't treat blank as an explicit unassign/invalid id:

```go
assignment := map[string]any{}
if adminAssignee != "" {
    assignment["admin_assignee_id"] = adminAssignee
}
if teamAssignee != "" {
    assignment["team_assignee_id"] = teamAssignee
}
if len(assignment) > 0 {
    payload["assignment"] = assignment
}
```

acmacalister (commented):
Solid Intercom adapter — wiring, dispatch/compaction/markdown parity, Bearer + version headers, and the happy-path tests all look right. CI is green (build, test, lint, security, rust-sdk, compose).

A few API-shape issues on ticket update assignment and create/reply validation are worth fixing before this hits real workspaces; details inline.

integrations/intercom/handlers.go
daltoniam: Fixed: update now sends top-level `assignee_id` (2.16 shape). `admin_assignee_id`/`team_assignee_id` still work as fallbacks when `assignee_id` is omitted.

integrations/intercom/handlers.go
daltoniam: Fixed: create now fails early with `provide contact_id or contacts` when neither is set.

integrations/intercom/handlers.go
daltoniam: Fixed: user replies now require `intercom_user_id`, matching the admin_id check. Same validation on ticket replies.

integrations/intercom/handlers.go
daltoniam: Fixed: create assignment now only includes the keys that were actually set.

integrations/intercom/handlers.go
```diff
@@ -0,0 +1,685 @@
...
+	}
+	var query any
+	if queryRaw != "" {
+		q, err := parseJSONObject(queryRaw)
+		if err != nil {
+			return mcp.ErrResult(err)
+		}
+		query = q
+	} else if state != "" {
+		query = equalityQuery("ticket_state", state)
```
acmacalister: The tickets search Accepted Fields table uses `state` (string), not `ticket_state`. `ticket_state` is the nested object on the Ticket model response, so filtering with `state: submitted` here will fail at Intercom once someone uses the shortcut instead of raw `query`.

```go
query = equalityQuery("state", state)
```

Worth a quick httptest test that the body field is `state` when only `state` is passed.

integrations/intercom/handlers.go
```diff
@@ -0,0 +1,685 @@
...
+
+func listTags(ctx context.Context, c *intercom, _ map[string]any) (*mcp.ToolResult, error) {
+	data, err := c.get(ctx, "/tags")
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	return mcp.RawResult(data)
+}
+
+func applyReplyActor(payload map[string]any, replyType, adminID, userID string) error {
```
acmacalister, line 680: Anything other than `"admin"` is treated as a user reply here — including typos like `type: "note"` (easy to confuse with `message_type`) or `type: "team"`. Those fail with `intercom_user_id is required` instead of a clear invalid-type error.

```go
switch replyType {
case "admin":
    if adminID == "" {
        return fmt.Errorf("admin_id is required when type is admin")
    }
    payload["admin_id"] = adminID
case "user":
    if userID == "" {
        return fmt.Errorf("intercom_user_id is required when type is user")
    }
    payload["intercom_user_id"] = userID
default:
    return fmt.Errorf("type must be admin or user, got %q", replyType)
}
```

integrations/intercom/tools.go
```diff
@@ -0,0 +1,244 @@
...
+			"starting_after": "Pagination cursor from pages.next.starting_after",
+		},
+	},
+	{
+		Name: mcp.ToolName("intercom_get_ticket"), Description: "Get an Intercom ticket including attributes, contacts, and assignment. Use after search_tickets.",
+		Parameters: map[string]string{"ticket_id": "Intercom ticket ID"},
+		Required:   []string{"ticket_id"},
+	},
+	{
+		Name: mcp.ToolName("intercom_create_ticket"), Description: "Create an Intercom ticket for a contact. Requires a ticket type from the workspace.",
```
acmacalister: `create_ticket` needs a `ticket_type_id`, but there's no list/search tool for ticket types (Intercom has `GET /ticket_types`). Same class of gap as the contact requirement we already guard in the handler — the LLM has nowhere to discover valid type IDs short of guessing from prior context.

Adding `intercom_list_ticket_types` (and mentioning it in this description) would make the create path actually usable end-to-end.

acmacalister (commented):
Nice follow-up on the assignment/reply validation — that matches the 2.16 shapes. CI is green (build, test, lint, security, rust-sdk, compose). One real search-field mismatch plus a couple of LLM-usability nits inline.

integrations/intercom/handlers.go
daltoniam: Fixed: ticket search now queries `state`, with a test covering the shortcut body.

integrations/intercom/handlers.go
daltoniam: Fixed: `type` is now an explicit admin/user switch; anything else returns `type must be admin or user`.

integrations/intercom/tools.go
daltoniam: Fixed: added `intercom_list_ticket_types` and pointed `create_ticket` at it for `ticket_type_id` discovery.

integrations/intercom/handlers.go
```diff
@@ -0,0 +1,697 @@
...
+	r := mcp.NewArgs(args)
+	payload, err := contactPayload(r)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	if len(payload) == 0 {
+		return mcp.ErrResult(fmt.Errorf("provide at least one of email, external_id, or name"))
```
acmacalister: This early check and the tool description disagree a bit. The description says "Provide email and/or external_id", and Intercom's create-contact payload is one of email / external_id / role — name alone is optional and will 400 if that's all we send.

Worth failing when neither identifier is present, same spirit as the create-ticket contacts guard:

```go
if payload["email"] == nil && payload["external_id"] == nil {
    return mcp.ErrResult(fmt.Errorf("provide email or external_id"))
}
```

(Or allow `role` if we intentionally support that create variant.)

integrations/intercom/tools.go
```diff
@@ -0,0 +1,248 @@
...
+			"contact_id":        "Intercom contact ID to attach (used when contacts is omitted)",
+			"contacts":          `JSON array of contacts (e.g. [{"id":"abc"}])`,
+			"title":             "Ticket title",
+			"description":       "Ticket description",
+			"admin_assignee_id": "Admin ID to assign",
+			"team_assignee_id":  "Team ID to assign",
+		},
+		Required: []string{"ticket_type_id"},
+	},
+	{
```
acmacalister, line 178: `Required` is only `ticket_type_id`, but the handler (correctly) rejects calls with no contact. LLMs tend to trust the schema more than free-text param docs, so they'll still hit the early error until they retry.

Adding `contact_id` here (or documenting that `contacts` is an alternate required form in the description more loudly) would match the handler guard and make create actually one-shot usable.

acmacalister (approved):
Solid Intercom adapter — prior assignment/reply/search/ticket-type feedback looks addressed, and CI is green (build, test, lint, security, rust-sdk, compose). Two small LLM-usability nits inline on create-contact validation and create-ticket Required; otherwise LGTM.

integrations/intercom/handlers.go
daltoniam: Fixed: create-contact now requires email or external_id; name alone is rejected.

integrations/intercom/tools.go
daltoniam: Fixed: `contact_id` is now in Required. `contacts` remains a documented alternate.

acmacalister (approved):
Clean PR — CI is green (build, test, lint, security, rust-sdk, compose) and prior assignment/reply/search/ticket-type/create-contact feedback looks addressed. Solid adapter with dispatch/compaction/markdown parity and good validation coverage. LGTM.

PR #229, opened 2026-09-06
Add HubSpot CRM integration

daltoniam:
## Summary
- Adds a HubSpot CRM adapter so agents can look up and update contacts, companies, deals, tickets, and associations from Switchboard.
- Uses private-app bearer tokens (`HUBSPOT_ACCESS_TOKEN`) against HubSpot CRM v3, with optional `HUBSPOT_BASE_URL`.

## Test plan
- [x] `go test ./integrations/hubspot/ ./config/`
- [x] Adapter dispatch/compaction tests
- [ ] CI on this PR


integrations/hubspot/handlers.go
```diff
@@ -0,0 +1,532 @@
...
+	if filters != nil {
+		body["filterGroups"] = []any{map[string]any{"filters": filters}}
+		return body, nil
+	}
+	if query != "" {
+		groups, err := queryFilterGroups(objectType, query)
+		if err != nil {
+			return nil, err
+		}
+		body["filterGroups"] = groups
```
acmacalister: `query` is expanded into up to five OR `filterGroups` with `CONTAINS_TOKEN`, but HubSpot already has a top-level search `query` field that hits the default searchable properties (including `email` / `hs_additional_emails` for contacts).

Two practical issues with the current path:
1. Contacts already sit at HubSpot's ~5 `filterGroups` cap, so we can't OR in extra properties later.
2. Token matching on a full email like `ada@example.com` is less reliable than HubSpot's own text search (or an `EQ` on `email` for exact lookups).

Worth preferring the API's `query` field for the simple text case, and only building filter groups when the caller passes `filters` / `filter_groups`:

```go
if query != "" {
    body["query"] = query
}
```

That also matches what the tool description promises ("simple text search") and is the path agents will hit first via `hubspot_search_contacts`.

acmacalister (approved):
Solid HubSpot adapter — clean hexagonal shape, dispatch/compaction parity, env wiring, and good httptest coverage. CI is green (build/test/lint/security/compose/rust-sdk). One note on how `query` is mapped to the CRM Search API; otherwise LGTM.

integrations/hubspot/handlers.go
daltoniam: Addressed in 7d5f9ba. Simple `query` now goes through HubSpot's top-level search `query` field; `filters` / `filter_groups` still build explicit filter groups.

acmacalister (approved):
Clean PR — CI is green (build/test/lint/security/compose/rust-sdk) and the earlier search-query feedback is addressed in 7d5f9ba. Solid HubSpot adapter with dispatch/compaction parity, shared helpers, path validation, and good httptest coverage. LGTM.

PR #230, opened 2026-09-06
Add ServiceNow ITSM Table API integration

daltoniam:
## Summary
- Add a ServiceNow adapter so agents can triage incidents, problems, changes, knowledge, and CMDB records from one MCP endpoint.
- Support basic username/password or OAuth bearer auth against a customer instance URL.

## Test plan
- [x] `go test ./config/ ./integrations/servicenow/`
- [ ] CI green on this PR
- [ ] Optional live smoke: search `servicenow incident`, then execute `servicenow_list_incidents`


integrations/servicenow/servicenow.go
```diff
@@ -0,0 +1,272 @@
...
+}
+
+func listQueryParams(args map[string]any, defaultFields string) map[string]string {
+	r := mcp.NewArgs(args)
+	query := r.Str("query")
+	fields := r.Str("fields")
+	displayValue := r.Str("display_value")
+	limit := r.OptInt("limit", 25)
+	offset := r.Int("offset")
+	_ = r.Err()
```
acmacalister: This drops type-coercion errors from `r.Str` / `r.Int` (e.g. `offset: "abc"` or a non-string `fields`). Every list tool goes through here, so a bad arg just silently becomes defaults and we still hit the API.

Worth returning the error instead of discarding it:

```go
    // caller should surface this via mcp.ErrResult
    // either return (nil, err) or change the signature to (*params, error)
}
```

`listTable` is the natural place to check and `return mcp.ErrResult(err)`.

integrations/servicenow/markdown.go
```diff
@@ -0,0 +1,349 @@
...
+		b.Attribution(parts...)
+	}
+	for _, kv := range rec.Extra {
+		if kv[1] != "" {
+			b.Raw("**" + kv[0] + ":** " + kv[1] + "\n")
+		}
+	}
+	if rec.Description != "" {
+		b.BlankLine()
+		b.Raw(rec.Description + "\n")
```
acmacalister: Incident/problem/change descriptions often come back as HTML from ServiceNow, but we dump them straight through `b.Raw` here. Knowledge articles already go through `markdown.FromHTML`, and comments through `stripHTML` — descriptions should match so the LLM doesn't get raw `<p>`/`<br>` noise.

```go
b.WriteMarkdown(markdown.FromHTML(rec.Description))
```

(or `stripHTML` if we want plain text instead of markdown).

integrations/servicenow/handlers.go
```diff
@@ -0,0 +1,454 @@
...
+		return mcp.ErrResult(err)
+	}
+	if err := validTable(table); err != nil {
+		return mcp.ErrResult(err)
+	}
+	body, err := parseDataJSON(dataStr)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	data, err := s.patch(ctx, fmt.Sprintf("/api/now/table/%s/%s", url.PathEscape(table), url.PathEscape(sysID)), body)
```
acmacalister, line 343: `updateIncident` rejects an empty body, but `updateRecord` will PATCH `{}` if `data` is `{}` or only whitespace-empty fields. Easy footgun for the generic path — same guard would help:

```go
if len(body) == 0 {
    return mcp.ErrResult(fmt.Errorf("provide at least one field to update"))
}
```

integrations/servicenow/compact.yaml
```diff
@@ -0,0 +1,244 @@
...
+            - result.short_description
+            - result.ip_address
+            - result.fqdn
+            - result.serial_number
+            - result.manufacturer
+            - result.model_id
+            - result.location
+            - result.owned_by
+            - result.sys_updated_on
+    servicenow_list_records:
```
acmacalister, line 219: `servicenow_list_records` has a compact spec, but `servicenow_get_record` / `servicenow_aggregate` don't. Generic get can return a full CI or custom-table row with a lot of noise (and no `max_bytes`), so the LLM pays for fields we already strip on the typed getters.

A minimal keep-list (or even `spec: []` + `max_bytes` like NetSuite's get) would keep the escape hatch usable without dumping unbounded payloads.

acmacalister (approved):
Solid new adapter — dispatch/compaction/markdown parity, dual auth, table validation, and CI (build/test/lint/security/compose) are all green. A few non-blocking notes inline on arg errors, HTML in descriptions, empty generic updates, and get_record compaction.

integrations/servicenow/servicenow.go
daltoniam: Fixed: `listQueryParams` now returns the `r.Err()` so `listTable` surfaces type-coercion failures instead of silently defaulting.

integrations/servicenow/markdown.go
daltoniam: Fixed: incident/problem/change descriptions now go through `markdown.FromHTML` like knowledge articles.

integrations/servicenow/handlers.go
daltoniam: Fixed: `updateRecord` now rejects empty `{}` bodies the same way `updateIncident` does.

integrations/servicenow/compact.yaml
daltoniam: Fixed: `servicenow_get_record` and `servicenow_aggregate` now have empty keep-lists plus `max_bytes` caps so generic gets stay bounded.

integrations/servicenow/markdown.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+	if rec.Updated != "" {
+		parts = append(parts, "Updated: "+rec.Updated)
+	}
+	if len(parts) > 0 {
+		b.Attribution(parts...)
+	}
+	for _, kv := range rec.Extra {
+		if kv[1] != "" {
+			b.Raw("**" + kv[0] + ":** " + kv[1] + "\n")
```
acmacalister: Description now goes through `markdown.FromHTML`, but these Extra fields (`close_notes`, `workaround`, `cause_notes`, `fix_notes`, plans, etc.) still land via `b.Raw`. ServiceNow often stores the same HTML there, so we can still leak tags into the rendered ticket.

Same treatment as description would cover it:

```go
for _, kv := range rec.Extra {
	if kv[1] != "" {
		b.Raw("**" + kv[0] + ":** ")
		b.WriteMarkdown(markdown.FromHTML(kv[1]))
		b.Raw("
")
	}
}
```

acmacalister (approved):
Nice follow-up — prior review items look fixed, and CI is green across build/test/lint/security/compose/rust-sdk. One leftover HTML path on Extra fields inline; everything else looks solid.

integrations/servicenow/markdown.go
daltoniam: Fixed: Extra fields (`close_notes`, workaround, plans, etc.) now go through `markdown.FromHTML` the same way descriptions do.

integrations/servicenow/markdown.go
```diff
@@ -0,0 +1,351 @@
...
+	if !ok {
+		return "", false
+	}
+	comments := make([]renderedComment, 0, len(list))
+	for _, m := range list {
+		comments = append(comments, renderedComment{
+			Author:  field(m, "sys_created_by"),
+			Created: field(m, "sys_created_on"),
+			Element: field(m, "element"),
+			Body:    stripHTML(field(m, "value")),
```
acmacalister: Descriptions and Extra fields now go through `markdown.FromHTML`, but comment bodies still use the hand-rolled `stripHTML`. Journal/work-note values can carry the same HTML ServiceNow puts in ticket fields, so we lose structure (lists, bold, links) and still risk odd leftovers.

Worth aligning this path:

```go
Body: string(markdown.FromHTML(field(m, "value"))),
```

(or drop `stripHTML` entirely if nothing else needs it).

integrations/servicenow/handlers.go
```diff
@@ -0,0 +1,460 @@
...
+		return mcp.ErrResult(err)
+	}
+	if err := validTable(table); err != nil {
+		return mcp.ErrResult(err)
+	}
+	body, err := parseDataJSON(dataStr)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	data, err := s.post(ctx, "/api/now/table/"+url.PathEscape(table), body)
```
acmacalister, line 318: `updateRecord` rejects an empty body, but `createRecord` will still POST `{}` when `data` is `{}` or only whitespace-empty fields. Same guard as the update path would keep the generic create tool from opening blank rows by accident:

```go
if len(body) == 0 {
    return mcp.ErrResult(fmt.Errorf("provide at least one field to create"))
}
```

acmacalister (approved):
Solid adapter — prior review items look fixed, and CI is green across build/test/lint/security/compose/rust-sdk. Two small leftovers inline on comment HTML and empty generic creates; nothing blocking.

integrations/servicenow/markdown.go
daltoniam: Fixed: comment bodies now go through `markdown.FromHTML`, and `stripHTML` is gone.

integrations/servicenow/handlers.go
daltoniam: Fixed: `createRecord` now rejects empty `{}` bodies the same way `updateRecord` does.

integrations/servicenow/handlers.go
```diff
@@ -0,0 +1,463 @@
...
+	mergeField(body, "urgency", urgency)
+	mergeField(body, "impact", impact)
+	mergeField(body, "priority", priority)
+	mergeField(body, "caller_id", callerID)
+	mergeField(body, "assignment_group", assignmentGroup)
+	mergeField(body, "assigned_to", assignedTo)
+	mergeField(body, "category", category)
+	mergeField(body, "subcategory", subcategory)
+	mergeField(body, "cmdb_ci", cmdbCI)
+	data, err := s.post(ctx, "/api/now/table/incident", body)
```
acmacalister, line 140: `createRecord` / `updateIncident` already reject empty bodies, but `createIncident` can still POST `{}` when `short_description` is `""` (or whitespace-only) and nothing else is set. Schema presence checks don't catch empty strings, so we end up creating a blank incident.

Same guard as the generic create path would close it:

```go
if len(body) == 0 {
    return mcp.ErrResult(fmt.Errorf("provide at least one field to create"))
}
```

Or, given the tool marks `short_description` required, require a non-empty value after the merges (including anything pulled in via `data`).

acmacalister (approved):
Solid adapter — prior review items look fixed, and CI is green across build/test/lint/security/compose/rust-sdk. One small leftover on empty typed creates inline; nothing blocking.

integrations/servicenow/handlers.go
daltoniam: Fixed: `createIncident` now rejects empty bodies after the field merges, same as the generic create path.

acmacalister (approved):
## Verification
| Check | Result |
|-------|--------|
| Build | Pass (CI) |
| Tests | Pass (CI) |
| Lint | Pass (CI) |
| Security | Pass (CI) |
| compose / rust-sdk | Pass (CI) |

Clean PR — CI is green across build/test/lint/security/compose/rust-sdk, and I didn't find any new blocking issues. Prior review feedback (arg errors, HTML rendering, empty creates/updates, generic compact caps) looks addressed. LGTM.

PR #231, opened 2026-09-10
Add FigJam planning through Figma's hosted MCP

aleksclark:
## Summary

- Add a focused Figma hosted-MCP integration so coding agents can inspect and update native, editable FigJam planning boards.
- Add guided Figma OAuth setup with provider-specific registration compatibility, configuration support, tests, and operator documentation.

## Test plan

- `make ci`
- Isolated live Switchboard + Crush smoke test: discovered all seven FigJam tools, authenticated with `figma_whoami`, and read node `2:57` from a real FigJam board without mutations.

💘 Generated with Crush


remotemcp/oauth.go
```diff
@@ -127,7 +133,11 @@ func StartOAuth(name, serverURL, redirectURI string) (string, error) {
 		return "", fmt.Errorf("remote server does not support dynamic client registration")
 	}
 
-	clientID, clientSecret, err := registerClient(meta.RegistrationEndpoint, redirectURI)
+	clientName := "Switchboard"
+	if len(scopes) > 0 && strings.TrimSpace(scopes[0]) == "mcp:connect" {
+		clientName = "Codex"
```
acmacalister: Registering as `Codex` when the scope is `mcp:connect` looks like an allowlist workaround rather than a real client identity. Figma's hosted MCP is documented as catalog-gated, and our runtime MCP client still announces itself as `switchboard` in `remotemcp.connect`, so OAuth identity and session identity already disagree.

That coupling is also pretty opaque: any future caller that needs `mcp:connect` silently inherits the Codex name. If we need this short-term for DCR to succeed, can we isolate it behind an explicit option (e.g. `ClientName: "Codex"`) with a short comment on why, and call out the ToS/breakage risk in the Figma setup docs? Long-term, getting Switchboard onto the catalog is much safer than impersonating another product.

web/web.go
```diff
@@ -1540,13 +1569,21 @@ func (w *WebServer) handleRemoteMCPOAuthStart(rw http.ResponseWriter, r *http.Re
...
+	}
 	if serverURL == "" {
 		json.NewEncoder(rw).Encode(map[string]string{"error": "Not a remote MCP integration"})
 		return
 	}
 
 	redirectURI := fmt.Sprintf("http://localhost:%d/api/remote/%s/oauth/callback", w.port, name)
-	authorizeURL, err := remotemcp.StartOAuth(name, serverURL, redirectURI)
+	scope := "read,write"
+	if name == "figma" {
```
acmacalister: The generic remote-MCP OAuth start path is starting to grow provider-specific branches (`figma` redirect + scope here, Linear `api_key` clearing in the callback). A root `GET /callback` also permanently claims a very generic path for one integration.

Have we considered a small per-integration OAuth profile (redirect path, scope, post-connect credential tweaks) instead of `if name == "figma"` / `if name == "linear"` in the shared handlers? That would keep the next remote MCP provider from needing another special case, and avoid baking Figma into a global `/callback` route.

remotemcp/remotemcp_test.go
```diff
@@ -284,6 +289,38 @@ func TestOAuth_RegisterClient_Failure(t *testing.T) {
...
+			registrationRedirect = redirects[0].(string)
+			w.WriteHeader(http.StatusCreated)
+			json.NewEncoder(w).Encode(map[string]string{"client_id": "figma-client"})
+		default:
+			http.NotFound(w, r)
+		}
+	}))
+	defer srv.Close()
+
+	authorizeURL, err := StartOAuth("figma-test", srv.URL, "http://localhost/callback", "mcp:connect")
```
acmacalister: This covers scope + `resource`, but not the side effect that `mcp:connect` forces `client_name=Codex` in `StartOAuth`. Worth asserting the registration body here too so that behavior can't drift from the intended Figma workaround without a failing test.

acmacalister (approved):
Nice focused FigJam slice on top of remotemcp — filtering, skill injection, and setup wiring look solid, and CI is green across build/test/lint/security/compose/rust-sdk. A few maintainability notes inline around the Figma OAuth compatibility hacks.

remotemcp/oauth.go
aleksclark: Addressed in 854d7e4: hosted OAuth now takes explicit OAuthOptions, and the Figma profile explicitly selects the compatibility client identity and resource. README documents the catalog/compatibility risk.

web/web.go
aleksclark: Addressed in 854d7e4: remote OAuth provider behavior now lives in remoteOAuthProfiles, including callback path, scope/options, credential destination, and post-connect clearing.

remotemcp/remotemcp_test.go
aleksclark: Addressed in 854d7e4: the OAuth test now asserts the explicit registered client name in addition to scope, resource, and redirect URI.

web/web.go
```diff
@@ -1540,13 +1590,27 @@ func (w *WebServer) handleRemoteMCPOAuthStart(rw http.ResponseWriter, r *http.Re
...
+	if !ok {
+		profile = remoteOAuthProfile{
+			CallbackPath:  "/api/remote/" + name + "/oauth/callback",
+			Options:       remotemcp.OAuthOptions{Scope: "read,write"},
+			CredentialKey: "mcp_access_token",
+		}
+	}
+	if name == "figma" {
+		profile.Options.Resource = serverURL + "/mcp"
+	}
```
acmacalister, line 1613: The profile map cleaned up most of the provider branching, but Resource still gets wired with a hard-coded `if name == "figma"` after the lookup. That leaves one Figma-only special case outside the profile.

Could we put the resource path on the profile itself (e.g. `ResourcePath: "/mcp"`) and derive `Options.Resource = serverURL + profile.ResourcePath` for any provider that needs RFC 8707 binding? Then the start handler stays generic and the next remote MCP client doesn't need another name check.

acmacalister (approved):
Solid FigJam slice — focused tool filter, skill injection, explicit OAuthOptions, and green CI across build/test/lint/security/compose/rust-sdk. One small leftover from the profile refactor inline; nothing blocking.

web/web.go
aleksclark: Addressed in c09bd55: resource binding is now declared as ResourcePath on the provider profile and derived generically from the server URL.

acmacalister (approved):
## Verification
| Check | Result |
|-------|--------|
| Build | Pass (CI) |
| Tests | Pass (CI) |
| Lint | Pass (CI) |
| Security | Pass (CI) |
| Compose / rust-sdk | Pass (CI) |

## What's Good
- Focused FigJam-only surface with clear tool descriptions and automatic  skill injection
- OAuth compatibility is explicit via  +  (scope, client name, resource path, credentials)
- Prior review feedback addressed cleanly; README documents the catalog/Codex compatibility risk
- Solid unit coverage for filtering, skill injection, OAuth options, and setup routes

## Must Fix (Blocking)
No issues found

## Should Fix (Non-Blocking)
No issues found

## Consider (Nice to Have)
No issues found

## Questions
None

Clean PR — CI is green across build/test/lint/security/compose/rust-sdk, prior review threads are resolved, and I didn't find any blocking issues. LGTM.

PR #232, opened 2026-09-10
Add LikeC4 Excalidraw MCP integration

aleksclark:
## Summary
- expose LikeC4 architecture inspection, validation, editing, and canvas screenshots through Switchboard's search and execute workflow
- extend remote MCP proxying for optional bearer authentication, argument-less calls, and native image/resource results

## Test plan
- `make ci`
- live smoke test against a locally built LikeC4 Excalidraw Streamable HTTP MCP server, covering architecture retrieval and PNG screenshots

💘 Generated with Crush


integrations/likec4excalidraw/likec4excalidraw.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	return i.remote.Execute(ctx, toolName, args)
+}
+
+func (i *integration) CompactSpec(toolName mcp.ToolName) ([]mcp.CompactField, bool) {
+	fields, ok := fieldCompactionSpecs[toolName]
+	return fields, ok
+}
+
+func (i *integration) MaxBytes(toolName mcp.ToolName) (int, bool) {
```
acmacalister, line 120: Screenshots are the headline feature here, but `get_canvas_screenshot` still goes through the default 50KB execute cap (`responseLimitFor` → `toolResultBytes`, which includes `Media`). Paperless raises this for image tools via `PerToolMaxResponseBytesIntegration` (5MB). Without something similar, a normal whole-canvas PNG will get rejected as oversized in production even though the adapter test with a tiny payload passes.

```go
const screenshotLimit = 5 * 1024 * 1024

func (i *integration) MaxResponseBytesForTool(toolName mcp.ToolName) (int, bool) {
	if toolName == "likec4excalidraw_get_canvas_screenshot" {
		return screenshotLimit, true
	}
	return 0, false
}
```

integrations/likec4excalidraw/likec4excalidraw.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	parsed.Path = strings.TrimSuffix(strings.TrimRight(parsed.Path, "/"), "/mcp")
+	parsed.RawQuery = ""
+	parsed.Fragment = ""
+	remote := remotemcp.NewOptionalToken(integrationName, strings.TrimRight(parsed.String(), "/"))
+	if err := remote.Configure(ctx, mcp.Credentials{"access_token": strings.TrimSpace(credentials["mcp_token"])}); err != nil {
+		return err
+	}
+	closeRemote(i.remote)
+	i.remote = remote
```
acmacalister, line 79: `Configure` swaps `i.remote` with no lock, and health refresh / setup handlers can call `Configure` while execute traffic is in flight. `slackmcp` serializes this kind of swap on purpose. A concurrent `Execute`/`Healthy` can observe a half-swapped or already-closed remote.

Worth a small `sync.Mutex` (or RWMutex) around the remote pointer read/write path, same shape as the other remote proxies.

integrations/likec4excalidraw/tools.go
```diff
@@ -0,0 +1,103 @@
...
+import mcp "github.com/daltoniam/switchboard"
+
+var tools = []mcp.ToolDefinition{
+	{
+		Name:        "likec4excalidraw_get_architecture",
+		Description: "Get the current LikeC4 architecture model, including elements, relationships, views, and diagnostics. Start here to inspect or navigate an architecture diagram before editing it.",
+		Parameters:  map[string]string{},
+	},
+	{
+		Name:        "likec4excalidraw_get_scene",
```
acmacalister, line 12: `get_scene` is a full Excalidraw document dump (elements, files, appState). With no compaction spec and the default 50KB cap, real scenes will either blow the limit or ship a huge unfiltered payload. Either add a tight whitelist / `max_bytes` in `compact.yaml`, or document that this tool is intentionally raw and raise/cap it deliberately.

integrations/likec4excalidraw/compact.yaml
```diff
@@ -0,0 +1,39 @@
+# LikeC4 Excalidraw response compaction specs.
+version: 1
+tools:
+  likec4excalidraw_get_architecture:
```
acmacalister, line 4: Other adapters with `compact.yaml` ship a `compact_specs_test.go` (parse losslessness, mutation exclusion, and especially shape parity against representative payloads). Right now we only assert "no orphan specs," so a path typo like `elements[].fqn` vs the real upstream shape would silently compact to `{}` in production. A small shape-parity table for architecture / diagnostics / validate would lock this in.

integrations/likec4excalidraw/likec4excalidraw.go
```diff
@@ -0,0 +1,110 @@
...
+	return map[string]string{"base_url": "http://127.0.0.1:4242"}
+}
+
+func (i *integration) Configure(ctx context.Context, credentials mcp.Credentials) error {
+	baseURL := strings.TrimSpace(credentials["base_url"])
+	if baseURL == "" {
+		return fmt.Errorf("likec4excalidraw: base_url is required")
+	}
+	parsed, err := url.Parse(baseURL)
+	if err != nil || parsed.Scheme == "" || parsed.Host == "" {
```
acmacalister: This accepts any scheme with a host (`ftp://…`, etc.). Paperless restricts to `http`/`https` and requires HTTPS off-loopback. Given the README guidance about keeping tokenless servers on localhost, matching that check here would make the "local only without a token" story enforceable instead of advisory.

acmacalister (commented):
Nice remote-MCP wrapper overall — optional bearer, nil-args fix, and image/resource passthrough are the right shared primitives. A few production issues around screenshot size limits and concurrent Configure worth tightening before merge.

CI: build, test, lint, security, rust-sdk, and compose all green on this head.

acmacalister (approved):
Clean PR — previous review items are addressed (screenshot response cap, Configure race safety, scene compaction + shape-parity tests, http/https + loopback checks). CI is green across build, test, lint, security, rust-sdk, and compose. LGTM.

PR #234, opened 2026-09-10
Add native OAuth login and refresh for WASM integrations

aleksclark:
## Summary
- Let WASM integrations use public-client authorization-code PKCE and native refresh after initial browser consent, without a companion or browser-dependent token renewal.
- Keep refresh tokens host-only and protect rotations through atomic config updates, same-origin mutation checks, reload/uninstall failure handling, and exact userinfo identity binding.

## Verification
- Full `make ci` passed, including race tests, lint, gosec and govulncheck.
- Focused OAuth/config/web/WASM race suites and MCP response-shape smoke passed.
- Real production Clerk public-client login completed for the expected user; native refresh exchanged and persisted a new access token without browser automation.
- Primer's deployed Tasks API requires a session ID and rejects delegated OAuth; the separately reviewed Tasks boundary is tracked at https://git.fleet.clark.team/aleksclark/primer/pulls/11. No live API acceptance is claimed yet.

Setup and recovery details: `docs/plugin-oauth.md`.

💘 Generated with Crush

### DIFF
--- a/pluginoauth/manager.go
+++ b/pluginoauth/manager.go
@@ -7,6 +7,8 @@ import (
 	"crypto/subtle"
 	"encoding/base64"
 	"errors"
+	"fmt"
+	"io"
 	"maps"
 	"net/http"
 	"net/url"
@@ -44,24 +46,25 @@ type authorization struct {
 }
 
 type Manager struct {
-	mu       sync.Mutex
-	name     string
-	store    Store
-	client   *http.Client
-	now      func() time.Time
-	creds    mcp.Credentials
-	metadata *metadata
-	pending  *authorization
-	dirty    bool
-	verified bool
-	blocked  bool
-	updates  mcp.Credentials
-	previous mcp.Credentials
-	baseline mcp.Credentials
+	mu         sync.Mutex
+	name       string
+	store      Store
+	client     *http.Client
+	randReader io.Reader
+	now        func() time.Time
+	creds      mcp.Credentials
+	metadata   *metadata
+	pending    *authorization
+	dirty      bool
+	verified   bool
+	blocked    bool
+	updates    mcp.Credentials
+	previous   mcp.Credentials
+	baseline   mcp.Credentials
 }
 
 func New(name string, store Store) *Manager {
-	return &Manager{name: name, store: store, now: time.Now, client: &http.Client{
+	return &Manager{name: name, store: store, now: time.Now, randReader: rand.Reader, client: &http.Client{
 		Timeout:       15 * time.Second,
 		CheckRedirect: func(*http.Request, []*http.Request) error { return http.ErrUseLastResponse },
 	}}
@@ -169,7 +172,14 @@ func (m *Manager) Start(ctx context.Context, creds mcp.Credentials, redirect, br
 		return "", err
 	}
 	creds = maps.Clone(creds)
-	state, verifier := randomValue(), randomValue()
+	state, err := randomValue(m.randReader)
+	if err != nil {
+		return "", fmt.Errorf("oauth: generate state: %w", err)
+	}
+	verifier, err := randomValue(m.randReader)
+	if err != nil {
+		return "", fmt.Errorf("oauth: generate PKCE verifier: %w", err)
+	}
 	hash := sha256.Sum256([]byte(verifier))
 	u, _ = url.Parse(md.AuthorizationEndpoint)
 	q := u.Query()
@@ -206,10 +216,12 @@ func (m *Manager) Start(ctx context.Context, creds mcp.Credentials, redirect, br
 	return u.String(), nil
 }
 
-func randomValue() string {
+func randomValue(reader io.Reader) (string, error) {
 	value := make([]byte, 32)
-	_, _ = rand.Read(value)
-	return base64.RawURLEncoding.EncodeToString(value)
+	if _, err := io.ReadFull(reader, value); err != nil {
+		return "", err
+	}
+	return base64.RawURLEncoding.EncodeToString(value), nil
 }
 
 func (m *Manager) Callback(ctx context.Context, code, state, browser, issuer string) error {
--- a/pluginoauth/manager.go
+++ b/pluginoauth/manager.go
@@ -1,0 +1,405 @@
+package pluginoauth
+
+import (
+	"context"
+	"crypto/rand"
+	"crypto/sha256"
+	"crypto/subtle"
+	"encoding/base64"
+	"errors"
+	"maps"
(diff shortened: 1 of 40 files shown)


pluginoauth/manager.go
```diff
@@ -0,0 +1,405 @@
...
+		m.previous, m.updates = nil, nil
+	}
+	clearTokens(creds)
+	m.pending = &authorization{baseline: baseline, updates: updates, state: state, browser: sha256.Sum256([]byte(browser)), verifier: verifier, redirect: redirect, expires: m.now().Add(10 * time.Minute), creds: maps.Clone(creds), metadata: *md}
+	return u.String(), nil
+}
+
+func randomValue() string {
+	value := make([]byte, 32)
+	_, _ = rand.Read(value)
```
acmacalister: If `rand.Read` fails here we still mint `state`/`code_verifier` from a zeroed buffer. Same pattern on the browser cookie in `web/plugin_oauth.go`. For OAuth entropy we should fail closed instead of proceeding:

```go
if _, err := rand.Read(value); err != nil {
    return "" // or bubble an error from randomValue
}
```

Worth making `randomValue() (string, error)` and treating failure as `ErrConfiguration` / start failure so we never ship a weak PKCE challenge or state.

web/templates/pages/integration.templ
```diff
@@ -108,6 +109,26 @@ templ IntegrationDetail(page layouts.PageData, data IntegrationDetailData) {
...
+		if data.ConnectOAuth {
+			<div class="card">
+				<button type="button" class="btn" data-oauth-start={ "/api/integrations/" + data.Name + "/oauth/start" } onclick="connectPluginOAuth(this)">Connect OAuth</button>
+				<p>Open this page using 127.0.0.1. OAuth tokens stay managed by Switchboard; connecting does not enable the integration.</p>
+			</div>
+			<script>
+				async function connectPluginOAuth(button) {
+					button.disabled = true;
+					try {
+						const response = await fetch(button.dataset.oauthStart, {method: "POST", credentials: "same-origin"});
```
acmacalister: This start call always posts an empty body, so it only ever uses already-saved credentials. The start API and docs both support a credentials overlay from the current form, and the form fields are right below this button.

If someone edits issuer/client/subject/email in the form and hits Connect without Save first, we silently authorize against the old saved settings. Collecting `cred_*` inputs (or requiring an explicit save and disabling Connect while dirty) would match the documented flow and avoid that footgun.

```javascript
const form = button.closest("form") || document.querySelector("form");
const credentials = {};
for (const input of form.querySelectorAll("[name^='cred_']")) {
  credentials[input.name.slice(5)] = input.value;
}
const response = await fetch(button.dataset.oauthStart, {
  method: "POST",
  credentials: "same-origin",
  headers: {"Content-Type": "application/json"},
  body: JSON.stringify({credentials}),
});
```

web/web.go
```diff
@@ -804,12 +829,27 @@ func (w *WebServer) handleUpdateCredentials(rw http.ResponseWriter, r *http.Requ
...
+		return
+	}
 	var creds mcp.Credentials
 	if err := json.NewDecoder(r.Body).Decode(&creds); err != nil {
 		writeJSON(rw, http.StatusBadRequest, map[string]string{"error": "invalid JSON: " + err.Error()})
 		return
 	}
 
+	if oauth {
+		if err := w.editPluginCredentials(r, name, creds, true); err != nil {
```
acmacalister: OAuth credential PUTs always pass `enabled=true`, so a hot-reload against a deliberately disabled plugin will re-enable it. The non-OAuth branch of this handler does the same, but for plugin OAuth the docs call out that connect must not enable a disabled integration, and form saves correctly thread the checkbox through `editPluginCredentials`.

If supervisors only need credential merge/refresh here, preserving the existing `Enabled` bit (or accepting it in the JSON body) would keep disabled plugins disabled:

```go
enabled := true
if existing, ok := w.services.Config.GetIntegration(name); ok && existing != nil {
    enabled = existing.Enabled
}
if err := w.editPluginCredentials(r, name, creds, enabled); err != nil {
```

acmacalister (approved):
Solid piece of work — PKCE, host-only refresh, identity binding, atomic config writes, and the concurrent rotation tests are thorough. CI is green across build/test/lint/security/rust-sdk/compose. A few non-blocking notes inline on entropy failure handling, the Connect button ignoring unsaved form fields, and PUT always re-enabling OAuth plugins.

pluginoauth/manager.go
aleksclark: Addressed in 006c30c29d0cfe13374f03707b6a6d577cdc23ad. Entropy failures now propagate from state/verifier generation and browser binding before issuance. Injected-reader regression tests cover each failure path. Focused race tests and make ci passed.

web/templates/pages/integration.templ
aleksclark: Addressed in 006c30c29d0cfe13374f03707b6a6d577cdc23ad. Connect now collects current cred_* fields from its associated configuration form and posts the credentials overlay. Rendered-template JavaScript regression exercises edited values and excludes unrelated forms. Focused race tests and make ci passed.

web/web.go
aleksclark: Addressed in 006c30c29d0cfe13374f03707b6a6d577cdc23ad. OAuth credential PUT now preserves the current Enabled value, including the CredentialEditor path. Regression tests cover disabled plugins and unchanged ordinary credential PUT behavior. Focused race tests and make ci passed.

acmacalister (approved):
Clean PR — CI is green across build/test/lint/security/rust-sdk/compose, and the earlier notes on entropy fail-closed, Connect form overlay, and preserving disabled OAuth plugins look addressed in 006c30c. Host-owned PKCE/refresh, atomic config writes, and the concurrent rotation coverage are in good shape. LGTM.

aleksclark:
Live follow-up: production Clerk public-client PKCE login completed for the pinned user. With managed Chromium fully stopped and the companion stopped, restarted Switchboard from persisted credentials after expiring only its isolated access-token cache. Switchboard performed the refresh grant itself, verified userinfo, and persisted a different access token. No browser/CDP or companion was involved. API acceptance is still separate: deployed Primer Tasks rejects OAuth as parent-session-required; Primer PR https://git.fleet.clark.team/aleksclark/primer/pulls/11 is merged at 9f5f5c3c0ce745146e07c3754646cf969fd15d59 and awaits approved image publication/deployment. Custom tasks:read/tasks:write scopes are registered, but the additional user consent has not been received.

PR #235, opened 2026-09-11
Add Front shared inbox integration

daltoniam:
## Summary
- Add a Front adapter so shared-inbox conversations, messages, comments, contacts, and outbound send/reply are available through search and execute.
- Support Bearer API/OAuth tokens with optional regional `base_url`, plus markdown rendering for conversation and message threads.

## Test plan
- [ ] `go test ./integrations/front ./config ./compact`
- [ ] Enable Front with `FRONT_ACCESS_TOKEN` and search for `front conversations`
- [ ] Execute `front_search_conversations` then `front_get_conversation` on a real inbox thread

💘 Generated with Crush


integrations/front/handlers.go
```diff
@@ -0,0 +1,435 @@
...
+)
+
+func resourcePath(id string) string {
+	return strings.ReplaceAll(id, "/", "%2F")
+}
+
+func listQuery(args map[string]any, extra map[string]string) map[string]string {
+	r := mcp.NewArgs(args)
+	limit := r.OptInt("limit", 25)
+	pageToken := r.Str("page_token")
```
acmacalister: `_pagination.next` from Front is a full URL (see the fixture in `compact_specs_test.go`), not a bare cursor. Tool copy says "Pagination cursor from `_pagination.next`", so models will often paste the whole URL straight into `page_token` and pagination will 400.

Ramp already handles this by accepting a full next URL and pulling query params out — same idea would work here:

```go
func normalizePageToken(raw string) string {
\tif raw == "" {
\t\treturn ""
\t}
\tif u, err := url.Parse(raw); err == nil && u.Scheme != "" {
\t\tif tok := u.Query().Get("page_token"); tok != "" {
\t\t\treturn tok
\t\t}
\t}
\treturn raw
}
```

Worth applying in `listQuery` / search so both a bare token and the full next URL work.

integrations/front/tools.go
```diff
@@ -0,0 +1,180 @@
...
+
+import mcp "github.com/daltoniam/switchboard"
+
+var tools = []mcp.ToolDefinition{
+	{
+		Name: mcp.ToolName("front_search_conversations"), Description: "Search Front inbox conversations, shared inbox threads, customer emails, and support tickets with Front search syntax. Start here for helpdesk triage, open conversations, unreplied mail, and finding threads by recipient, assignee, inbox, or tag.",
+		Parameters: map[string]string{
+			"query":      "Front search query (e.g. 'lost shipment', 'is:open is:unassigned', 'to:user@example.com', 'inbox:inb_41w25'). Text searches subject and body; filters use name:value.",
+			"limit":      "Results per page (default 25, max 100)",
+			"page_token": "Pagination cursor from _pagination.next",
```
acmacalister: This reads like `_pagination.next` is itself the cursor. Front returns a full next URL with `page_token` in the query string, so "pass `_pagination.next` as `page_token`" is the natural (and wrong) reading.

If we keep a bare-token input, something like "token from the `page_token` query param on `_pagination.next` (not the full URL)" would help. Prefer accepting the full next URL in the handler so either form works.

integrations/front/tools.go
```diff
@@ -0,0 +1,180 @@
...
+		Parameters: map[string]string{
+			"conversation_id": "Front conversation ID",
+			"channel_id":      "Optional sending channel ID. Defaults to the conversation channel.",
+			"to":              "Comma-separated recipient handles",
+			"cc":              "Comma-separated CC handles",
+			"bcc":             "Comma-separated BCC handles",
+			"subject":         "Email subject",
+			"body":            "Reply body",
+			"author_id":       "Optional teammate ID sending on behalf of",
+			"archive":         "Archive the conversation after sending (true/false)",
```
acmacalister: Front's create-reply API defaults `options.archive` to `true` when omitted. `create_message` calls that out; `reply_conversation` does not, and `outboundPayload` only sets `archive` when the arg is present — so a normal reply with no `archive` arg will archive the thread.

For agent use that is an easy footgun (triage reply quietly closes the conversation). Worth documenting the default here the same way as create, and considering an explicit `options.archive=false` when the arg is omitted so replies stay open unless asked.

integrations/front/handlers.go
```diff
@@ -0,0 +1,435 @@
...
+	if subject != "" {
+		payload["subject"] = subject
+	}
+	if authorID != "" {
+		payload["author_id"] = authorID
+	}
+	if senderName != "" {
+		payload["sender_name"] = senderName
+	}
+	options := map[string]any{}
```
acmacalister, line 352: When `archive` is omitted we leave `options` unset and Front archives by default. That matches the upstream API, but for send/reply tools driven by an LLM it is surprising.

If we want safer agent defaults:

```go
if archive != "" {
\toptions["archive"] = archive == "true" || archive == "1"
} else {
\toptions["archive"] = false // keep thread open unless asked
}
```

Either way, reply tool copy should match whatever we actually send.

acmacalister (approved):
Solid Front adapter — dispatch/compaction/markdown parity, shared helpers, and CI all look good. A couple of LLM footguns around pagination and reply archive defaults worth tightening before this sees heavy use.

CI: build, test, lint, security, rust-sdk, compose all green.

integrations/front/markdown.go
```diff
@@ -0,0 +1,227 @@
...
+			ID:        m.ID,
+			Type:      m.Type,
+			Inbound:   m.IsInbound,
+			CreatedAt: unixString(m.CreatedAt),
+			Author:    personLabel(m.Author.FirstName, m.Author.LastName, m.Author.Username, m.Author.Email),
+			Subject:   m.Subject,
+			Body:      markdown.Markdown(body),
+		})
+	}
+	return messagesToMarkdown(msgs), true
```
acmacalister: `RenderMarkdown` replaces the whole JSON payload, so `_pagination.next` never reaches the model. Front message timelines are cursor-paginated (default 25, max 100), and long threads will often need a second page — but the markdown only shows `Messages (N)` with no next cursor, so agents tend to stop as if the thread is complete.

Worth parsing `_pagination` on the list response and appending something like `Next page_token: …` (or the full next URL, since handlers already accept either) when it is non-null, same idea as keeping `_pagination.next` in the compact spec.

acmacalister (approved):
Nice Front adapter — dispatch/compaction/markdown parity, the page_token URL normalization, and the safer reply archive default all look solid. CI is green across build/test/lint/security/rust-sdk/compose. One remaining LLM footgun on message markdown pagination inline.

integrations/front/markdown.go
```diff
@@ -0,0 +1,240 @@
...
+	Type      string
+	Inbound   bool
+	CreatedAt string
+	Author    string
+	Subject   string
+	Body      markdown.Markdown
+}
+
+var markdownRenderers = map[mcp.ToolName]func([]byte) (markdown.Markdown, bool){
+	"front_get_conversation":           renderConversationMD,
```
acmacalister: `front_get_conversation` is almost pure structured metadata (status, assignee, tags, ticket IDs) with no document body. Markdown rendering skips compaction, so those chaining fields never reach the model — agents only get subject/status/name strings.

That sits on the wrong side of the markdown decision framework (`docs/markdown-rendering.md`: metadata/status tools stay JSON). Message threads are a good markdown fit; this one is closer to `github_get_pull`.

Two options that both work:
1. Drop `front_get_conversation` from `markdownRenderers` and let the compact spec handle it (preferred — matches the tool description promising ticket IDs / tags).
2. Keep markdown but surface `assignee.id`, tag id/name, and `ticket_ids` in the attribution so follow-up tools still have IDs to use.

integrations/front/handlers.go
```diff
@@ -0,0 +1,447 @@
...
+	pinned := r.Str("is_pinned")
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	payload := map[string]any{"body": body}
+	if authorID != "" {
+		payload["author_id"] = authorID
+	}
+	if pinned != "" {
+		payload["is_pinned"] = pinned == "true"
```
acmacalister: `is_pinned` only treats the string `"true"` as pinned. Elsewhere we already accept bool/`"1"` via `ArgStr` + broader checks (see `archive` just below). Worth using `r.Bool("is_pinned")` with a presence check, or at least the same `== "true" || == "1"` path, so MCP bool args and `1` behave consistently.

acmacalister (approved):
Nice Front adapter — dispatch/compaction/markdown parity, page_token URL normalization, and the safer reply archive default all landed cleanly. CI is green across build/test/lint/security/rust-sdk/compose. Two non-blocking notes inline on get_conversation markdown and is_pinned parsing.

integrations/front/handlers.go
```diff
@@ -0,0 +1,447 @@
...
+
+func outboundPayload(r *mcp.Args, defaultArchive bool) (map[string]any, error) {
+	to := r.Str("to")
+	cc := r.Str("cc")
+	bcc := r.Str("bcc")
+	subject := r.Str("subject")
+	body := r.Str("body")
+	authorID := r.Str("author_id")
+	senderName := r.Str("sender_name")
+	archive := r.Str("archive")
```
acmacalister: `is_pinned` now goes through `r.Bool`, but `archive` still uses `r.Str` and only treats `"true"` / `"1"` as true. A string like `"TRUE"` (or anything `ParseBool` would accept) ends up as `archive=false` here.

Same presence-check pattern as add_comment would keep create/reply consistent:

```go
if _, ok := r /* or args */["archive"]; ok {
	options["archive"] = r.Bool("archive")
} else {
	options["archive"] = defaultArchive
}
```

Or pull `archive` with `r.Bool` after checking `args` for the key, matching `addComment`.

docs/adapter-reference.md
```diff
@@ -110,6 +110,7 @@ Each adapter uses either a typed SDK or raw HTTP. Auth varies:
...
 - **HubSpot**: Hand-rolled REST over `net/http` against CRM v3 (`https://api.hubapi.com`). Auth via `Authorization: Bearer <access_token>` (private app token). ~31 tools covering contacts, companies, deals, tickets, generic objects, associations, owners, pipelines, and properties. Cursor pagination via `after`.
+- **Front**: Hand-rolled REST over `net/http` against Core API (`https://api2.frontapp.com`). Auth via `Authorization: Bearer <access_token>` (API token or OAuth). ~20 tools covering conversation search/list/get, messages, comments, assignment/status, inboxes, teammates, tags, channels, contacts, accounts, drafts, and sending. Cursor pagination via `page_token` / `_pagination.next`. Markdown rendering for conversations and message threads.
```
acmacalister: Small leftover from dropping conversation markdown: this still says "Markdown rendering for conversations and message threads", but only `front_list_conversation_messages` renders markdown now. Something like "Markdown rendering for message threads" would match the code.

acmacalister (approved):
Solid Front adapter — prior pagination/archive/markdown feedback is addressed, dispatch/compaction/parity tests look complete, and CI is green (build, test, lint, security, compose). Two small leftover nits inline.

integrations/front/markdown.go
```diff
@@ -0,0 +1,170 @@
...
+		body := m.Text
+		if body == "" {
+			body = string(markdown.FromHTML(m.Body))
+		}
+		msgs = append(msgs, renderedMessage{
+			ID:        m.ID,
+			Type:      m.Type,
+			Inbound:   m.IsInbound,
+			CreatedAt: unixString(m.CreatedAt),
+			Author:    personLabel(m.Author.FirstName, m.Author.LastName, m.Author.Username, m.Author.Email),
```
acmacalister: Front's `author` is the teammate who sent the message — inbound customer mail usually has `author: null` and puts the sender in `recipients` with `role: "from"` (see [Get message](https://dev.frontapp.com/reference/get-message)). As written, those threads render as `unknown`, which is rough for the main triage path.

Worth falling back to the `from` recipient when author fields are empty:

```go
type rawRecipient struct {
	Name   string `json:"name"`
	Handle string `json:"handle"`
	Role   string `json:"role"`
}
// on rawMessage:
Recipients []rawRecipient `json:"recipients"`

author := personLabel(m.Author.FirstName, m.Author.LastName, m.Author.Username, m.Author.Email)
if author == "" {
	for _, r := range m.Recipients {
		if r.Role == "from" {
			author = personLabel(r.Name, "", "", r.Handle)
			break
		}
	}
}
```

acmacalister (approved):
Solid Front adapter — prior pagination/archive/bool/markdown feedback is addressed, dispatch/compaction/parity coverage looks complete, and CI is green (build, test, lint, security, rust-sdk, compose). One remaining LLM footgun on inbound message attribution.

integrations/front/markdown.go
```diff
@@ -0,0 +1,192 @@
...
+	Inbound   bool
+	CreatedAt string
+	Author    string
+	Subject   string
+	Body      markdown.Markdown
+}
+
+var markdownRenderers = map[mcp.ToolName]func([]byte) (markdown.Markdown, bool){
+	"front_list_conversation_messages": renderMessagesMD,
+}
```
acmacalister, line 26: `front_list_conversation_comments` is still JSON-only while messages get markdown. Internal notes are the same kind of read-the-thread content the markdown docs call out (and Zendesk/Jira/Confluence/ServiceNow/Forgejo all render comment lists).

For triage, agents often need both the customer thread and teammate notes in the same readable form. Worth adding a comments renderer next to `renderMessagesMD` (body + author + posted_at/is_pinned, plus next-page if present) so the two conversation timeline tools stay consistent.

acmacalister (approved):
Solid Front adapter — prior pagination/archive/bool/markdown feedback is addressed, dispatch/compaction/parity coverage looks complete, and CI is green (build, test, lint, security, rust-sdk, compose). One remaining LLM-usability note on comment markdown inline.

integrations/front/handlers.go
```diff
@@ -0,0 +1,447 @@
...
+func normalizePageToken(raw string) string {
+	if raw == "" {
+		return ""
+	}
+	if u, err := url.Parse(raw); err == nil && u.Scheme != "" {
+		if tok := u.Query().Get("page_token"); tok != "" {
+			return tok
+		}
+	}
+	return raw
```
acmacalister: `normalizePageToken` only pulls `page_token` out of a full `_pagination.next` URL, then `listQuery` always rewrites `limit` from args (default 25). Front's pagination docs say the next URL already carries `limit` and should be used as-is, and Ramp does the stronger version of this by adopting *all* query params from `next`.

So when a model pastes the full next URL (which tool copy now encourages) after a `limit=100` page, we quietly drop back to 25 and can desync the cursor. Worth either requesting the next URL path+query wholesale, or seeding params from the parsed URL before applying explicit overrides:

```go
func pageParamsFromNext(raw string) map[string]string {
	if raw == "" {
		return nil
	}
	u, err := url.Parse(raw)
	if err != nil || u.Scheme == "" {
		return map[string]string{"page_token": raw}
	}
	out := map[string]string{}
	for k, vs := range u.Query() {
		if len(vs) > 0 && vs[0] != "" {
			out[k] = vs[0]
		}
	}
	return out
}
```

integrations/front/markdown.go
```diff
@@ -0,0 +1,269 @@
...
+		if body == "" {
+			body = string(markdown.FromHTML(m.Body))
+		}
+		msgs = append(msgs, renderedMessage{
+			ID:        m.ID,
+			Type:      m.Type,
+			Inbound:   m.IsInbound,
+			CreatedAt: unixString(m.CreatedAt),
+			Author:    messageAuthor(m),
+			Subject:   m.Subject,
```
acmacalister, line 83: `Subject` (and `ID`) get copied onto `renderedMessage` but `messagesToMarkdown` never writes them, so email threads lose subject lines once markdown replaces the JSON. Easy fix is to fold a non-empty subject into the attribution context, e.g. `email, inbound, Help, 2022-...`, so the model still sees what each message was about.

acmacalister (approved):
Solid Front adapter — prior pagination/archive/bool/markdown feedback is addressed, dispatch/compaction/markdown parity looks complete, and CI is green (build, test, lint, security, rust-sdk, compose). Two non-blocking notes inline on next-URL param preservation and message subject in markdown.

acmacalister (approved):
Clean Front adapter — prior pagination/archive/bool/markdown feedback is all landed, dispatch/compaction/markdown parity looks complete, and CI is green (build, test, lint, security, rust-sdk, compose). LGTM.

PR #236, opened 2026-09-11
Post Crush review as a waitable PR check

daltoniam:
## Summary
- Crush review currently runs after CI via `workflow_run`, so GitHub reports the job against `main` instead of the PR head. That makes `gh pr checks` and required status checks unable to wait for it.
- Post an in-progress check run on the PR head SHA, then complete it when Crush finishes, so the review is waitable on the pull request.

## Test plan
- [ ] Open a follow-up PR after this lands and wait for CI
- [ ] Confirm `Crush PR Review` appears on the PR checks list against the head SHA
- [ ] Confirm `gh pr checks <n>` stays pending until Crush finishes, then reports success or failure

After merge, add `Crush PR Review` as a required status check if we want it to gate merge.

💘 Generated with Crush

### DIFF
--- a/.github/workflows/crush-pr-review.yml
+++ b/.github/workflows/crush-pr-review.yml
@@ -5,9 +5,13 @@ on:
     workflows: ["CI"]
     types: [completed]
 
+# workflow_run jobs report against the default branch, so they never appear
+# on the PR. Post a check run on the PR head SHA so gh pr checks / required
+# status checks can wait on Crush review.
 permissions:
   contents: read
   pull-requests: write
+  checks: write
 
 jobs:
   crush-review:
@@ -26,6 +30,22 @@ jobs:
         run: |
           PR_NUMBER=$(gh api repos/${{ github.repository }}/commits/${{ github.event.workflow_run.head_sha }}/pulls --jq '.[0].number')
           echo "number=$PR_NUMBER" >> "$GITHUB_OUTPUT"
+
+      - name: Start Crush review check
+        id: check
+        env:
+          GH_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
+        run: |
+          payload=$(jq -n \
+            --arg name "Crush PR Review" \
+            --arg sha "${{ github.event.workflow_run.head_sha }}" \
+            --arg started "$(date -u +%Y-%m-%dT%H:%M:%SZ)" \
+            '{name: $name, head_sha: $sha, status: "in_progress", started_at: $started}')
+          CHECK_ID=$(gh api repos/${{ github.repository }}/check-runs \
+            --method POST \
+            --input - \
+            --jq '.id' <<<"$payload")
+          echo "id=$CHECK_ID" >> "$GITHUB_OUTPUT"
 
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/crush-pr-review.yml
```diff
@@ -27,6 +31,22 @@ jobs:
           PR_NUMBER=$(gh api repos/${{ github.repository }}/commits/${{ github.event.workflow_run.head_sha }}/pulls --jq '.[0].number')
           echo "number=$PR_NUMBER" >> "$GITHUB_OUTPUT"
 
+      - name: Start Crush review check
+        id: check
+        env:
+          GH_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
```
acmacalister: This step posts check runs with `secrets.GH_ACCESS_TOKEN`, but the new `permissions: checks: write` only applies to `GITHUB_TOKEN`. Check Runs write also isn't available to classic PATs — only GitHub Apps and fine-grained tokens with Checks: write ([docs](https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#create-a-check-run)).

If `GH_ACCESS_TOKEN` is a classic PAT (or a fine-grained token without Checks: write), this POST 403s and the whole Crush job dies before review runs.

Safer split: create/update the check run with `github.token` so the workflow `checks: write` permission is what actually authorizes it, and keep `GH_ACCESS_TOKEN` for PR comments only.

```yaml
env:
  GH_TOKEN: ${{ github.token }}
```

.github/workflows/crush-pr-review.yml
```diff
@@ -27,6 +31,22 @@ jobs:
           PR_NUMBER=$(gh api repos/${{ github.repository }}/commits/${{ github.event.workflow_run.head_sha }}/pulls --jq '.[0].number')
           echo "number=$PR_NUMBER" >> "$GITHUB_OUTPUT"
 
+      - name: Start Crush review check
```
acmacalister, line 34: Because this sits before Crush runs with no `continue-on-error`, any check-run create failure (auth, transient API blip) skips the review entirely. That couples "make the check waitable" to "post the review," and is a regression vs today.

Worth `continue-on-error: true` here so a failed check create still lets Crush review, and rely on `if: always() && steps.check.outputs.id != ''` on the complete step (which already no-ops when `id` is empty).

.github/workflows/crush-pr-review.yml
```diff
@@ -27,6 +31,22 @@ jobs:
...
+        id: check
+        env:
+          GH_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
+        run: |
+          payload=$(jq -n \
+            --arg name "Crush PR Review" \
+            --arg sha "${{ github.event.workflow_run.head_sha }}" \
+            --arg started "$(date -u +%Y-%m-%dT%H:%M:%SZ)" \
+            '{name: $name, head_sha: $sha, status: "in_progress", started_at: $started}')
```
acmacalister: Minor: a `details_url` pointing at this workflow run makes the PR Checks UI one click from the Crush logs instead of a bare status.

```jq
--arg url "${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
'{name: $name, head_sha: $sha, status: "in_progress", started_at: $started, details_url: $url}'
```

acmacalister (approved):
Nice fix for the workflow_run-vs-PR-check blind spot — in_progress → complete on the head SHA is the right shape. CI is green (build/test/lint/security/rust-sdk/compose). A couple of token/resilience notes inline so the new check doesn't accidentally gate or break the review itself.

acmacalister (approved):
Clean PR — CI is green (build/test/lint/security/rust-sdk/compose) and the earlier token/resilience notes are already addressed (`github.token` + `checks: write`, `continue-on-error`, `details_url`). Waitable check-run on the PR head SHA is the right shape. LGTM.

PR #237, opened 2026-09-14
Add Forgejo Actions run, job, and log tools

aleksclark:
## Summary
- Expose Forgejo 16 Actions CI runs, jobs, and plaintext job logs so operators can inspect pipeline status without leaving Switchboard.
- Keep follow-ups on `run_id`/`job_id` from list tools, with a 1 MiB budget for job logs.

## Test plan
- [x] `make ci`
- [x] Isolated live smoke: search `CI run`, then execute list runs/jobs and get logs against `aleksclark/fleet-iac`

💘 Generated with Crush


integrations/forgejo/actions.go
```diff
@@ -0,0 +1,173 @@
...
+	if workflowID != "" || ref != "" {
+		next := make([]*sdk.ActionRun, 0, len(filtered))
+		for _, run := range filtered {
+			if run == nil {
+				continue
+			}
+			if workflowID != "" && run.WorkflowID != workflowID {
+				continue
+			}
+			if ref != "" && run.PrettyRef != ref {
```
acmacalister: Local `ref` filtering compares against `run.PrettyRef`, but Forgejo's `prettyref` is a display form (`main`, or `#12` for PRs), not the full Git ref. The tool docs tell callers to pass values like `refs/heads/main`, so this filter will silently drop every matching run.

Forgejo's list-runs endpoint already accepts `ref` and `workflow_id` server-side (exact match on the stored `ref` column). The SDK's `ListActionRunsOption` just doesn't expose them yet. Prefer forwarding those query params (raw HTTP or a tiny options extension) over page-local filtering on `PrettyRef` — local filtering also only sees the current page, so a busy repo can return an empty page even when matching runs exist later.

```go
// Prefer server-side filters; PrettyRef is display-only.
if ref != "" {
    // query.Set("ref", ref) on GET /actions/runs
}
```

acmacalister (commented):
Solid Forgejo Actions surface — dispatch/compaction/search parity and the raw jobs/logs path all look careful. CI is green (build, test, lint, security, rust-sdk, compose). A couple of filter/docs issues around `ref`/`workflow_id` inline; the rest looks ready once those are sorted.

integrations/forgejo/actions.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	runs, resp, err := c.ListRepoActionRuns(owner, repo, opts)
+	if err != nil || (resp != nil && resp.Response != nil && resp.StatusCode >= 300) {
+		return finish(nil, resp, err)
+	}
+	if runs == nil {
+		return finish([]*sdk.ActionRun{}, resp, nil)
+	}
+	filtered := runs.WorkflowRuns
+	if workflowID != "" || ref != "" {
```
acmacalister: Same concern for `workflow_id`: filtering only the current page after `ListRepoActionRuns` means `per_page` no longer means "up to N matching runs". On a repo with mixed workflows, page 1 can easily come back empty even though later pages have hits.

Since Forgejo 16's list endpoint already filters by `workflow_id` (and `ref`) in `FindRunOptions`, wiring those through avoids the pagination footgun and matches what the tool description implies.

docs/adapter-reference.md
```diff
@@ -77,12 +77,12 @@ Dynamic adapters without a static dispatch map (e.g. `slackmcp`) must instead pr
...
+  - **Search verification**: discovery reads local tool metadata, not the Forgejo API. Disabled tools are hidden unless `discoverAll` is enabled, when they are marked unconfigured. Short queries such as `Forgejo repo`, `Forgejo issue`, and `Forgejo file content` cover common entry points. `Forgejo pull review` is intentionally broad: finding PRs, inspecting diffs, and submitting reviews are useful results. For existing reviews use `{"integration":"forgejo","query":"list pull reviews"}` rather than tuning ranking to force the review list above other workflow tools. `Forgejo CI run` and `Forgejo job logs` cover Actions run/job/log tools. `server/forgejo_test.go` checks the real catalog, top-3/top-5 discovery, stable ordering, and disabled visibility without upstream HTTP.
```
acmacalister, line 85: This line says Forgejo 16 does not accept `workflow_id`/`ref` on the list endpoint, but current Forgejo does — `ListActionRuns` passes both into `FindRunOptions` (`Ref`, `WorkflowID`). Worth correcting so future readers do not reintroduce client-side filtering for an API that already supports it.

(The gap is in the v3 SDK's `ListActionRunsOption`, not the server.)

integrations/forgejo/actions.go
aleksclark: Addressed on 735b9ce47867f86420d7abe134e944d98804b931. `list_action_runs` now forwards `ref` and `workflow_id` as query params to `GET /repos/{owner}/{repo}/actions/runs` instead of comparing against `prettyref` on the current page.

integrations/forgejo/actions.go
aleksclark: Addressed on 735b9ce47867f86420d7abe134e944d98804b931. `workflow_id` is now a server-side list-runs filter, so `per_page` is matching runs rather than a locally sliced page.

docs/adapter-reference.md
aleksclark: Addressed on 735b9ce47867f86420d7abe134e944d98804b931. Adapter-reference now documents that Forgejo 16 accepts `workflow_id`/`ref` and that the v3 SDK is the gap, not the server.

acmacalister (approved):
Clean PR — CI is green (build, test, lint, security, rust-sdk, compose), prior filter/docs feedback is resolved on 735b9ce, and the Actions surface looks solid end-to-end (dispatch/compaction/search parity, server-side run filters, 1 MiB log budget). LGTM.

PR #238, opened 2026-09-15
Add Grist spreadsheet and relational-document integration

aleksclark:
## Summary
- Give agents one Switchboard path to discover Grist orgs, workspaces, and workbooks, then read and write tables, columns, and rows without a separate Grist client.
- Cover the REST surface LLMs need for spreadsheet-style work: records CRUD/upsert, SQL SELECT, webhooks, and attachments, with compaction on read tools.

## Test plan
- [x] `make ci`
- [x] Isolated MCP search discovers `grist_list_orgs`
- [ ] Live `grist_list_orgs` against a Grist instance that is not behind extra HTTP Basic

💘 Generated with Crush


integrations/grist/handlers.go
```diff
@@ -0,0 +1,733 @@
...
+	"strings"
+
+	mcp "github.com/daltoniam/switchboard"
+)
+
+func pathEscape(id string) string {
+	return url.PathEscape(id)
+}
+
+func jsonArg(raw string, dest any, name string) error {
```
acmacalister: `jsonArg` only accepts a pre-stringified JSON blob, and every write path feeds it via `r.Str(...)`. That works for MCP `execute` when the model double-encodes, but scripts (and any client that passes real arrays/objects) hit `cannot convert []interface {} to string` before we ever talk to Grist — which is the natural shape for bulk `records` / `tables` / `columns` / `webhooks` / SQL `args`.

Ramp and gchat already take either form. Something like:

```go
func anyJSONArg(args map[string]any, key string) (any, error) {
	v, ok := args[key]
	if !ok || v == nil {
		return nil, fmt.Errorf("%s is required", key)
	}
	switch v := v.(type) {
	case string:
		var out any
		if err := json.Unmarshal([]byte(v), &out); err != nil {
			return nil, fmt.Errorf("invalid JSON for %s: %w", key, err)
		}
		return out, nil
	default:
		return v, nil
	}
}
```

Worth covering both call shapes in a test so `api.call('grist_add_records', {records: [{fields: {...}}]})` stays green.

integrations/grist/handlers.go
```diff
@@ -0,0 +1,733 @@
...
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	if docID == "" {
+		return mcp.ErrResult(fmt.Errorf("doc_id is required"))
+	}
+	if tableID == "" {
+		return mcp.ErrResult(fmt.Errorf("table_id is required"))
+	}
+	if _, ok := args["limit"]; !ok {
```
acmacalister: Nice that this avoids `OptInt` so an explicit `limit: 0` still means "no limit" per Grist. One thing that bites similar list tools: there is no local upper bound before we hit the wire, so a large table + `limit: 0` (or a huge limit) can still pull a multi‑MB payload before `max_bytes` trims the compacted result. A soft cap (or refusing `0` unless the caller opts in) would keep agents from accidentally dumping whole workbooks into context.

acmacalister (approved):
Nice Grist adapter — clean hexagonal shape, dispatch/compaction parity, careful limit defaulting, and solid HTTP helpers. One structured-arg issue on the write path worth fixing for scripts; small note on unbounded list pulls. CI is green (build/test/lint/security/compose/rust-sdk).

integrations/grist/handlers.go
aleksclark: Fixed on a6bb3466dc85220cc62c4f58b883995287f0014e.

`anyJSONArg` now accepts native arrays/objects or a JSON string. `TestAddRecords_NativeArray` covers `records: [{fields: {...}}]`.

integrations/grist/handlers.go
aleksclark: Fixed on a6bb3466dc85220cc62c4f58b883995287f0014e.

List tools now default to 50, clamp to 500, and require `unlimited=true` for `limit: 0` or anything above 500. Tests: `TestListRecords_RejectsZeroWithoutUnlimited`, `TestListRecords_ZeroWithUnlimited`, `TestListRecords_ClampsHugeLimit`.

integrations/grist/handlers.go
```diff
@@ -0,0 +1,771 @@
...
+		return mcp.ErrResult(err)
+	}
+	return mcp.RawResult(data)
+}
+
+func listRecords(ctx context.Context, g *grist, args map[string]any) (*mcp.ToolResult, error) {
+	r := mcp.NewArgs(args)
+	docID := r.Str("doc_id")
+	tableID := r.Str("table_id")
+	filter := r.Str("filter")
```
acmacalister: `filter` still goes through `r.Str`, so the same script shape we just fixed for write args fails here:

```js
api.call('grist_list_records', {
  doc_id: '...',
  table_id: 'People',
  filter: { Status: ['Open'] },
})
```

hits `cannot convert map[string]interface {} to string` before the request goes out. `list_attachments` has the same path.

Worth routing through something like `optionalJSONArg` and JSON-encoding when it is not already a string, since Grist wants the filter as a query-string JSON blob either way.

integrations/grist/handlers.go
```diff
@@ -0,0 +1,771 @@
...
+		return mcp.ErrResult(err)
+	}
+	return mcp.RawResult(data)
+}
+
+func deleteRecords(ctx context.Context, g *grist, args map[string]any) (*mcp.ToolResult, error) {
+	r := mcp.NewArgs(args)
+	docID := r.Str("doc_id")
+	tableID := r.Str("table_id")
+	recordIDs := r.Str("record_ids")
```
acmacalister: Same class of issue on `record_ids`: scripts naturally pass `[1, 2, 3]`, but `r.Str` + `parseIntIDs` only accept a comma-separated string.

Accepting a native int array (or a JSON array string) alongside the CSV form would keep deletes consistent with the bulk write tools.

acmacalister (approved):
Solid follow-up on the native JSON write args and list-limit caps — those fixes look right, and the adapter shape (dispatch/compaction parity, HTTP helpers, Start here entry point) is in good shape.

CI is green across build/test/lint/security/compose/rust-sdk. Two leftover script-arg spots on `filter` and `record_ids` inline; neither blocks merge.

integrations/grist/handlers.go
aleksclark: Fixed on 042d447b838c4048fd43a5c2deef3832b68c1834.

`filterQuery` now accepts a native object or a JSON string and encodes it for the query. `TestListRecords_NativeFilter` covers `{Status:['Open']}`.

integrations/grist/handlers.go
aleksclark: Fixed on 042d447b838c4048fd43a5c2deef3832b68c1834.

`recordIDsArg` now accepts CSV, a JSON array string, or a native `[1,2,3]`. `TestDeleteRecords_NativeIDs` covers the array form.

integrations/grist/handlers.go
```diff
@@ -0,0 +1,870 @@
...
+}
+
+func recordIDsArg(args map[string]any) ([]int, error) {
+	v, ok := args["record_ids"]
+	if !ok || v == nil {
+		return nil, fmt.Errorf("record_ids is required")
+	}
+	switch v := v.(type) {
+	case string:
+		return parseIntIDs(v)
```
acmacalister: The string branch always goes through `parseIntIDs`, so a JSON array string like `"[1,2,3]"` (the same shape models use for `records` / `args`) fails on `[1` instead of unmarshalling.

The `default` arm below looks like it was meant to cover that, but strings never reach it. Something like trying JSON first when the string starts with `[` keeps CSV working and matches the native-array path:

```go
case string:
	s = strings.TrimSpace(v)
	if strings.HasPrefix(s, "[") {
		var items []any
		if err := json.Unmarshal([]byte(s), &items); err != nil {
			return nil, fmt.Errorf("invalid JSON for record_ids: %w", err)
		}
		// convert items via intFromAny...
	}
	return parseIntIDs(s)
```

Worth a quick test for `record_ids: "[1,2,3]"` so we don't regress the double-encoded form.

integrations/grist/handlers.go
```diff
@@ -0,0 +1,870 @@
...
+	if _, ok := args["limit"]; !ok {
+		return defaultLimit, nil
+	}
+	if limit == 0 {
+		if !unlimited {
+			return 0, fmt.Errorf("limit 0 fetches the whole table; pass unlimited=true to confirm")
+		}
+		return 0, nil
+	}
+	if limit > maxLimit && !unlimited {
```
acmacalister, line 72: `limit < 0` slips past both the zero-check and the max clamp, so we send a negative `limit` query param straight to Grist. Cheap to reject (or treat as missing and fall back to the default 50) before the request goes out.

acmacalister (approved):
Solid Grist adapter — hexagonal shape, dispatch/compaction parity, native JSON write args, and the list-limit caps all look right. CI is green (build/test/lint/security/compose/rust-sdk). Two small leftover spots on `record_ids` string parsing and negative limits inline; neither blocks merge.

integrations/grist/handlers.go
aleksclark: Fixed on b655f35c431bd7cd88f067ca39398f9492cb3b85.

String values starting with `[` now unmarshal as a JSON array before falling back to CSV. `TestDeleteRecords_JSONArrayString` covers `record_ids: "[1,2,3]"`.

integrations/grist/handlers.go
aleksclark: Fixed on b655f35c431bd7cd88f067ca39398f9492cb3b85.

Negative limits now fall back to the default 50. `TestListRecords_NegativeLimitDefaults` covers `limit: -1`.

acmacalister (approved):
Clean PR — CI is green across build/test/lint/security/compose/rust-sdk, and I didn't find any blocking issues after the native JSON args, filter/record_ids, and list-limit follow-ups. LGTM.

## Verification
| Check | Result |
|-------|--------|
| Build | Pass (CI) |
| Tests | Pass (CI) |
| Lint | Pass (CI) |
| Security | Pass (CI) |
| Compose / rust-sdk | Pass (CI) |

## What's Good
- Hexagonal adapter shape matches project conventions (unexported struct, exported `New`, dispatch map, shared `mcp` helpers).
- Dispatch + compaction parity tests, Start-here entry point, and limit defaults/clamps are in good shape.
- Prior review items (native JSON write args, filter objects, record id arrays/JSON strings, negative limits) look correctly addressed with regression coverage.

PR #239, opened 2026-09-16
Add native GitLab REST integration

daltoniam:
## Summary

Replaces the earlier MCP-proxy approach with a **native GitLab REST API v4** adapter (Forgejo-shaped: `base_url` + personal access token). No `remotemcp`, no official `/api/v4/mcp` proxy, no Hosted OAuth platform changes.

### Tools (scoped, not 200+)

| Area | Tools |
|------|--------|
| Projects | `gitlab_list_projects` (Start here), `gitlab_get_project`, `gitlab_get_current_user` |
| Merge requests | `gitlab_list_merge_requests` (Start here for MRs), `gitlab_get_merge_request`, `gitlab_create_merge_request_note` |
| Issues | `gitlab_list_issues`, `gitlab_get_issue` |
| CI/CD | `gitlab_list_pipelines`, `gitlab_list_pipeline_jobs`, `gitlab_get_job_trace` |

Read tools have `compact.yaml` specs; job trace is capped at 1 MiB response.

### Config

```json
{
  "integrations": {
    "gitlab": {
      "enabled": true,
      "credentials": {
        "base_url": "https://gitlab.com",
        "token": "glpat-..."
      }
    }
  }
}
```

Env: `GITLAB_BASE_URL`, `GITLAB_TOKEN`. Web setup: `/integrations/gitlab/setup` (instance URL + PAT).

### Security CI

Bumps `google.golang.org/grpc` to **v1.83.2** (and tidy-driven transitive updates) so `go tool govulncheck ./...` passes (fixes GO-2026-6443 / GO-2026-6348 call chains through gRPC server and GCP clients). Unrelated to GitLab adapter logic.

### Out of scope

- Official GitLab MCP proxy / Phase 2 REST fallback (this PR **is** the native REST path)
- README env-table expansion (reverted MCP README section)

## Test plan

- [x] `go test ./integrations/gitlab`
- [x] `go tool govulncheck ./...`
- [x] `make build vet`
- [ ] CI full matrix on push

<div><a href="https://cursor.com/agents/bc-984e2c61-1857-5463-846e-fa6ef6e420b1?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-984e2c61-1857-5463-846e-fa6ef6e420b1&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


integrations/gitlab/gitlab.go
```diff
@@ -0,0 +1,241 @@
...
+			s += "?" + query.Encode()
+		}
+		return s
+	}
+	prefix := strings.TrimSuffix(inst.Path, "/")
+	apiPath := prefix + apiV4Path + path
+	inst.Path = apiPath
+	if strings.Contains(path, "%2F") {
+		inst.RawPath = apiPath
+	}
```
acmacalister, line 98: Path-based `project_id` values like `gitlab-org/gitlab` end up double-encoded here. `encodeProjectID` turns `/` into `%2F`, then we set both `Path` and `RawPath` to that already-escaped string. `url.URL.String()` / `EscapedPath()` re-escape the `%`, so the request goes out as `%252F` and GitLab 404s.

Verified with httptest: escaped path is `/api/v4/projects/gitlab-org%252Fgitlab` instead of `/api/v4/projects/gitlab-org%2Fgitlab`. Numeric IDs still work, which is why the existing handler tests pass.

Go needs the unescaped form in `Path` and the escaped form in `RawPath`:

```go
unescaped := strings.ReplaceAll(apiPath, "%2F", "/")
inst.Path = unescaped
if strings.Contains(apiPath, "%2F") {
	inst.RawPath = apiPath
}
```

Or skip pre-encoding and build with `url.JoinPath` / `url.PathEscape` per segment. Worth adding a handler test that asserts the wire path for `project_id: "group/repo"` so this doesn't regress.

integrations/gitlab/gitlab.go
```diff
@@ -0,0 +1,241 @@
...
+
+func encodeProjectID(id string) string {
+	id = strings.TrimSpace(id)
+	if id == "" {
+		return id
+	}
+	if strings.Contains(id, "%") {
+		return id
+	}
+	return strings.ReplaceAll(id, "/", "%2F")
```
acmacalister: This is the other half of the path-id bug: replacing `/` with `%2F` by hand fights `net/url`, which wants to own escaping. `url.PathEscape(id)` is the right primitive if we keep pre-encoding, but only if `apiURL` treats the result as `RawPath` with a decoded `Path` (or we stop encoding here and let `JoinPath` do it).

Also, the `strings.Contains(id, "%")` short-circuit trusts caller-encoded input without validating it is a correct escape of the path.

integrations/gitlab/projects.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func handleListProjects(ctx context.Context, g *gitlab, args map[string]any) (*mcp.ToolResult, error) {
+	r := mcp.NewArgs(args)
+	q := paginationQuery(r)
+	q.Set("membership", "true")
+	if args != nil {
+		if v, ok := args["membership"]; ok {
+			if b, ok := v.(bool); ok && !b {
+				q.Del("membership")
```
acmacalister, line 26: `membership` only flips off when the value is a real Go `bool`. Tool args from LLMs are often strings (`"false"` / `"true"`) or JSON numbers, so `membership=false` can silently keep the default `membership=true`.

Something like `mcp.ArgBool` / a small truthy helper would match how other adapters handle optional booleans.

web/web.go
```diff
@@ -1239,6 +1243,81 @@ func (w *WebServer) handleFigmaSetup(rw http.ResponseWriter, r *http.Request) {
...
+	}
+	baseURL := strings.TrimSpace(r.FormValue("base_url"))
+	if baseURL == "" {
+		baseURL = "https://gitlab.com"
+	}
+	ic, _ := w.services.Config.GetIntegration("gitlab")
+	if ic == nil {
+		ic = &mcp.IntegrationConfig{Credentials: mcp.Credentials{}}
+	}
+	ic.Credentials["base_url"] = baseURL
```
acmacalister: We store `base_url` raw here, while `Configure` runs it through `normalizeInstanceURL`. Invalid schemes, credentials-in-URL, or a pasted `/api/v4` path can sit in config until the next health check, and the setup page will keep showing the unnormalized value.

Calling the same normalizer on save (and flashing the error on failure) would keep the UI and adapter in sync.

integrations/gitlab/handlers_test.go
```diff
@@ -0,0 +1,80 @@
...
+	cases := []struct {
+		tool     mcp.ToolName
+		args     map[string]any
+		method   string
+		path     string
+		body     string
+		response string
+	}{
+		{"gitlab_list_projects", nil, "GET", "/api/v4/projects", "", `[{"id":1,"name":"demo"}]`},
+		{"gitlab_get_project", map[string]any{"project_id": "42"}, "GET", "/api/v4/projects/42", "", `{"id":1}`},
```
acmacalister, line 25: All of these cases use numeric `project_id`s, so they never exercise path-style IDs (`group/project`) that the tool docs advertise. That is exactly the encoding bug above — a single case with `project_id: "group/repo"` and an assertion on `r.URL.EscapedPath()` would have failed CI.

acmacalister (commented):
Solid native REST adapter — dispatch parity, compaction specs, redirect blocking, and the Forgejo-shaped config/UI all look right. CI is green (build/test/lint/security/compose/rust-sdk). One real correctness issue: path-based `project_id` values get double-encoded to `%252F`, so anything beyond numeric IDs will 404. A couple of smaller follow-ups on `membership` coercion and base_url save validation.

integrations/gitlab/compact.yaml
```diff
@@ -0,0 +1,135 @@
...
+      - status
+      - stage
+      - ref
+      - web_url
+      - duration
+      - failure_reason
+      - started_at
+      - finished_at
+  gitlab_get_job_trace:
+    max_bytes: 262144
```
acmacalister: PR description and `MaxResponseBytesForTool` both advertise a 1 MiB job-trace ceiling, but `max_bytes: 262144` here is the post-compaction hard stop in `processResult`. A ~300-900 KiB failed job log is fully downloaded by `doText` (1 MiB cap), then replaced with a `response_too_large` envelope before the LLM ever sees it.

That makes the main CI-debug path fragile for real traces. Worth aligning this with the 1 MiB fetch/transport cap (or dropping `max_bytes` and relying on `MaxResponseBytesForTool` alone, like `github_get_pull_diff`).

acmacalister (approved):
Solid native REST adapter — scoped tools, dispatch/compaction parity, path-encoding fix with regression tests, and the earlier membership/base_url notes all look addressed. CI is green (build/test/lint/security/compose/rust-sdk). One leftover size-cap mismatch on job traces inline.

integrations/gitlab/pipelines.go
```diff
@@ -0,0 +1,63 @@
...
+	jobID := r.Int("job_id")
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	path := fmt.Sprintf("/projects/%s/jobs/%d/trace", projectID, jobID)
+	trace, _, err := g.doText(ctx, path, url.Values{}, 1024*1024)
+	if err != nil {
+		return mcp.ErrResult(err)
+	}
+	return mcp.JSONResult(map[string]string{"trace": trace})
```
acmacalister: `doText` and `MaxResponseBytesForTool` both use a 1 MiB ceiling, but the trace is returned as `JSONResult({"trace": ...})`. JSON string escaping adds overhead (newlines alone turn into `\n`), so a full 1 MiB fetch often lands a few KB over the post-process cap and gets replaced with `response_too_large` — undoing the compact `max_bytes` fix for real logs.

`github_get_pull_diff` avoids this by returning plain text (`ToolResult{Data: diff}`). Same idea here, or leave a little headroom in the fetch cap so the wrapped payload stays under 1 MiB.

acmacalister (approved):
Nice native REST adapter — scoped tools, dispatch/compaction parity, and the earlier path-encoding / membership / job-trace compact fixes all look solid. CI is green (build, test, lint, security, compose, rust-sdk). One leftover size-cap wrinkle on job traces inline.

integrations/gitlab/gitlab.go
```diff
@@ -0,0 +1,250 @@
...
+	resp, err := hc.Do(req)
+	if err != nil {
+		return nil, 0, err
+	}
+	defer func() { _ = resp.Body.Close() }()
+	data, err := io.ReadAll(io.LimitReader(resp.Body, 8*1024*1024))
+	if err != nil {
+		return nil, resp.StatusCode, err
+	}
+	if resp.StatusCode >= 400 {
```
acmacalister: `CheckRedirect` stops at the first hop (good — keeps the PAT off redirect targets), but we only treat `status >= 400` as failure. A `301`/`302` (common for `http→https` or a trailing-slash instance URL) lands here as "success" with an empty or HTML body, then becomes `null` or garbage via `jsonResult`.

Forgejo uses the same redirect policy and fails anything `>= 300` — matching that here (and in `doText`) would surface misconfigured `base_url` instead of a silent empty result.

```go
if resp.StatusCode >= 300 {
    msg := strings.TrimSpace(string(data))
    if msg == "" {
        msg = http.StatusText(resp.StatusCode)
    }
    return nil, resp.StatusCode, fmt.Errorf("gitlab: HTTP %d: %s", resp.StatusCode, msg)
}
```

integrations/gitlab/pipelines.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func handleListPipelineJobs(ctx context.Context, g *gitlab, args map[string]any) (*mcp.ToolResult, error) {
+	r := mcp.NewArgs(args)
+	projectID := encodeProjectID(r.Str("project_id"))
+	pipelineID := r.Int("pipeline_id")
+	if err := r.Err(); err != nil {
+		return mcp.ErrResult(err)
+	}
+	q := paginationQuery(r)
```
acmacalister: Minor consistency nit: the other list handlers call `paginationQuery(r)` before `r.Err()`. Here it sits after the guard. `OptInt` ignores type errors and `validateArgs` already type-checks `page`/`per_page`, so this isn't a silent-zero bug today — just worth aligning with `handleListPipelines` so future extractions don't land after the guard by habit.

acmacalister (approved):
Nice native REST adapter — scoped tools, dispatch/compaction parity, path-encoding + job-trace plain-text fixes from earlier rounds all look solid. CI is green (build, test, lint, security, compose, rust-sdk). One reliability note on 3xx handling inline, plus a small consistency nit.

acmacalister (approved):
Clean PR — CI is green (build, test, lint, security, compose, rust-sdk). Native REST adapter looks solid: dispatch/compaction parity, path encoding, 3xx handling, plain-text job traces, and the earlier membership/base_url fixes all check out. LGTM.

integrations/gitlab/gitlab.go
```diff
@@ -0,0 +1,253 @@
...
+	if err != nil || inst.Scheme == "" || inst.Host == "" {
+		s := base + apiV4Path + path
+		if len(query) > 0 {
+			s += "?" + query.Encode()
+		}
+		return s
+	}
+	prefix := strings.TrimSuffix(inst.Path, "/")
+	apiPath := prefix + apiV4Path + path
+	inst.Path = strings.ReplaceAll(apiPath, "%2F", "/")
```
acmacalister: The `%2F` fix from earlier still leaves other escapes half-decoded. `encodeProjectID` correctly produces `%20` / `%C3%BC` / etc., but here we only turn `%2F` into `/` for `Path` and leave the rest as literal percent sequences. `url.URL.String()` then re-escapes those `%` signs, so a path-style id like `group/repo with space` goes out as `.../group/repo%2520with%2520space` (and `group/über` becomes `%25C3%25BCber`).

`group/repo` and numeric ids are fine — that's why the existing tests pass — but any project path with spaces or non-ASCII still 404s.

Full `PathUnescape` for `Path` (and `RawPath = apiPath` whenever it differs) matches what Go expects:

```go
unescaped, err := url.PathUnescape(apiPath)
if err != nil {
	unescaped = strings.ReplaceAll(apiPath, "%2F", "/")
}
inst.Path = unescaped
if unescaped != apiPath {
	inst.RawPath = apiPath
}
```

Worth extending `TestAPIURL_ProjectPathNotDoubleEncoded` / the handler encoding test with a space or unicode path so this doesn't regress again.

acmacalister (approved):
Solid native REST adapter — scoped tools, dispatch/compaction parity, and the earlier rounds of path/membership/3xx/job-trace fixes all look good. CI is green (build, test, lint, security, compose, rust-sdk). One leftover encoding edge on non-slash percent escapes inline.

acmacalister (approved):
Clean PR — CI is green (build, test, lint, security, compose, rust-sdk) and the earlier path-encoding / membership / 3xx / job-trace rounds all look fully addressed on 3d3bac0. Native REST shape, dispatch/compaction parity, and Forgejo-aligned setup look solid. LGTM.

PR #241, opened 2026-09-17
Add Switchboard PR shepherd skill for Orca automations

daltoniam:
## Summary
- Adds a `daltoniam/switchboard` PR shepherd skill that uses `make ci` instead of the hosted Switchboard-hosted `task pr:check` loop.
- Documents the skill in AGENTS.md so the daily Orca integration automations can open and clean PRs without merging.

## Test plan
- [x] Skill exists at `.agents/skills/pr-shepherd/SKILL.md` and is mirrored through the `.claude/skills` symlink.
- [ ] Confirm Orca research/implement automations stay disabled until this lands on main.


cmd/server/composition_test.go
```diff
@@ -77,7 +77,7 @@ func TestProductionGRPC_UnixSocketDoesNotServeHTTP(t *testing.T) {
 		CatalogEnabled: true,
 		WritesEnabled:  true,
 	})
-	path := filepath.Join(t.TempDir(), "awm.sock")
+	path := filepath.Join(os.TempDir(), "sb-awm.sock")
```
acmacalister: Shortening the path for macOS `sun_path` limits makes sense, but a fixed name under `os.TempDir()` is shared process-wide. A leftover listener, a crashed prior run, or two packages hitting `/tmp` at once can turn this into `already in use` / flaky failures — and it steps outside the isolation `t.TempDir()` normally gives us.

A short *and* unique dir keeps the path length fix without the global collision:

```go
dir, err := os.MkdirTemp("", "sb")
require.NoError(t, err)
t.Cleanup(func() { _ = os.RemoveAll(dir) })
path := filepath.Join(dir, "awm.sock")
```

`MkdirTemp("", "sb")` lands under `/tmp/sb-<rand>/awm.sock`, which stays well under macOS’s ~104-byte limit while still being test-private.

acmacalister (approved):
CI is green across build/test/lint/security/rust-sdk/compose. The shepherd skill is thorough and correctly gates on `make ci` plus exact-head rebinding without auto-merge.

One non-blocking note on the unix socket path in the composition test — short is good, fixed global name is a flake risk.

cmd/server/composition_test.go
daltoniam: Fixed in 40ef71ebcd99ac8c255538816285b9321e46f483. The composition test now uses os.MkdirTemp("", "sb") so the unix socket stays short and unique. go test -race ./cmd/server -run TestProductionGRPC_UnixSocketDoesNotServeHTTP and make ci both passed on that head.

acmacalister (approved):
Clean PR — CI/status checks are passing (build, test, lint, security, rust-sdk, compose on 40ef71e) and I didn't find any blocking issues. Prior unix-socket path feedback is already fixed with MkdirTemp. LGTM.

